PR #13, opened 2019-06-10
Initial cleanup/removal of IChannelListener

mconnew:
Removed IChannelListener interface and common listener base classes
Removed some additional unnecessary interfaces.
Removed lingering NetTcp channel listener implementation
Some nameof cleanup


paulomorgado:
Just out of curiosity, why is `IChannelListener` being removed?

And non-functional changes like the use of `nameof` should have their own PRs.

mconnew:
@paulomorgado, IChannelListener is a pull model where you have a loop which repeatedly calls Accept to get a channel, and then on that channel you need another loop to Receive a message which then pushes it to the dispatcher. As we're now built on top of Asp.NET Core, this model no longer works. Asp.NET Core uses a middleware model where the request is passed along a series of middleware in a push model. If we kept IChannelListener, we would need to have code which translates the push model of Asp.NET to IChannelListener and a whole bunch of new problems need to be solved. For example, if Asp.NET pushes a new message and there's not a pending Accept, you need to hold on to the message until there is one. So then you need to start queuing the requests and need a mechanism to prevent too many building up. You could start sending a 500 response, but then do you do that to items currently queued or to new requests? And all this just to go back to a push model again at the dispatcher. All the built-in binding elements will be made to work with the new model and once the api is stabilized, it will be documented how to create a new binding element this way. It should be relatively easy to modify any existing custom channel layer code to the new model.  
As for non-functional cleanup such as `nameof`, I would normally agree with you but with the current stage this code is in, that will slow development down a lot. In the coming weeks I will be making large changes to the code base and getting things to a state where others can contribute easily. Until that point though, I'm going to be taking a lot of liberties in the regular code development process. I hope this stage will only be a few weeks.

Kralizek:
As a lover of WCF, I'm really curious to see how WCF over ASP.NET will work!

paulomorgado:
With analyzers/fixers like [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/) and [codecracker](https://www.nuget.org/packages/codecracker.CSharp/) this is almost one shot.

An `.editorconfig` file can also be used with [dotnet-format](https://www.nuget.org/packages/dotnet-format/) for normalizing style.

Mixing style and functionality makes it hard to understand the changes.

mconnew:
@paulomorgado, please feel free to contribute to improve those areas. I would suggest opening an issue first to discuss what checks, style and config would be good as well as what tools to use. Then you can submit a PR to implement/fix some of this stuff. 
I'm mostly concentrating on the code functionality at the moment and as I know this code back to front, style hasn't been a priority for me so far. But one of my goals with this re-worked code base is to make the code easy to reason about and to lower the technical skills bar for contributions and clean code helps with that.

mconnew:
@tibor19, I've done another round of code cleanup. Can you have a quick look and ack so I can merge?

PR #26, opened 2019-07-02
Dispatch cleanup

mconnew:
Added back a lot of extensibility capability. No tests for any of them yet.


src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -571,6 +571,9 @@ static bool HaveCommonInitiatingActions(EndpointFilterProvider x, EndpointFilter
             {
                 var cd = cdb as ChannelDispatcher;
                 cd.Init();
+                var openTask = cd.OpenAsync();
+                Fx.Assert(openTask.IsCompleted, "ChannelDispatcher should open synchronously");
```
tibor19: What is the point of calling OpenAsync if the expectation is to open synchronously?

mconnew: The class derives from CommunicationObject. On .NET Framework that class has Open and BeginOpen/EndOpen methods and I've replaced them with just OpenAsync. The class ChannelDispatcher allows modification when it's in the Created state. A behavior would typically modify something on ChannelDispatcher, for example throttles. When you "open" a ChannelDispatcher, it doesn't do anything other than transition to the Open state. All the setters on ChannelDispatcher throw an exception if it's not in the Created state. Basically once it's open it becomes immutable. The Fx.Assert is there to make sure the class isn't modified at a later date to do something asynchronously when OpenAsync is called without forcing awareness that OpenAsync is being synchronously waited on here.

mconnew:
@tibor19, have you finished reviewing this PR?

tibor19:
I am done!

PR #33, opened 2019-08-22
Implemented server side connection reuse

mconnew:
Also implemented channel/connection close on app close
Also implemented unwrapping of upgraded connection
Upgraded Kestrel dependency version


src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        public RawStream(PipeReader input, PipeWriter output)
+        {
+            _input = input;
+            _output = output;
+            _canRead = true;
+            _thisLock = new object();
+        }
+
+        public override bool CanRead
```
tibor19: Why the inconsistency between CanRead and the other Can... methods?

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
```diff
@@ -0,0 +1,264 @@
...
+            // https://github.com/dotnet/corefx/blob/f9da3b4af08214764a51b2331f3595ffaf162abe/src/System.Threading.Tasks.Extensions/src/System/Threading/Tasks/ValueTask.cs#L156
+            return ReadAsyncInternal(new Memory<byte>(buffer, offset, count)).Result;
+        }
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            return ReadAsyncInternal(new Memory<byte>(buffer, offset, count)).AsTask();
+        }
+
+#if NETCOREAPP2_1
```
tibor19: What about .net core 3 and above?

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
```diff
@@ -0,0 +1,264 @@
...
+        public RawStream(PipeReader input, PipeWriter output)
+        {
+            _input = input;
+            _output = output;
+            _canRead = true;
+            _thisLock = new object();
+        }
+
+        public override bool CanRead
+        {
```
tibor19: Should not this be used only to say that the Stream supports read operations? IMO Blocking has no place here, but when you do the reading itself, and can read should always return true.

PR #45, opened 2019-09-16
Need to use || instead of &&

DamirAinullin:
Fix of incorrect condition and replacing string arg name with nameof operator.


src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
```diff
@@ -1080,9 +1080,9 @@ public int GetSessionKey(int value)
 
         int GetPrefixOffset(char prefix)
         {
-            if (prefix < 'a' && prefix > 'z')
+            if (prefix < 'a' || prefix > 'z')
```
mconnew: Nice catch. This is a bug which exists in .NET Framework too!! I'll have to take a look and work out what the ramifications are for this bug and get it fixed if it's severe enough.

PR #47, opened 2019-09-16
Added throw operator before TraceUtility.ThrowHelperError

DamirAinullin:
in order to throw exception

### DIFF
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Dispatcher/ErrorBehaviorHelper.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Dispatcher/ErrorBehaviorHelper.cs
@@ -32,7 +32,7 @@ namespace CoreWCF.Dispatcher
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


src/CoreWCF.NetTcp/src/CoreWCF/Dispatcher/ErrorBehaviorHelper.cs
```diff
@@ -32,7 +32,7 @@ internal static void ThrowAndCatch(Exception e, Message message)
                     }
                     else
                     {
-                        TraceUtility.ThrowHelperError(e, message);
+                        throw TraceUtility.ThrowHelperError(e, message);
```
mconnew: This was intentional. The idea is to throw if a debugger is attached, otherwise don't throw. Your change makes the if block and the else block identical.

DamirAinullin: Ok, I see

PR #48, opened 2019-09-16
Removed excessive code

DamirAinullin:
Seems like no sense in this code so I suggest to delete it andreplace just with return statement.


mconnew:
There's some important functionality which has been kind of semi removed here and needs to be put back in. On .NET Framework you can specify an identity for the service. This then gets configured through to the IChannelListener and the EndpointIdentity property can be retrieved. With the code you are removing, I'd rather put back the original code which can be found [here](https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/Dispatcher/EndpointDispatcher.cs,a5b882920a67c8ea) and comment it out. Then add a TODO that a mechanism to configure a custom endpoint identity needs to be created and then used here.

PR #51, opened 2019-10-10
Added a Postman SOAP 1.1 Collection to illustrate raw requests to the basichttp endpoint

VorTechS:
The included file can be imported into Postman to send a request to either the DesktopServer or NetCoreServer sample applications.  

It shows how the headers and body should be constructed to correctly send a SOAP request to the server, ideal for testing on-the-fly


mconnew:
Can you add a README.MD in the same folder as the postman json file with your "How to use" instructions. I don't think you need to elaborate more than what you've already put in the comment.

VorTechS:
Sorry, I had a few days with head down at work and chilling out of an evening!

I've added that file as requested.  I'm not familiar with Git, so hopefully I've done that right!  I also added some sample code into the client applications showing the same envelope being sent by code, which might also be useful.

src/README.MD
```diff
@@ -0,0 +1,9 @@
+Using the CoreWCF Postman Samples Collection.json within Postman
```
mconnew: Can you move this file to the src/Samples folder instead of being at the top level

src/Samples/DesktopClient/Properties/launchSettings.json
```diff
@@ -0,0 +1,7 @@
+{
```
mconnew: I presume this is from VSCode? Can you remove this from your PR. While I do see benefit in having this, it should be done purposefully and across the entire repo for every project where appropriate rather than a single adhoc project.

src/Samples/NetCoreClient/Program.cs
```diff
@@ -15,14 +21,43 @@ static void Main(string[] args)
...
-            factory = new ChannelFactory<Contract.IEchoService>(new BasicHttpBinding(), new EndpointAddress("http://localhost:8080/basichttp"));
+            factory = new ChannelFactory<Contract.IEchoService>(new BasicHttpBinding(), new EndpointAddress(_basicHttpEndPointAddress));
             factory.Open();
             channel = factory.CreateChannel();
             ((IClientChannel)channel).Open();
             Console.WriteLine("http Echo(\"Hello\") => " + channel.Echo("Hello"));
             ((IClientChannel)channel).Close();
             factory.Close();
 
+            //
```
mconnew: Can you refactor the two ways into two methods and have the Main method call each in turn. I'll leave the method naming to you, but as a suggestion, "CallUsingWCF" and "CallUsingWebRequest".

src/Samples/DesktopClient/Properties/launchSettings.json
VorTechS: I use VS2019, I'll post edit accordingly.

src/Samples/NetCoreClient/Program.cs
VorTechS: Certainly!  I refrained from doing so based on the relevant rules!  (Yes, I'm the one person that did read them!) ;)

tibor19:
There are some conflicts. @VorTechS Can you have a look and fix them so we can commit this?

tibor19:
Ping @VorTechS

VorTechS:
Sorry!  Had a machine re-build and lost track of things.  I'll try and get this done ASAP!

VorTechS:
Today's GitHub lesson; committing extra code doesn't resolve conflicts.  Sorry, I'm still getting used to it coming from a SourceSafe / TFS background!

Anyway, before I had my machine rebuild (and pulling latest changes) I had a WCF complex type sample working.  But now it appears it no longer is.

The appropriate method is being called, but the SOAP message parameters aren't being unwrapped correctly.  I thought perhaps it was a namespace conflict and modifying the namespace for the classes so that the EchoMessage class was the same (NetCoreServer) made no difference, nor did moving it to the Contract class.

Sample SOAP message being sent:

```
{<?xml version="1.0" encoding="utf-16"?>
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
  <s:Header>
    <To s:mustUnderstand="1" xmlns="http://schemas.microsoft.com/ws/2005/05/addressing/none">http://localhost:8080/basichttp</To>
    <Action s:mustUnderstand="1" xmlns="http://schemas.microsoft.com/ws/2005/05/addressing/none">http://tempuri.org/IEchoService/ComplexEcho</Action>
  </s:Header>
  <s:Body>
    <ComplexEcho xmlns="http://tempuri.org/">
      <text xmlns:a="http://schemas.datacontract.org/2004/07/NetCoreClient" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
        <a:Text>Complex Hello</a:Text>
      </text>
    </ComplexEcho>
  </s:Body>
</s:Envelope>}
```

PR #52, opened 2019-10-10
Add azure pipelines configuration

### LINKED ISSUE #21: Add Azure Pipelines for CI
Would be good to have the tests run on PRs to make sure nothing is breaking. Since this is a OSS project, it should be able to leverage the free offering

Daniel-Svensson:
* Setup a *basic* build, test and package pipeline using azure pipelins
* Use gitversion to get incremented release numbers based on git tags
* Installs latests Microsoft.Test.SDK into all test projects to enable code coverage
* Disable packaging for sample projects

**net472 is not built and tested** since the tests hangs since #50 

For build status see
https://svensson-daniel.visualstudio.com/CoreWCF/_build/results?buildId=628&view=results

This is *one of the steps* required to fix #21
Once this is in master the azure pipelines github app needs to be installed, a devops project with corresponding build setup, but at least the build can be setup with an existing yml file

Future work could include adding some kind of static analysis (I personally find sonarcloud quite good) support for deterministic builds, sourcelink support as well as building on different os.
 (and maybe a release pipeline to create github releases and push to nuget)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <ProjectDir>$(MSBuildThisFileDirectory)</ProjectDir>
     <SourceDir>$(ProjectDir)src/</SourceDir>
-    
+
     <!-- Output directories -->
     <BinDir Condition="'$(BinDir)'==''">$(ProjectDir)bin/</BinDir>
     <ObjDir Condition="'$(ObjDir)'==''">$(BinDir)obj/</ObjDir>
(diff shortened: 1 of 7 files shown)


Daniel-Svensson:
@mconnew  If you dont want to setup the devops bits by yourself I can help with that but then you would need to provide member access to the github repo (during the setup) and install the devops app (i think it requires admin access).

That said, nice with with this project :+1: I really hope that it succeeds and gain traction

mconnew:
Thanks for this. I won't be able to look at this until Tuesday next week and I want to understand what these changes all mean first. I've been fighting an auth issue in Azure DevOps to get that part set up but this will definitely help with getting this going.

Daniel-Svensson:
I think it should be considered to to split the testing step into one per target framework since the result view in devops does not make the targetframework very easy to se in the test results

mconnew:
I'm playing around with things trying to get a simple pipeline working. I'm about to create a PR for fixing some test issues and a couple of product issues which I think were getting hit by running in Azure devops. I believe I've removed all problems with socket reuse between tests and with exceptions being thrown resulting in missed cleanup.

mconnew:
I'm going to close this PR. I'm still going to refer back to it for inspiration for other changes. Feel free to submit a PR to suggest improvements to the current pipelines file.

PR #54, opened 2019-10-10
Replace Task to ToApm wrapper with a leaner implementation

Daniel-Svensson:
Replace the current ToApm wrapper with a more lightweight approach
* Allocates less (1 allocation instead of TaskCompletionSource + Task + Tuple)
* Performs slightly less work
* Properly implement the `CompletedSynchronously` property of IAsyncResult
* Ready for wrapping `ValueTask` based methods to make it easier to start using it

I understand if this does not have a high priority, but I hade implemented this not to long ago for a project based on WCF and wanted to share it.

### DIFF
--- a/src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
+++ b/src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
@@ -30,35 +30,25 @@ namespace CoreWCF.Runtime
         // In the EndMethod, you would use ToApmEnd<TResult> to ensure the correct exception handling
         // This will handle throwing exceptions in the correct place and ensure the IAsyncResult contains the provided
         // state object
-        public static IAsyncResult ToApm<T>(this Task<T> task,
-                                   AsyncCallback callback,
-                                   object state)
+        public static IAsyncResult ToApm<T>(this Task<T> task, AsyncCallback callback, object state)
             => ToApm<T>(new ValueTask<T>(task), callback, state);
 
         /// <summary>
         /// Helper method to convert from Task async method to "APM" (IAsyncResult with Begin/End calls)
         /// </summary>
-        public static IAsyncResult ToApm<T>(this ValueTask<T> task,
-                                    AsyncCallback callback,
-                                    object state)
+        public static IAsyncResult ToApm<T>(this ValueTask<T> task, AsyncCallback callback, object state)
         {
             var result = new AsyncResult<T>(task, callback, state);
             if (result.CompletedSynchronously)
             {
                 result.ExecuteCallback();
-                return result;
             }
-            else
+            else if (callback != null)
             {
-                task.AsTask()
-                    .ContinueWith((res, asyncResult) =>
-                    {
-                        ((AsyncResult<T>)asyncResult).ExecuteCallback();
-                    },
-                result,
-                CancellationToken.None,
-                TaskContinuationOptions.HideScheduler,
-                TaskScheduler.Default);
+                task.AsTask().ContinueWith((res, asyncResult) =>
+                {
+                    ((AsyncResult<T>)asyncResult).ExecuteCallback();
+                }, result, CancellationToken.None, TaskContinuationOptions.HideScheduler, TaskScheduler.Default);
             }
 
             return result;
@@ -67,26 +57,19 @@ namespace CoreWCF.Runtime
         /// <summary>
         /// Helper method to convert from Task async method to "APM" (IAsyncResult with Begin/End calls)
         /// </summary>
-        public static IAsyncResult ToApm(this Task task,
-                                    AsyncCallback callback,
-                                    object state)
+        public static IAsyncResult ToApm(this Task task, AsyncCallback callback, object state)
         {
             var result = new AsyncResult(task, callback, state);
             if (result.CompletedSynchronously)
             {
                 result.ExecuteCallback();
-                return result;
             }
-            else
+            else if (callback != null)
             {
                 task.ContinueWith((res, asyncResult) =>
-                    {
-                        ((AsyncResult)asyncResult).ExecuteCallback();
-                    },
-                result,
-                CancellationToken.None,
-                TaskContinuationOptions.HideScheduler,
-                TaskScheduler.Default);
+                {
+                    ((AsyncResult)asyncResult).ExecuteCallback();
+                }, result, CancellationToken.None, TaskContinuationOptions.HideScheduler, TaskScheduler.Default);
             }
 
             return result;
@@ -116,50 +99,40 @@ namespace CoreWCF.Runtime
             }
         }
 
-        internal class AsyncResult : IAsyncResult
+        private class AsyncResult : IAsyncResult
         {
             private readonly Task _task;
-            private readonly object _asyncState;
             private readonly AsyncCallback _asyncCallback;
 
             public AsyncResult(Task task, AsyncCallback asyncCallback, object asyncState)
             {
-                Debug.Assert(asyncCallback != null);
-
                 _task = task;
                 _asyncCallback = asyncCallback;
-                _asyncState = asyncState;
+                AsyncState = asyncState;
                 CompletedSynchronously = task.IsCompleted;
             }
 
             public void GetResult() => _task.GetAwaiter().GetResult();
 
-            // Calls the async callback with this
-            public void ExecuteCallback() => _asyncCallback(this);
+            public void ExecuteCallback() => _asyncCallback?.Invoke(this);
 
-            #region IAsyncResult implementation forwarded to Task implementation
-            object IAsyncResult.AsyncState => _asyncState;
+            public object AsyncState { get; }
             WaitHandle IAsyncResult.AsyncWaitHandle => ((IAsyncResult)_task).AsyncWaitHandle;
 
             public bool CompletedSynchronously { get; }
             public bool IsCompleted => _task.IsCompleted;
-
-            #endregion
         }
 
         internal class AsyncResult<T> : IAsyncResult
         {
             private readonly ValueTask<T> _task;
-            private readonly object _asyncState;
             private readonly AsyncCallback _asyncCallback;
 
             public AsyncResult(ValueTask<T> task, AsyncCallback asyncCallback, object asyncState)
             {
-                Debug.Assert(asyncCallback != null);
-
                 _task = task;
                 _asyncCallback = asyncCallback;
-                _asyncState = asyncState;
+                AsyncState = asyncState;
                 CompletedSynchronously = task.IsCompleted;
             }
 
@@ -168,17 +141,13 @@ namespace CoreWCF.Runtime
             public bool IsFaulted => _task.IsFaulted;
             public AggregateException Exception => _task.AsTask().Exception;
 
-            // Calls the async callback with this
-            public void ExecuteCallback() => _asyncCallback(this);
-
-            #region IAsyncResult implementation forwarded to Task implementation
-            object IAsyncResult.AsyncState => _asyncState;
+            // Calls the async callback with this as parameter
+            public void ExecuteCallback() => _asyncCallback?.Invoke(this);
+            public object AsyncState { get; }
             WaitHandle IAsyncResult.AsyncWaitHandle => !CompletedSynchronously ? ((IAsyncResult)_task.AsTask()).AsyncWaitHandle : throw new NotImplementedException();
 
             public bool CompletedSynchronously { get; }
             public bool IsCompleted => _task.IsCompleted;
-
-            #endregion
         }
 
         // Awaitable helper to await a maximum amount of time for a task to complete. If the task doesn't
--- a/src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
+++ b/src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
@@ -48,7 +48,7 @@ namespace CoreWCF.Runtime
                 valueTask.AsTask().ContinueWith((_, asyncResult) =>
                 {
                     ((AsyncResult<T>)asyncResult).ExecuteCallback();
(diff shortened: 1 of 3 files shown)


src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
```diff
@@ -29,120 +30,124 @@ public static async Task AsyncWait<TException>(this Task task)
...
             {
-                var tuple = obj as Tuple<TaskCompletionSource<TResult>, AsyncCallback>;
-                var tcsObj = tuple.Item1;
-                var callbackObj = tuple.Item2;
-                if (antecedent.IsFaulted) tcsObj.TrySetException(antecedent.Exception.InnerException);
-                else if (antecedent.IsCanceled) tcsObj.TrySetCanceled();
-                else tcsObj.TrySetResult(antecedent.Result);
-
-                if (callbackObj != null)
+                task.AsTask().ContinueWith((res, asyncResult) =>
```
Daniel-Svensson: do you prefer to use "_" (discard) for unused parameter?

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
```diff
@@ -29,120 +30,124 @@ public static async Task AsyncWait<TException>(this Task task)
...
+        /// <summary>
+        /// Helper method to convert from Task async method to "APM" (IAsyncResult with Begin/End calls)
+        /// </summary>
+        public static IAsyncResult ToApm(this Task task, AsyncCallback callback, object state)
         {
-            // When using APM, the returned IAsyncResult must have the passed in state object stored in AsyncState. This
-            // is so the callback can regain state. If the incoming task already holds the state object, there's no need
-            // to create a TaskCompletionSource to ensure the returned (IAsyncResult)Task has the right state object.
-            // This is a performance optimization for this special case.
-            if (task.AsyncState == state)
```
Daniel-Svensson: The "fast path" is skipped since it never happens where it is used in the framework since state parameters (and the callback) are always used and I don't see why it should change

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SyncMethodInvoker.cs
```diff
@@ -63,17 +63,12 @@ public IAsyncResult InvokeBegin(object instance, object[] inputs, AsyncCallback
...
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentException(SR.SFxInvalidCallbackIAsyncResult));
-            }
-
-            var tuple = task.Result;
+            var tuple = TaskHelpers.ToApmEnd<Tuple<object, object[]>>(result);
             outputs = tuple.Item2;
             return tuple.Item1;
         }
 
+        // This should be changed to ValueTuple instead
```
Daniel-Svensson: I can remove this comment and try to undo indentation change caused by save if you like

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/TaskMethodInvoker.cs
```diff
@@ -80,9 +80,8 @@ public object InvokeEnd(object instance, out object[] outputs, IAsyncResult resu
             {
                 //AsyncMethodInvoker.GetActivityInfo(ref activity, ref boundOperation);
 
-                Task<Tuple<object, object[]>> invokeTask = result as Task<Tuple<object, object[]>>;
-
-                if (invokeTask == null)
+                var asyncResult = result as TaskHelpers.AsyncResult<Tuple<object, object[]>>;
```
Daniel-Svensson: I did not want to rewrite this method more than necessary but I think It would benefit from calling `ToApmEnd` instead which would involve moving the processing of the AggregateException "ae" into a try/catch block around the call to `ToApmEnd`

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
mconnew: I do prefer the new discard feature. It's quite a nice addition to the language.

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
mconnew: I've seen consumers of APM methods use this pattern:
```c#
  var result = EndFoo(BeginFoo(param, null, null));
```
I know that's a dirty pattern, but I've seen it enough, and actually used it myself in unit tests when I know the underlying implementation is synchronous. Although I'm not sure how important this is now because I believe I've removed every public APM api that I could find in CoreWCF and replaced with Task based ones. If there are any lingering, then they should be removed. So in theory this whole set of methods should be redundant.

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
```diff
@@ -29,120 +30,124 @@ public static async Task AsyncWait<TException>(this Task task)
...
+        /// </summary>
+        public static IAsyncResult ToApm<T>(this ValueTask<T> task, AsyncCallback callback, object state)
         {
-            // When using APM, the returned IAsyncResult must have the passed in state object stored in AsyncState. This
-            // is so the callback can regain state. If the incoming task already holds the state object, there's no need
-            // to create a TaskCompletionSource to ensure the returned (IAsyncResult)Task has the right state object.
-            // This is a performance optimization for this special case.
-            if (task.AsyncState == state)
+            var result = new AsyncResult<T>(task, callback, state);
+            if (result.CompletedSynchronously)
```
mconnew: This will call the callback twice in the scenario where the APM method is expected to have synchronous completion optimized. In other words, I should be able to write the following:
```c#
var iar = BeginFoo(param1, callback, state);
if (iar.CompletedSynchronously)
{
    var result = EndFoo(iar);
    callback(state);
}
```
This wasn't an issue before as the old implementation also returned false for CompletedSynchronously.

Daniel-Svensson: Based on my understanding i think code like that is "flawed" i.e it should not need the line "`callback(state);`" unless the user intend the callback to be executed twice.

I do belive the following is the correct usage.

``` csharp
var iar = BeginFoo(param1, callback, state);
if (iar.CompletedSynchronously)
{
    result = EndFoo(iar);
}
```
With having a corresponding check in the callback which just does nothing if CompletedSynchronously is true, just as the `Callback` method in the same file does.


This is based on [TaskToApm wrapper from the .Net Framework](https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/TaskToApm.cs,45)  (which I found today) 
as well as the respose to [this stackoverflow question](https://stackoverflow.com/questions/1372053/asynccallback-completedsynchronously) which reference this old [indigo (wcf) blog post](https://blogs.msdn.microsoft.com/mjm/2005/05/04/asynchronous-programming-in-indigo/) 

It seems like the .Net framework version is possibly even more generic and may handle many different usages better (apart from no ValueTask support) so you could consider using that instead.

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
Daniel-Svensson: I did only find the usage at 3 places and as you says it should probably not be used very much, so I don't believe it makes sense to add additional code optimizations which would only apply to tests. 

As for tasks before making the first "stable" release I would recommend you to consider if you maybe should use ValueTask at least for per call methods that are only awaited once. 
I don't have the links at the moment there is a proposal for corefx to make ValueTask 0 allocation (amortized) for simple usage and it is quite good for sync return.

Daniel-Svensson:
I will have another look at this and fix review comments if you find it makes sense to continue with this.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SyncMethodInvoker.cs
mconnew: Please do remove. Just make the file clean and don't worry about the diff looking clean.

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
mconnew: You're right. There are a couple of different variations of the APM pattern used in .NET Framework and I was getting my wires crossed.

mconnew:
I'm happy with the logic. Once you are happy with the code as well, just ack and I'll squash and merge.

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
mconnew: I 100% agree. What I first ported this code, ValueTask wasn't a thing yet which is why everything is still a Task. We definitely need to switch to using ValueTask before going stable as that could could be a nasty surprise.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/TaskMethodInvoker.cs
mconnew: The key thing is does it still work. I can't remember what range of tests we have around this but before making any drastic changes, we need to make sure there's tests for:
1. Task based method which completes synchronously
2. Task based method which completes async
3. Task<TResult> based method which completes synchronously
4. Task<TResult> based method which completes async
5. Task based method which returns a synchronously completed Task which denies child attach (continuation must execute async)
6. Task<TResult> based method which returns a synchronously completed Task which denies child attach (continuation must execute async)
7. Task based method which completes async and the returned Task denies child attach (continuation must execute async)
8. Task<TResult> based method which completes async and the returned Task denies child attach (continuation must execute async)

And any other variant that might cause difficulties with the callback being executed.

Daniel-Svensson: I will not make any changes but will add a comment with a link to your test cases

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SyncMethodInvoker.cs
Daniel-Svensson: comment removed

src/Common/src/CoreWCF/Runtime/TaskHelpers.cs
```diff
@@ -29,120 +30,124 @@ public static async Task AsyncWait<TException>(this Task task)
...
-
-                if (callbackObj != null)
+                valueTask.AsTask().ContinueWith((_, asyncResult) =>
                 {
-                    callbackObj(tcsObj.Task);
-                }
-            }, continuationState, CancellationToken.None, TaskContinuationOptions.HideScheduler, TaskScheduler.Default);
-            return tcs.Task;
+                    ((AsyncResult<T>)asyncResult).ExecuteCallback();
+                }, result, CancellationToken.None, TaskContinuationOptions.RunContinuationsAsynchronously, TaskScheduler.Default);
```
Daniel-Svensson: I made one new change by: 

1. adding `RunContinuationsAsynchronously` inspired by the .Net framework  code so that continuations are not allowed to executed sync on the thread where the task is completed. 
This should work fine, but has not been stress tested by me (tests still pass).

I am happy to revert this change if you think the old way was better

2. removing `HideScheduler` (which should not have any effect since TaskScheduler.Default will be the ambient scheduler both with and without the flag)

Daniel-Svensson:
@mconnew I am happy with the code now, but feel free to hold up merging to next week if you want to setup azure devops first.

davidfowl:
Consider using this https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Threading/Tasks/TaskToApm.cs

Kralizek:
> Consider using this https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Threading/Tasks/TaskToApm.cs

Hi @davidfowl, can I ask you the rationale behind keeping methods or classes like this one internal? As a library author I find this behavior very annoying.

davidfowl:
Public API adds support costs to us as library maintainers. We’re very intentional about what is made public and if it isn’t an API that the majority of people need then we err on the side of internal rathe than public.

Kralizek:
I see. It makes sense. Thanks for your time.

mconnew:
@Daniel-Svensson ,we do want to run the continuations asynchronously. When the task has already completed, the callback is executed inline which is the right thing to do. But if it completes between the check which flags it as completed synchronously and the call to ContinueWith, it will then run the continuation inline but this isn't the expected behavior when something completes asynchronously.

mconnew:
You've also removed the option to hide the TaskScheduler. This needs to be there as if there's a non-default SynchronizationContext set, a TaskScheduler will be used which uses this sync context unless you specify to hide the scheduler. That was put in for a very specify reason, without it there are deadlock scenarios with single threaded sync contexts. And while ASP.NET Core might not be used very often with one, WCF is used a lot in UI apps such as WPF so I'm expecting some moderate use with WPF/WinForms apps.

Daniel-Svensson:
@mconnew fixed your by comments by using OnCompleted code as in corefx.
SyncronizationContext is explicitly not captured.

Regarding the HideScheduler comment it should not have made any difference as the code was written since `TaskScheduler.Default` was *explicitly* specified it would have set `TaskScheduler.Current` to `TaskScheduler.Defatult` within the callback both with or without the *HideScheduler*. At least according to the documentation for the enum and [New TaskCreationOptions and TaskContinuationOptions in .NET 4.5](https://devblogs.microsoft.com/pfxteam/new-taskcreationoptions-and-taskcontinuationoptions-in-net-4-5/) . But better to be safe than sorry :)

mconnew:
@Daniel-Svensson, thank you for this change and your patience as we've discussed some of the finer details. I've validated these changes locally so I'm going to merge them now.

PR #62, opened 2019-10-18
Add property to control NuGet package version

Treit:
Currently creating NuGet packages  for CoreWCF just uses a default package version of "1.0.0". I'd like to publish the packages to our internal feed so we can use them for private testing, but want to make it clear these are pre-release versions and not an official 1.0 release. So, adding a property to control the version of the packages generated.


Daniel-Svensson:
The proposed CI/CD pipeline in #52 solves the same issue using [gitversion](https://gitversion.readthedocs.io/en/latest/) .

If it is important to have the same release numbers with when right clicking from within VS or dunning dotnet pack without specifying release number one could consider using the nuget package.

Directory.Build.props
```diff
@@ -32,5 +32,10 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
   </PropertyGroup>
   
+  <!-- Common NuGet package properties -->
+  <PropertyGroup>
+    <PackageVersion>0.0.1-alpha</PackageVersion>
```
mconnew: Can you make this conditional on the property currently being empty? That way it can be overridden on the command line which would be useful for automated build scripts. Without this change, would setting it on the command line with `/p:PackageVersion=0.0.1-alpha` work today?

Treit: Yes, I should have made it conditional on it not already being set. Fixed, and also verified `dotnet pack -p:PackageVersion=0.0.2` works as expected.

mconnew:
@Daniel-Svensson, this is a temporary thing until we have real packages being produced from devops. I had a look at GitVersion and it definitely looks like a good choice for managing versions in the long run.

Kralizek:
Hey!

If you want to have a look at GitVersion, you can have a look at what I have on some of my repos like [this one](https://github.com/nybus-project/nybus).

The repo uses AppVeyor (still working on getting Azure Pipelines to work in a similar way)...

PR #65, opened 2019-10-23
Implemented and wrote tests for using DI for all InstanceContextMode values

### LINKED ISSUE #56: Use Dependency injection to provide instance for all InstanceContextMode's
I need to inject a depency to a service. How do I do that? The service only supports parameterless constructor.

mconnew:
Additional work included:
* Cleaned up test namespaces
* Fixed Windows test targetframeworks for net472
* Fixed session cleanup in Dispatcher
Fixes #56


src/CoreWCF.Primitives/tests/DependencyInjection/ServiceInstanceContextModeTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        [Fact]
+        public static void InstanceContextMode_Single()
+        {
+            SingleInstanceContextSimpleService.CreationCount = 0;
+            SingleInstanceContextSimpleService.DisposalCount = 0;
+            var services = new ServiceCollection();
+            var serviceInstance = new SingleInstanceContextSimpleService();
+            services.AddSingleton(serviceInstance);
+            TestHelper.BuildDispatcherAndCallService<SingleInstanceContextSimpleService>(services);
```
Daniel-Svensson: Did you consider making 2 calls. so that the CreateCount would be 2 if it created multiple instances but still skipped Dispose ?

src/CoreWCF.Primitives/tests/DependencyInjection/ServiceInstanceContextModeTests.cs
```diff
@@ -0,0 +1,174 @@
...
+            services.AddTransient<PerCallInstanceContextSimpleService>();
+            string serviceAddress = "http://localhost/dummy";
+            var serviceDispatcher = TestHelper.BuildDispatcher<PerCallInstanceContextSimpleService>(services, serviceAddress);
+            IChannel mockChannel = new MockReplyChannel(services.BuildServiceProvider());
+            var dispatcher = serviceDispatcher.CreateServiceChannelDispatcher(mockChannel);
+            // Instance created as part of service startup to probe if type is availale in DI
+            Assert.Equal(1, PerCallInstanceContextSimpleService.CreationCount);
+            Assert.Equal(1, PerCallInstanceContextSimpleService.DisposalCount);
+
+            PerCallInstanceContextSimpleService.CreationCount = 0;
```
Daniel-Svensson: nit: test would be slightly easier to follow if you also reset DisposalCount

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelDispatcherBase.cs
```diff
@@ -10,10 +10,20 @@ public abstract class ChannelDispatcherBase : CommunicationObject
...
+        internal void AttachInternal(ServiceHostBase host)
+        {
+            Attach(host);
+        }
+
         protected virtual void Attach(ServiceHostBase host)
         {
         }
 
+        internal void DetachInternal(ServiceHostBase host)
```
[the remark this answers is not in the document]
mconnew: This is code copied from .NET Framework. The whole ChannelDispatcherBase/ChannelDispatcher thing needs tidying up. I believe it's an artifact left over for a back compat scenario from an earlier release of WCF. Unfortunately ChannelDispatcherBase bleeds into ChannelDistpatcherCollection public api so people might be relying on the specific type in some extensibility. At some point there's going to be some compat breaks with modernizing some things like make message inspectors async. That's the time to clean this up. Until then, I'm keeping it the same as .NET Framework to prevent having to think about repercussions of changes too much.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelDispatcherCollection.cs
```diff
@@ -16,7 +16,15 @@ internal ChannelDispatcherCollection(ServiceHostBase service, object syncRoot)
 
         protected override void ClearItems()
         {
-            throw new PlatformNotSupportedException();
+            ChannelDispatcherBase[] array = new ChannelDispatcherBase[Count];
+            CopyTo(array, 0);
+            base.ClearItems();
```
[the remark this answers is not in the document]
mconnew: ClearItems should only get called on service shutdown. There's still some cleanup work needed with cleaning everything up with the Kestrel based host is shutting down so this will tie into that. Basically the logic in ServiceHost.Close needs to be moved to the shutdown handler registered with ASP.NET Core.

src/CoreWCF.Primitives/src/CoreWCF/ServiceHostBase.cs
```diff
@@ -343,12 +371,41 @@ protected override void OnOpened()
...
         }
-        
+
+        internal void OnRemoveChannelDispatcher(ChannelDispatcherBase channelDispatcher)
+        {
+            lock (ThisLock)
+            {
+                ThrowIfClosedOrOpened();
+                channelDispatcher.DetachInternal(this);
+            }
```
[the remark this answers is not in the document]
mconnew: This code would only get called on service close/abort so it becomes irrelevant. This code is technically dead code unless someone calls it explicitly, but it's here as part of preparation for improved shutdown/cleanup code tied in to ASP.NET Core.

src/CoreWCF.Primitives/tests/DependencyInjection/ServiceInstanceContextModeTests.cs
mconnew: Good point. I'll add that. I was concentrating on making sure that the injected instance is the only instance created and making a single request is sufficient to show this. I think you can safely infer that it's doing the right thing with the existing code but explicit is better then inference if you can test something explicitly so I'll make the change.

src/CoreWCF.Primitives/tests/DependencyInjection/ServiceInstanceContextModeTests.cs
mconnew: not a nit, clear code is better code! The less you need to think about why something isn't how you would immediately expect, the better. I'll make the change.

mconnew:
@Daniel-Svensson, good call on adding a second call for Single mode. I found a bug by adding an extra call. It wasn't setting ServiceHostObjectModel.DisposableInstance when it implemented IDisposable which meant it got disposed after the first call. Then it created a new instance for the second call.

PR #69, opened 2019-10-30
Fix connection reuse problems

mconnew:
* Streamed transfer mode for NetTcp wasn't using our own RawStream implementation which can safely and cleanly unwrap an upgraded stream. This was preventing the socket from being reused for multiple requests as NegotiateStream still had a pending read on the raw connection.
* Fixed an issue when the socket is closed while trying to read the framing mode.
* Fixed tests to correctly close the channel factory which results in pooled connections being correctly closed. Added a helper to ensure the channel and factory are fully closed even if there's an exception.
* Added logger when building tests in Debug mode which outputs AspNetCore logs to ITestOutputHelper (xunit class) for capturing any error logging from AspNetCore.
* Added binding helper to configure the client bindings to have a 20 minute timeout for all timeouts when a debugger is attached. This is to help when stepping through code while debugging a test without having the connection aborted due to a timeout.


mconnew:
Latest update on this PR fixes the problems with running the tests in Azure DevOps.

mconnew:
@tibor19, ping?

PR #72, opened 2019-11-17
Make the ParameterInspectors collection on DispatchOperation public

Treit:
In order for developers writing WCF services to actually hook up their own custom parameter inspectors, the ParameterInspectors collection on DispatchOperation needs to be changed from internal to public.

This change also fixes the default version specified in the Directory.Build.props to apply to the assembly as well as the NuGet package when, for instance, running `dotnet pack`. Previously the version was applying to the NuGet package but the assembly version was always 1.0.0.


Directory.Build.props
```diff
@@ -33,9 +33,10 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
   </PropertyGroup>
   
-  <!-- Common NuGet package properties -->
+  <!-- Common version-related properties -->
   <PropertyGroup>
-    <PackageVersion Condition="'$(PackageVersion)'==''">0.0.1-alpha</PackageVersion>
+    <VersionPrefix Condition="'$(VersionPrefix)'==''">0.0.3</VersionPrefix>
```
mconnew: How does this behave if PackageVersion has been explicitly set?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperation.cs
```diff
@@ -188,7 +188,7 @@ public string Name
             get { return name; }
         }
 
-        internal SynchronizedCollection<IParameterInspector> ParameterInspectors { get; }
+        public SynchronizedCollection<IParameterInspector> ParameterInspectors { get; set; }
 
         public DispatchRuntime Parent { get; }
```
Daniel-Svensson: Does the setter provide any additional value? Or is it enough to be able to modify the collection?

Treit: You are right, the setter is not needed. I updated the code to remove it.

Directory.Build.props
Treit: The explicitly set PackageVersion will override these. So your package will have one version and the assembly inside the package will have the version from VersionPrefix and VersionSuffix. Seems like the right behavior.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperation.cs
mconnew: @Treit, I should have looked at this property more closely as it's an interesting one. The client corresponding equivalent of `DispatchOperation` is `ClientOperation`. At some point `ClientOperation` was modified to hide the `ParameterInspectors` property and added a new one called `ClientParameterInspectors` which was declared as `ICollection<IParameterInspector>`, ie:
```c#
        public ICollection<IParameterInspector> ClientParameterInspectors { get; }

        [EditorBrowsable(EditorBrowsableState.Never)]
        public new SynchronizedCollection<IParameterInspector> ParameterInspectors { get; }
```

I'm not suggesting a fully equivalent change here, but I do see there being value in changing the type of `DispatchOperation.ParameterInspectors` to be `ICollection<IParameterInspector>`. While that's technically a breaking change, moving to CoreWCF is technically one big breaking change anyway with the namespace change. I'm not seeing anything on `SynchronizedCollection` which isn't declared in one of the collection interfaces so I don't think this would cause any problems for anyone.
So I'm soft suggesting a change but asking for opinions. @Daniel-Svensson, what are your thoughts on my suggestion? The original comment was made against the wrong line to make a code suggestion so I'll make another comment with the suggestion so you can one-click apply.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperation.cs
```diff
@@ -188,7 +188,7 @@ public string Name
             get { return name; }
         }
 
-        internal SynchronizedCollection<IParameterInspector> ParameterInspectors { get; }
+        public SynchronizedCollection<IParameterInspector> ParameterInspectors { get; }
```
mconnew: This is my suggested change.
```suggestion
        public ICollection<IParameterInspector> ParameterInspectors { get; }
```

Directory.Build.props
mconnew: I'm still trying to understand the versioning landscape here. Is it typical for a package version to be different than the assembly version?  
If PackageVersion isn't explicitly set, does it automatically get the value VersionPrefix-VersionSuffix?

mconnew: Also I understand that you probably used 0.0.3 because of your own internal usage and you need this to be monotonically increasing, but in the public repo we haven't released anything yet so starting at 0.0.3 isn't the right thing.

Treit: Some NuGet packages have the same package version and assembly version, some don't. I seriously dislike not having them be in sync because given an assembly and knowing its version, I don't have to guess what version of the NuGet package it came from.

Yes if the PackageVersion is not explicitly set it gets the value VersionPrefix-VersionSuffix automatically.

Also, yes, I need the version to be increasing. Since the public repo isn't publishing a NuGet package yet does it even matter what the version is? I am publishing my own NuGet package for Core WCF since there isn't an official one yet. 

If you don't want my version, what would you like the version to be set to in the public repo?

mconnew: I suppose it can stay at 0.0.3 for now as we're not publishing anything but be aware that when I get the versioning stuff all hooked up with GitVersion, this will change.

Treit: Sure, I can't wait for the day that I don't have to publish my own package any longer.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperation.cs
Daniel-Svensson: Sounds ok with ICollection as long as SynchronizedCollection does not imply some very specific behavior (I expect it provides some thread safety?) in which case it might make more sense to add a comment about that behavior instead. I se no really strong arguments for changing or keeping as is. 

pro: I like that it mirrors the client behavior, should the name also be changed to mimic *Client*ParameterInspectors? (or is it be better of changing the client to only have a single public `ParameterInspectors`). 

cons: The downside I see is that you will not be able to implicitly cast to IReadOnly.. or ICollection interfaces, but that is a somewhat minor issue.

Directory.Build.props
Daniel-Svensson: I personally prefer the approach taken by EF where assembly version is kept at major version of the nuget package. But you should keep the informal version and file version in sync with the nuget and commit used (via sourcelink)

That way you keep binding redirects at a minimum and can always easily see what nuget version (and commit) corresponds to a dll

PR #73, opened 2019-11-27
Fix async services

### LINKED ISSUE #67: Calling async WCF method timeout
here is my service:
```csharp
    public class EchoService : Contract.IEchoService
    {
        public async Task<string> Echo(string text)
        {
            System.Console.WriteLine($"Received {text} from client!");
            await Task.Delay(1000);
            System.Console.WriteLine($"Received {text} from client 2!");
            return text;
        }
    }
```
and my client:
```csharp
            var factory = new ChannelFactory<IEchoService>(new BasicHttpBinding(), new EndpointAddress("http://localhost:8080/basichttp"));
            factory.Open();
            var channel = factory.CreateChannel();
            ((IClientChannel)channel).Open();
            while (true)
            {
                Console.WriteLine("Echo(\"Hello\") => " + await channel.Echo("Hello"));
                Thread.Sleep(1000);

            }
            ((IClientChannel)channel).Close();
            factory.Close();

            Console.WriteLine("Hit enter to exit");
            Console.ReadLine();
```
on server in console:
```
Received Hello from client!
Received Hello from client 2!
```
on client in console:

```
Unhandled Exception: System.TimeoutException: The request channel timed out attempting to send after 00:01:00. Increase the timeout value passed to the call to Request or increase the SendTimeout value on the Binding. The time allotted to this operation may have been a portion of a longer timeout. ---> System.TimeoutException: The HTTP request to 'http://localhost:8080/basichttp' has exceeded the allotted timeout of 00:01:00. The time allotted to this operation may have been a portion of a longer timeout.
```
 
This is working correctly in the previous version.
I downloaded the previous version on 13.10.2019 8:55

mconnew:
Also should fix running on Linux as were using some Windows only api's
Fixes #67


mconnew:
@tibor19, can you review this? After this I think it's ready to make some preview packages.

PR #74, opened 2019-12-03
Fix for "Why is XmlSerializerFormatAttribute internal?" #43

### LINKED ISSUE #43: Why is XmlSerializerFormatAttribute internal?
I have an existing WCF application that I have been looking to port to .NET Core for a while now. I am currently evaluating of this is feasible with this library, however the first thing I run into is the fact that XmlSerializerFormatAttribute is not supported?

The application I have needs to have this attribute in order to serialize the XML to conform to the existing standard. I already tried using the XmlSerializerFormatAttribute from System.ServiceModel.Primitives, but that does not work.
If I change the XmlSerializerFormatAttribute from internal to public (together with some enums to satisfy the compiler) it does exactly what I need, but I'm sure there is a reason why this attribute is not public.

comburo:
As discussed in issue #43.
Finally got approval to support open source projects.
Fixes #43

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/OperationFormatStyle.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/OperationFormatStyle.cs
@@ -1,6 +1,6 @@
 ﻿namespace CoreWCF
 {
-    internal enum OperationFormatStyle
+    public enum OperationFormatStyle
     {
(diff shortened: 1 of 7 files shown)


mconnew:
Thank you. I really appreciate you providing the tests too.

PR #75, opened 2019-12-05
Add S.SM Binding which dispatches to CoreWCF to make unit testing easier

mconnew:


mconnew:
@tibor19, could you review these changes? This should make it easier to add more test variations.

src/CoreWCF.Primitives/tests/DispatcherClient/DispatcherChannelFactory.cs
```diff
@@ -0,0 +1,153 @@
...
+                _ = GetServiceChannelDispatcher(channel);
+                return (TChannel)(object)channel;
+            }
+
+            //if (typeof(TChannel) == typeof(IDuplexSessionChannel))
+            //{
+            //    return (TChannel)(object)new DispatcherDuplexSessionChannel((DispatcherChannelFactory<IDuplexSessionChannel>)(object)this);
+            //}
+
+            throw new InvalidOperationException("Only support IRequestChannel and IDuplexSessionChannel");
```
tibor19: It should be "Only support IRequestChannel and IRequestSessionChannel", not "Only support IRequestChannel and IDuplexSessionChannel"

src/CoreWCF.Primitives/tests/DispatcherClient/DispatcherTransportBindingElement.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (typeof(TChannel) == typeof(IRequestChannel) ||
+                typeof(TChannel) == typeof(IRequestSessionChannel))
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public override string Scheme => "COREWCF";
```
tibor19: Shouldn't Scheme be "corewcf"(small letters), for consistency purposes?

src/CoreWCF.Primitives/tests/Helpers/TestHelper.cs
```diff
@@ -1,16 +1,11 @@
...
+using CoreWCF.Channels;
 using System;
-using System.Collections.Generic;
+using System.IO;
 using System.Text;
-using System.Threading;
 using System.Xml;
 
-namespace CoreWCF.Primitives.Tests.Helpers
+namespace Helpers
```
tibor19: What is the rationale of making Helpers a "top level" namespace?

mconnew: None of these api's are being consumed by other libraries and the really long namespace just seems a little obnoxious to me. The purpose of the namespace is to prevent things like System.ServiceModel.ChannelFactory from being confused with CoreWCF.ChannelFactory when end user code is referencing both assemblies but that's not the case here.

src/CoreWCF.Primitives/tests/DispatcherClient/DispatcherTransportBindingElement.cs
mconnew: You're right. For some reason I had in my head that the scheme was uppercase in other transports but I was wrong.

src/CoreWCF.Primitives/tests/Helpers/TestHelper.cs
mconnew: If you feel strongly about it, let me know and I'll put this back and follow up with an other PR to update all the other namespaces to have the project name in them too.

mconnew: @tibor19, if you want this changed. you can add a comment post merge and I'll open a PR to fix all the top level namespaces. For now I'm going to merge this with the other things fixed.

PR #76, opened 2019-12-11
Create packages and publish to CoreWCF feed

mconnew:
Enable package creation in Azure Devops feed.


mconnew:
I'll work out how to get these packages signed and published to nuget at a later date.

PR #78, opened 2019-12-17
Adding code coverage to windows release test runs

mconnew:


tibor19:
@mconnew Can you fix the error? It seems that the task cannot find dotnet.

mconnew:
The problem isn't with being able to find dotnet. An earlier task installs a dotnet global tool (ReportGenerator), then it tries to run the reportgenerator and it's unable to find the global tool. So I need to work out what's going wrong there or try a different approach. It might need the tool being installed as a local tool instead of global, but then it needs to be executed in the context of a project.

mconnew:
So apparently there's a type of global tool you can install which doesn't need to be executed using the dotnet command. Calling `dotnet reportgenerator` was looking for a global tool installed with the name dotnet-reportgenerator. If you run `dotnet tool list -g`  you can see the tool is installed with the name `reportgenerator`. But there's an .exe installed in `%USERPROFILE%\.dotnet\tools` with the filename `reportgenerator.exe` so I just had to call it directly. Now I'm hitting a problem where the code coverage report isn't being generated for some reason.

mconnew:
@tibor19, can you have a look through this and see if there's anything looking out of place? The code coverage can be seen here: https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=26382&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab

mconnew:
@tibor19, I'm going to merge this to get coverage report information as part of the builds. If there's anything you see that you want changed, feel free to do a post-merge review and I can address them with another PR if needed.

PR #87, opened 2019-12-31
Add compatibility for using S.SM contract attributes

mconnew:
Runtime conversion of S.SM.[Service|Operation]ContractAttribute applied to contract interface to the CoreWCF conterparts to enable compat.


mconnew:
The code in DispatchHelper creates a ChannelFactory using the special dispatcher binding. Previously it was an Extensibility helper as it was used to helper with the Extensibility tests. But now it has broader use it's a DispatchHelper as it helps create a ChannelFactory which uses the dispatcher. The change it to reflect what it does rather than what the consumers do.

PR #97, opened 2020-01-30
Adding HttpRequestMessageProperty tests

mconnew:
Also fixing the API for `HttpRequestMessageProperty` to match .NET Framework better. I had switched the headers to use `System.Net.Http.Headers.HttpHeaders` to migrate away from `WebHeaderCollection` and use the newer way of doing things. Turns out ASP.NET Core decided to invent a new way of doing headers and invented an interface called `IHeaderDictionary`. Reverting back to WebHeaderCollection and will consider what's a better way to do this moving forward as there are multiple options.


mconnew:
Merging as this is blocking someone's migration.

PR #112, opened 2020-03-18
Make IOperationInvoker simpler with ValueTask and expose

### LINKED ISSUE #106: Make DispatchOperation Invoker public
I need to "customize" the Invoker process for a DispatcherOperation but it's internal in CoreWCF implementation.
I'm asking to make it "public"

mconnew:
Fixes #106

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
@@ -505,8 +505,6 @@ namespace CoreWCF.Description
                 }
             }
 
-            BindOperations(contractDescription, null, dispatch);
-
             //dispatcher.SetSupportedChannels(DispatcherBuilder.GetSupportedChannelTypes(contractDescription));
             int filterPriority = 0;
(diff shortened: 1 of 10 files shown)


mconnew:
@birojnayak, would you be able to review this change?

birojnayak:
sure.. unless you are in a hurry to push the changes.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -505,8 +505,6 @@ private static Type[] GetSupportedChannelTypes(ContractDescription contractDescr
                 }
             }
 
-            BindOperations(contractDescription, null, dispatch);
```
birojnayak: what was the reason for call and why did you remove ?

mconnew: This call causes the behavior OperationInvokerBehavior to be applied which sets DispatchOperation.Invoker. This method was being called twice which normally isn't a problem, unless you try to replace the Invoker. In which case applying OperationInvokerBehavior a second time causes it to repalce the Invoker with the default implementation again. It was a bug applying this twice.

Issue #113, opened 2020-03-19
NetHttpBinding: quotas can't be changed


0x53A:
I am porting a service to CoreWCF / NetHttp and so far it looks good, except for one issue:

> CoreWCF.QuotaExceededException: The maximum message size quota for incoming messages (65536) has been exceeded. To increase the quota, use the MaxReceivedMessageSize property on the appropriate binding element.

I _did_ set the property, but the setter is empty

https://github.com/CoreWCF/CoreWCF/blob/ebb95c7e7d397c843807ea1322e34946a15fd098/src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs#L30-L31

0x53A:
I'm currently testing https://github.com/CoreWCF/CoreWCF/pull/114, so far it seems to work.

PR #114, opened 2020-03-19
pass through quotas for the NetHttpBinding

0x53A:
ref https://github.com/CoreWCF/CoreWCF/issues/113

This is what I changed locally to get it to work. Not quite sure what to do about the BasicHttpBinding.

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/BasicHttpBinding.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/BasicHttpBinding.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Xml;
 using CoreWCF.Channels;
 
 namespace CoreWCF
@@ -16,6 +17,7 @@ namespace CoreWCF
             _basicHttpSecurity.Mode = securityMode;
         }
 
+        public override XmlDictionaryReaderQuotas ReaderQuotas { get => null; set { } }
         internal WSMessageEncoding MessageEncoding { get; set; } = BasicHttpBindingDefaults.MessageEncoding;
 
         internal override BasicHttpSecurity BasicHttpSecurity => _basicHttpSecurity;
--- a/src/CoreWCF.Http/src/CoreWCF/BasicHttpBinding.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/BasicHttpBinding.cs
@@ -17,7 +17,6 @@ namespace CoreWCF
             _basicHttpSecurity.Mode = securityMode;
         }
 
-        public override XmlDictionaryReaderQuotas ReaderQuotas { get => null; set { } }
         internal WSMessageEncoding MessageEncoding { get; set; } = BasicHttpBindingDefaults.MessageEncoding;
 
         internal override BasicHttpSecurity BasicHttpSecurity => _basicHttpSecurity;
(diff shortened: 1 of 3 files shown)


mconnew:
This isn't quite the correct solution. The full implementation should live on HttpBindingBase. You can find the original code [here](https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/HttpBindingBase.cs,52817be0254871f7). That way you don't need to override it. That code sets it on the text message encoder too so it will apply for BasicHttpBinding too.

0x53A:
Okay I tried to adapt it to be similar to the reference source.

mtom doesn't (yet) seem to be implemented in CoreWCF, whatever it is, and HttpBindingBase._httpsTransport is null, so I just commented that code also out.

mconnew:
I realized this still isn't going to get the job done for you. The SetReaderQuotas method isn't currently being called. I'm working on a set of changes to fix that. The reader quota's are just set on the binary message encoder so today you could just do this:
```c#
var netHttpBinding = new NetHttpBinding();
var customBinding = new CustomBinding(netHttpBinding);
var bmebe = customBinding.Elements.Find<BinaryMessageEncodingBindingElement>();
bmebe.ReaderQuotas = myReaderQuotas;
```
Then use customBinding for your endpoint. 
There's an unrelated failing test (InstanceProviderReleaseCalledWithCorrectObjectTest) which is a race condition in the test showing up because of some recent performance improvements so I'm going to merge this as it takes us a step closer even though it doesn't solve your problem fully.

0x53A:
> [...] The SetReaderQuotas method isn't currently being called. [...]

It is called in the setter of HttpBindingBase.ReaderQuotas, so I think it should actually work right now as is, but I didn't test it tbh.

My service seems to work for now, thank you for this project!

absinghal:
There seems to be another issue in NetHttpBinding with BinaryEncoding

The following code runs before _maxReceivedMessageSize is set by SetMaxReceivedMessageSizeFromTransport

``` var httpSettings = new HttpTransportSettings
            {
                BufferManager = BufferManager.CreateBufferManager(DefaultMaxBufferPoolSize, tbe.MaxBufferSize),
                OpenTimeout = _serviceDispatcher.Binding.OpenTimeout,
                ReceiveTimeout = _serviceDispatcher.Binding.ReceiveTimeout,
                SendTimeout = _serviceDispatcher.Binding.SendTimeout,
                CloseTimeout = _serviceDispatcher.Binding.CloseTimeout,
                MaxBufferSize = tbe.MaxBufferSize,
                MaxReceivedMessageSize = tbe.MaxReceivedMessageSize,
                MessageEncoderFactory = mebe.CreateMessageEncoderFactory(),
                ManualAddressing = tbe.ManualAddressing,
                TransferMode = tbe.TransferMode,
                KeepAliveEnabled = tbe.KeepAliveEnabled,
                AnonymousUriPrefixMatcher = new HttpAnonymousUriPrefixMatcher(),
                AuthenticationScheme = tbe.AuthenticationScheme,
                WebSocketSettings = tbe.WebSocketSettings.Clone()
            };
```

So the factory is initialized with MaxReceivedMessageSize =0
and so BinaryMessageEncoder also has  _maxReceivedMessageSize =0 
So this combination does not work

As a workaround for now, have set _maxReceivedMessageSize using reflection in my application

PR #118, opened 2020-04-04
fix for service auth behavior

birojnayak:


src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
```diff
@@ -112,8 +112,7 @@ public bool ImpersonateCallerForAllOperations
             }
             set
             {
-                ThrowIfImmutable();
-                impersonateCallerForAllOperations = value;
```
mconnew, line 116: Can you leave the old code commented like you did below so it's obvious what should be here. Does impersonation not work? I put some effort in to making impersonation work with different api's now.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/AuthorizationContext.cs
```diff
@@ -14,8 +14,8 @@ internal abstract class AuthorizationContext : IAuthorizationComponent
 
         public static AuthorizationContext CreateDefaultAuthorizationContext(IList<IAuthorizationPolicy> authorizationPolicies)
         {
-            //return SecurityUtils.CreateDefaultAuthorizationContext(authorizationPolicies);
-            throw new PlatformNotSupportedException();
+            return SecurityUtils.CreateDefaultAuthorizationContext(authorizationPolicies);
+            //throw new PlatformNotSupportedException();
         }
```
mconnew, line 19: No need to leave the throw behind in a comment.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/DefaultAuthorizationContext.cs
```diff
@@ -0,0 +1,68 @@
...
+            this.properties = evaluationContext.Properties;
+        }
+
+        public static DefaultAuthorizationContext Empty
+        {
+            get
+            {
+                /* commenting 
+                if (LocalAppContextSwitches.EnableCachedEmptyDefaultAuthorizationContext)
+                {
```
mconnew, line 30: You can remove this commented code. This was put there as a back compat switch to revert the behavior of a fix. The problem was custom code was incorrectly modifying the empty default authotization context which resulted in leaking auth info from one call to another. The fix was to return a new one each time. My preferred fix at the time was to make it immutable and throw if trying to modify but there was concern that would be an even bigger breaking change. What are your thoughts? Should we make this an immutable singleton or allocate a new one each time?

src/CoreWCF.Http/tests/ServiceAuthBehaviorTest.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        internal class Startup
+        {
+            public void ConfigureServices(IServiceCollection services)
+            {
+                services.AddServiceModelServices();
+                var authBehavior = new ServiceAuthorizationBehavior();
+                services.AddSingleton<IServiceBehavior>(authBehavior);
+                services.AddSingleton(authBehavior);
```
mconnew, line 72: Should we add this by default in `services.AddServiceModelServices()`? What do you think about enabling people to do `services.AddSingleton<ServiceAuthorizationManager, MyTestServiceAuthorizationManager>()` and have it auto-wire in to the ServiceAuthorizationBehavior?

src/CoreWCF.Http/tests/ClientContract/IEchoService.cs
```diff
@@ -30,5 +30,8 @@ public interface IEchoService
             ReplyAction = Constants.OPERATION_BASE + "EchoStreamAsyncResponse")]
         Stream EchoStreamAsync(Stream echo);
 
+        [OperationContract(Name = "EchoStringToFail", Action = Constants.OPERATION_BASE + "EchoStringToFail",
+     ReplyAction = Constants.OPERATION_BASE + "EchoStringResponse")]
+        string EchoToFail(string echo);
```
mconnew, line 35: ReplyAction should be EchoStringToFailResponse. This works on HTTP, but would likely break on NetTcp because it's duplex.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
birojnayak: Will write a test to see if it works

birojnayak:
Updated review comments and added more functionalities. NetTCP for linux needs to be fixed (when security is transport), so made few tests Windows only. Once I fix the NetTCP for Linux will enable tests for Linux env.

src/CoreWCF.NetTcp/tests/Services/TestService.cs
```diff
@@ -50,5 +54,34 @@ public TestMessage TestMessageContract(TestMessage testMessage)
...
             };
         }
+
+       
+        public string EchoForPermission(string echo)
+        {
+            IPrincipal principal = Thread.CurrentPrincipal;
+            if (echo.Contains(PrincipalPermissionMode.UseWindowsGroups.ToString()))
+            {
+                if (principal.Identity !=null
```
mconnew, line 64: Nit: Code formatting, missing space after =

src/CoreWCF.NetTcp/tests/Services/TestService.cs
```diff
@@ -50,5 +54,34 @@ public TestMessage TestMessageContract(TestMessage testMessage)
                 Body = new MemoryStream(Encoding.UTF8.GetBytes(text + " from server"))
             };
         }
+
+       
+        public string EchoForPermission(string echo)
+        {
+            IPrincipal principal = Thread.CurrentPrincipal;
+            if (echo.Contains(PrincipalPermissionMode.UseWindowsGroups.ToString()))
+            {
```
mconnew, line 63: would nameof(PrincipalPermissionMode.UseWindowsGroups) work here?

src/CoreWCF.NetTcp/tests/Services/TestService.cs
```diff
@@ -50,5 +54,34 @@ public TestMessage TestMessageContract(TestMessage testMessage)
...
+                if (principal.Identity !=null
+                     && principal.Identity.GetType().Equals(typeof(WindowsIdentity)))
+                    return echo;
```
mconnew, line 66: It might be better to use `typeof(WindowsIdentity).IsAssignableFrom(principal.Identity?.GetType())`. That way we don't break if NegotiateStream starts returning a type derived from WindowsIdentity and you can remove the null check.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntime.cs
```diff
@@ -403,7 +403,7 @@ internal async Task<MessageRpc> InvokeAsync(MessageRpc rpc)
                         SetClaimsPrincipalToOperationContext(rpc);
                     }
 
-                    if (parent.SecurityImpersonation?.IsSecurityContextImpersonationRequired(rpc) ?? false)
+                    if (parent.SecurityImpersonation !=null)
                     {
```
mconnew, line 407: Nit, formatting

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/DefaultAuthorizationContext.cs
```diff
@@ -0,0 +1,61 @@
...
+            this.expirationTime = evaluationContext.ExpirationTime;
+            this.properties = evaluationContext.Properties;
+        }
+
+        public static DefaultAuthorizationContext Empty
+        {
+            get
+            {
+               if(empty == null)
+                    empty = new DefaultAuthorizationContext(new DefaultEvaluationContext());
```
mconnew, line 29: This code is still using a cached single DefaultAuthorizationContext. We need to either return an immutable context or not cache a single instance. As this code currently is written, a badly written authorization manager can leak claims between calls.

mconnew:
Just a couple of minor issues to be addressed, then I'll merge.

src/CoreWCF.NetTcp/tests/Services/TestService.cs
FreeAndNil: Why not "is WindowsIdentity"?

mconnew: That works too 😄

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/DefaultAuthorizationContext.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.expirationTime = evaluationContext.ExpirationTime;
+            this.properties = evaluationContext.Properties;
+        }
+
+        public static DefaultAuthorizationContext Empty
+        {
+            get
+            {
+               return new DefaultAuthorizationContext(new DefaultEvaluationContext()); ;
+            }
```
mconnew, line 28: Nit: extra colon at end of line

Issue #119, opened 2020-04-05
HttpBindings don't work for minimal app without UseUrls


0x53A:
If you have a minimal asp.net core application without either

```c#
.UseUrls(...)
```

or

```c#
.UseKestrel(options => { options.ListenLocalhost(<port>);}
``` 

then at initialization time, ``app.ServerFeatures.Get<IServerAddressesFeature>().Addresses`` is empty.

This will cause CoreWCF to throw later:

```
System.InvalidOperationException: 'Could not find a base address that matches scheme http for the endpoint with binding NetHttpBinding. Registered base address schemes are [].'
```

-----------

I found two possible solutions:

1) add ``http://localhost:5000`` as default

In src\CoreWCF.Http\src\CoreWCF\Channels\ServiceModelHttpMiddleware.cs ``BuildBranch`` just add 

```c#
            if (!serviceBuilder.BaseAddresses.Any())
            {
                // https://github.com/aspnet/Announcements/issues/224
                // The WebHost will no longer add the default server address of http://localhost:5000 to the IServerAddressesFeature when none is specified.
                serviceBuilder.BaseAddresses.Add(new Uri("http://localhost:5000"));
                serviceBuilder.BaseAddresses.Add(new Uri("https://localhost:5001"));
            }
```

But this is a wild guess that may or may not be correct.

<br />
<br />
  
-----------
  

2) delay CoreWCF initialization until server has been started

After kestrel has been started, ``IServerAddressesFeature.Addresses`` WILL be populated correctly.

In the same file, ``private readonly RequestDelegate _branch;`` could be changed to ``private readonly Lazy<RequestDelegate> _branch;`` and it seems to work.

------------

Note, there is another issue, where if you call just ``UseNetTcp`` without ``UseUrls``, then it will listen __only__ on the nettcp port, and no longer on 5000/5001.

mconnew:
I've got a fix in the works. The problem is CoreWCF needs to know the URL at service setup as it is used as a parameter for getting tokens from a token provider. But ASP.NET Core doesn't populate the features until after Kestrel is running. I have a fix which lazily creates the service, but I'm getting an intermittent 404 response when testing. Once I've tracked down why I sometimes get a 404, I'll push the fix.

mconnew:
I've been rethinking this and I think my approach is wrong. It shouldn't be lazily created, but retrieving the address can be lazily done. I'm going to revert some of my changes and make the delay happen at the point of use.

[...]

PR #120, opened 2020-04-05
Allow Kestrel to listen on a random port by delaying CoreWCF initialization

0x53A:
This implements option 2 of https://github.com/CoreWCF/CoreWCF/issues/119

There is one big downside: If there is a configuration error in CoreWCF, then it no longer fails at startup time, but when the first request is made.

So I opened this Pull Request more for a discussion ...

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
@@ -67,7 +67,7 @@ namespace CoreWCF.Channels
 
             _httpSettings = httpSettings;
             WebSocketOptions = CreateWebSocketOptions(tbe);
-            if (WebSocketOptions == null)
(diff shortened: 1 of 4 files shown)


mconnew:
I'd like to take a slightly different approach to this. I've now made CoreWCF port and protocol agnostic for HTTP. This is to match the WCF behavior on IIS where you can use something like BasicHttpBinding and have the site bound to HTTPS and HTTP. Because HTTP and HTTPS run on different ports, I had to make it agnostic to the port number for HTTP, and to accept HTTP or HTTPS for the scheme. The only further change now is when there's an http based transport configured, to inject a dummy base address of http://localhost/ into the IServiceBuilder to keep the service creation happy.
You have a couple of good changes in your PR I'd like to incorporate, as well as your tests so I'm going to leave this PR open until I do the work to finish this so that it doesn't disappear off my radar.

mconnew:
Closing this PR as the desired capability was implemented via another PR. Thanks for prompting me to "do it right" and not accepting the annoyances.

PR #123, opened 2020-04-07
HttpBindingBase contructor

edberg:
Summary of the changes
 - The HttpsTransportBindingElement was never initialized which causes exceptions when using BasicHttpSecurityMode.Transport.


0x53A:
There are a few places where _https is commented out like

                //_httpsTransport.MaxReceivedMessageSize = value;

PR #131, opened 2020-05-12
update to core 3.1 and fix nettcpbinding for linux

birojnayak:
initial PR for making nettcpbinding  work in linux(using kerbos) and added support for core 3.1.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
@@ -13,6 +13,8 @@ using System.Security.Authentication;
 using System.Security.Principal;
 using System.Threading;
 using System.Threading.Tasks;
+using CoreWCF.Primitives.CoreWCF.IdentityModel.Selectors;
+using CoreWCF.Primitives.CoreWCF.IdentityModel.Tokens;
 
 namespace CoreWCF.Channels
 {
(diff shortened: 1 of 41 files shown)


mconnew:
.NET Core 3.1 already supports netstandard 2.0. I'm not seeing any conditional code, why are you adding netcoreapp3.1 as a target? A netstandard2.0 library will work on .NET Core 3.1 just fine, in fact all the tests run on it already.  
Also, don't upgrade to ASP.NET Core 2.**2**. The LTS version which is supported on .NET Framework and will be supported for a long time is 2.**1**. ASP.NET Core 2.**2** is now unsupported and reached EOL in December 2019.

src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -163,17 +163,24 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
...
+                {
+                    WindowsIdentity windowIdentity = (WindowsIdentity)remoteIdentity;
+                    Security.SecurityUtils.ValidateAnonymityConstraint(windowIdentity, false);
+                    WindowsSecurityTokenAuthenticator authenticator = new WindowsSecurityTokenAuthenticator(extractGroupsForWindowsAccounts);
+                    token = new WindowsSecurityToken(windowIdentity, SecurityUniqueId.Create().Value, windowIdentity.AuthenticationType);
+                    authorizationPolicies = authenticator.ValidateToken(token);
+                }
+                else
+                {
+                    token = new GenericSecurityToken(remoteIdentity.Name, TokenImpersonationLevel.Identification, parent.ServerCredential, SecurityUniqueId.Create().Value);
```
mconnew, line 179: Why are you passing parent.ServerCredential and TokenImpersonationLevel? GenericSecurityToken should only contain the things relevant to a GenericIdentity.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/GenericSecurityToken.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace CoreWCF.IdentityModel.Tokens
+{
+    public class GenericSecurityToken : SecurityToken
+    {
+        private string _id;
+        private DateTime _effectiveTime;
+        private DateTime _expirationTime;
+
+        internal GenericSecurityToken(string servicePrincipalName, TokenImpersonationLevel tokenImpersonationLevel, NetworkCredential networkCredential, string id)
```
mconnew, line 16: Change servicePrincipalName to just name as SPN is Kerberos specific. I don't think you need the impersonation level either.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/GenericSecurityToken.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            get { return _effectiveTime; }
+        }
+
+        public override DateTime ValidTo
+        {
+            get { return _expirationTime; }
+        }
+
+        public string ServicePrincipalName { get; }
```
mconnew, line 57: Rename to Name as it just represents the GenericIdentity

PR #132, opened 2020-05-13
Adding support to enable Https usage

### LINKED ISSUE #32: SSL/TLS support for https with BasicHttpBinding?
Does CoreWCF have SSL/TLS support when using the BasicHttpBinding?

mconnew:
Fixes #32

### DIFF
--- a/src/CoreWCF.Http/tests/HttpsService.cs
+++ b/src/CoreWCF.Http/tests/HttpsService.cs
@@ -1,0 +1,57 @@
+﻿using CoreWCF.Configuration;
+using Helpers;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CoreWCF.Http.Tests
+{
+    public class HttpsService
+    {
+        private ITestOutputHelper _output;
+
+        public HttpsService(ITestOutputHelper output)
+        {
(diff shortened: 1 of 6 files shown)


src/CoreWCF.Http/tests/HttpsService.cs
```diff
@@ -0,0 +1,57 @@
...
+            string testString = new string('a', 3000);
+            var host = ServiceHelper.CreateHttpsWebHostBuilder<Startup>(_output).Build();
+            using (host)
+            {
+                host.Start();
+                var httpsBinding = ClientHelper.GetBufferedModeHttpsBinding();
+                var factory = new System.ServiceModel.ChannelFactory<ClientContract.IEchoService>(httpsBinding,
+                    new System.ServiceModel.EndpointAddress(new Uri("https://localhost:8443/BasicWcfService/basichttp.svc")));
+                factory.Credentials.ServiceCertificate.SslCertificateAuthentication = new System.ServiceModel.Security.X509ServiceCertificateAuthentication();
+                factory.Credentials.ServiceCertificate.SslCertificateAuthentication.CertificateValidationMode = System.ServiceModel.Security.X509CertificateValidationMode.None;
```
birojnayak, line 33: should we add more tests in future ?

mconnew: Yes, for completeness. Once we can dispatch a call to an HTTPS endpoint successfully, all calls to an HTTPS endpoint should work. But it's worth adding more variations to validate that.

PR #133, opened 2020-05-13
added fix for net tcp binding linux

birojnayak:


src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/ClaimComparer.cs
```diff
@@ -105,7 +105,8 @@ public static IEqualityComparer<Claim> Upn
         {
             get
             {
-                throw new PlatformNotSupportedException();
+                return Default;
```
mconnew, line 108: Can you remove the commented out code and add a comment about this being different behavior than on .NET Framework due to UpnComparer having a dependency on NTAccount which isn't available on Core.

mconnew:
Just need the comment cleanup then this will be ready to merge

PR #138, opened 2020-06-04
Add FaultContractName Test.

v-fangfc:
Abstract: Use [FaultContract(typeof(T))] on a variety of contracts and operation types and set the Name property


v-fangfc:
@zhaodongtian

v-fangfc:
@mconnew @HongGit

PR #151, opened 2020-06-19
Port .NetFramework IContractBehavior basic tests

imcarolwang:


src/CoreWCF.Http/tests/Services/ContractBehaviorSerivce.cs
```diff
@@ -0,0 +1,86 @@
...
+using ServiceContract;
+
+namespace Services
+{
+    public abstract class IContractBehaviorBasic_ServiceBase
+    {
+        public string ServiceContractMethod(string s)
+        {
+            //TODO: validate behavior
+            //BehaviorInvokedVerifier.ValidateServiceInvokedBehaviors(OperationContext.Current.Host.Description);
```
imcarolwang, line 12: @mconnew This case is testing IContractBehavior, when validating behavior for service, the below object is unavailable: `OperationContext.Current.Host.Description`. There's `ServiceHostBase host `in `CoreWCF.OperationContext` and it is private, is there any plan to expose the property? If not do you have any suggestion on how to validate the behaviors after service be hosted?

mconnew: I do plan to expose the property. I've tried not to expose things unless they are tested. Could you make that property public as this test indirectly tests it.

imcarolwang: Updated: add public property OperationContext.Current.Host for test.

src/CoreWCF.Http/tests/ContractBehaviorTests.cs
```diff
@@ -0,0 +1,311 @@
...
+            {
+                app.UseServiceModel(builder =>
+                {
+                    switch(_method)
+                    {                     
+                        case "ByHand":
+                        case "ByHand_UsingHiddenProperty":
+                        case "ByHandImplementsOther":
+                        case "ByHandImplementsOther_UsingHiddenProperty":
+                            builder.AddService<ContractBehaviorBasic_ByHand_Service>();
```
imcarolwang, line 281: @mconnew , The original .NetFramework test case contains "byhand" test scenarios which added CustomContractBehaviorAttribute by hand via calling method below, do you have any idea on how to do similar config in CoreWCF?

ServiceHost.Description.Endpoints[0].Contract.ContractBehaviors.Add(cb);

mconnew: I've just brought forward and committed a change from some other work to enable this. Check out [this test](https://github.com/CoreWCF/CoreWCF/blob/2e98274e6b50cef235cea4ee5d921eeea663eafe/src/CoreWCF.Http/tests/SimpleTest.cs#L93) to see how to add a delegate which gets access to ServiceHostBase.Description. It's executed just before the equivalent of ServiceHost.Open().

imcarolwang: Thank you. I'll checkout that and update test.

src/CoreWCF.Http/tests/ContractBehaviorTests.cs
```diff
@@ -0,0 +1,311 @@
...
+                                serviceHost.Description.Endpoints[0].Contract.ContractBehaviors.Add(cb);
+                            });
+                            break;
+                        case "ByHandImplementsOther":
+                            builder.AddService<ContractBehaviorBasic_ByHand_Service>();
+                            builder.AddServiceEndpoint<ContractBehaviorBasic_ByHand_Service, ServiceContract.IContractBehaviorBasic_ByHand>(new BasicHttpBinding(), "/BasicWcfService/ContractBehaviorService.svc");
+                            builder.ConfigureServiceHostBase<ContractBehaviorBasic_ByHand_Service>(serviceHost =>
+                            {
+                                var cb = new ServiceContract.MyMultiFacetedBehaviorAttribute();
+                                serviceHost.Description.Endpoints[0].Contract.ContractBehaviors.Add(cb);
```
imcarolwang, line 281: @mconnew  now I am able to add custom ContractBehaviors this way ("by hand"), but test failed when validating the behavior implementation, the code [here](https://github.com/CoreWCF/CoreWCF/blob/8e24169c70fcf83e4be6a6e9602fd5d3b66dfa93/src/CoreWCF.Http/tests/ServiceContract/IContractBehavior.cs#L63) didn't execute.  

The other test variation which use the custom behavior as an attribute decorating the contract interface works, the ApplyDispatchBehavior(...) implementation is called with test flag being set as expected.

imcarolwang: The affected "byhand" test scenario validation has been skipped, tracking issue: #193.

mconnew:
@imcarolwang, this is failing on test compilation:
```
src\CoreWCF.Primitives\tests\MessageEncoderTest.cs(36,25): Error CS0234: The type or namespace name 'MessageTestUtilities' does not exist in the namespace 'CoreWCF.Primitives.Tests.Helpers' (are you missing an assembly reference?)
```
Additionally there's an xunit warning:
```
src\CoreWCF.Primitives\tests\MessageTests.cs(28,13): Warning xUnit2009: Do not use Assert.True() to check for substrings.
```
I think you can use `Assert.Contains` but I haven't checked.

mconnew:
@imcarolwang, I've added you as a contributor so you should be able to run `/azp run` now. If it still doesn't work for you, let me know.

imcarolwang:
> @imcarolwang, I've added you as a contributor so you should be able to run `/azp run` now. If it still doesn't work for you, let me know.

It still doesn't work, I tried it in PR #190. I see members in our team are all in the Contributors and none of us have the permission.

mconnew:
Did you accept my invite to the CoreWCF organization?

imcarolwang:
Yes, I clicked the invitation link from email and it was redirected to the CoreWCF repository site, which seems I've accepted it.

PR #153, opened 2020-06-24
Add interesting type test port from NetFramework.

imcarolwang:
@honggit, @mconnew .


src/CoreWCF.Http/tests/ClientContract/IMarshalledTypeService.cs
```diff
@@ -0,0 +1,51 @@
...
+
+		[OperationContract]
+		MarshalledType TwoWayMethod2();
+
+		[OperationContract]
+		MarshalledType TwoWayMethod3(MarshalledType t);
+	}
+
+	[DataContract(Namespace = "http://Microsoft.ServiceModel.Samples")]
+	public class MarshalledType : MarshalByRefObject
```
imcarolwang, line 22: Hi @mconnew , the MarshalledTypeTest case passed on .NET472 but failed on .NETCore due to error below. Is the behavior by design on .NETCore?

   

>  System.Runtime.Serialization.InvalidDataContractException : Type 'ClientContract.MarshalledType' cannot inherit from a type that is not marked with DataContractAttribute or SerializableAttribute.  **Consider marking the base type 'System.MarshalByRefObject' with DataContractAttribute or SerializableAttribute**, or removing them from the derived type.
>   Stack Trace: 
>     ClassDataContractCriticalHelper.ctor(Type type)
>     ClassDataContract.ctor(Type type)
>     DataContractCriticalHelper.CreateDataContract(Type type)
>     DataContractCriticalHelper.CreateDataContract(Int32 id, RuntimeTypeHandle typeHandle, Type type)
>     DataContractCriticalHelper.GetDataContractSkipValidation(Int32 id, RuntimeTypeHandle typeHandle, Type type)
>     DataContract.GetDataContract(RuntimeTypeHandle typeHandle, Type type, SerializationMode mode)
>     DataContractSerializer.get_RootContract()

mconnew: Yes, this is a change in behavior in .NET Core. The class MarshalByRefObject is only there to satisfy classes that derive from it to be able to compile. None of it's functionality is there, including being serializable. Basically this test isn't applicable to .NET Core. I don't think there's any benefit keeping the test for Net472 as that support is only there to enable a migration path to .NET Core.

imcarolwang: Thanks for the confirm, Matt. I have removed the related case in the PR.

mconnew:
All tests did actually pass. [Looking at the test run](https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=37242&view=results) on Azure DevOps shows everything passed so I don't know why this is showing otherwise.

PR #156, opened 2020-06-30
Connection header fix + X509 security classes exposed

mconnew:
Fixes a problem where adding a Connection header using HttpResponseMessageProperty.Headers wouldn't remove the existing Connection header based on HttpTransportBindingElement.KeepAliveEnabled. This resulted in a header which could look like this: `Connection: keep-alive, close".
Also tidied up the header adding code to using StringValues.Concat instead of concatenating using `+ ',' + newvalue`.
Exposed some classes needed as the basis for setting OperationContext.ServiceSecurityContext when using X509 certificates.
This allows manually providing a ServiceAuthorizationManager which does this. Automatic support for this will be coming in the future.


mconnew:
These changes were manually verified so committing.

PR #161, opened 2020-07-03
Add 2 BufferManagerTests testcases

willaimyou:
case1: VerifyBufferUsed( )
           <summary> Verify buffer object used by different threads before and after a thread gets trained
case2: VerifyBufferPools( )
            <summary> Verify BufferPools created for different buffer sizes

@ZhaodongTian


willaimyou:
@HongGit, @mconnew

PR #165, opened 2020-07-06
Streamed message body test

imcarolwang:
Service operation with ref/out parameter run failed, seems it's not supported yet, the related scenarios are currently commented out.

@HONGGIT, @mconnew.


mconnew:
@imcarolwang, looks like these tests are conflicting with a previous test that had been merged. Can you resolve the conflict. Also can you open an issue for the failing tests. ref/out should be working so this is a bug.

imcarolwang:
> @imcarolwang, looks like these tests are conflicting with a previous test that had been merged. Can you resolve the conflict. Also can you open an issue for the failing tests. ref/out should be working so this is a bug.

Issue opened for the ref/out problem: #196

PR #167, opened 2020-07-07
Fixed bug in Message and TextMessageEncoder

mconnew:


src/CoreWCF.Primitives/src/CoreWCF/Channels/Message.cs
```diff
@@ -1056,6 +1055,7 @@ protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
 
         public override async Task OnWriteMessageAsync(XmlDictionaryWriter writer)
         {
+            WriteMessagePreamble(writer);
```
BradBarnich, line 1058: This is the bug, yes? Or was there something wrong with using the async methods on XmlWriter?

mconnew: Yes. While most concrete implementations of XmlWriter support using the Async methods, XmlDictionaryWriter implementations don't.

PR #170, opened 2020-07-09
PerCall InstanceContextMode lifetime test

imcarolwang:
@honggit, @mconnew .

### DIFF
--- a/src/CoreWCF.Http/tests/ServiceContract/ILifeTimeTestService.cs
+++ b/src/CoreWCF.Http/tests/ServiceContract/ILifeTimeTestService.cs
@@ -1,0 +1,20 @@
+﻿using CoreWCF;
+
+namespace ServiceContract
+{
+    [ServiceContract]
+    public interface ILifeTimeTestService
+    {
+        [OperationContract]
+        void Start();
+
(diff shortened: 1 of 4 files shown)


src/CoreWCF.Http/tests/ServiceContract/ILifeTimeTestService.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace ServiceContract
+{
+    [ServiceContract]
+    public interface ILifeTimeTestService
+    {
+        [OperationContract]
+        void Start();
+
+        [OperationContract(IsOneWay = true)]
+        void OneWay();
```
imcarolwang, line 12: @mconnew , IsOneWay operation could build but client call would fail, it seems it is not fully supported, right?

mconnew: It should be working. If this isn't working, it's a bug. Please open an issue and add a comment pointing to the issue.

imcarolwang: Issue #197.

PR #171, opened 2020-07-10
Add complex serialization test cases

ZhaodongTian:
@mconnew @HongGit , Is there issues with Out and Ref when using in service ? the value cannot be returned .I comment out these verification temporary

### DIFF
--- a/src/CoreWCF.Http/tests/ClientContract/IDataContractEventsService.cs
+++ b/src/CoreWCF.Http/tests/ClientContract/IDataContractEventsService.cs
@@ -1,0 +1,54 @@
+using ServiceContract;
+using System.ServiceModel;
+
+namespace ClientContract
+{
+    [ServiceContract]
+    public interface IDataContractEventsService
+    {
+        [OperationContract()]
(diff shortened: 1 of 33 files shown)


mconnew:
@ZhaodongTian, this is a known issue. Can you update issue #196 with the test variations which are failing, disable those variations and add a comment with a link in the test pointing to issue #196.

mconnew:
@ZhaodongTian, I  think if you close the PR and re-open it, you will trigger the PR being re-run. I'm still trying to work out how to give external collaborators permission on how to use /azp run. Until I get that worked out, hopefully this will be sufficient.

mconnew:
@ZhaodongTian, please update issue #196 and add comments next to the disabled tests referencing the issue. I didn't check the disabled tests, but just in case you didn't, can you disable the test by commenting out the test attributes but leave the test to be compiled.

ZhaodongTian:
> @ZhaodongTian, please update issue #196 and add comments next to the disabled tests referencing the issue. I didn't check the disabled tests, but just in case you didn't, can you disable the test by commenting out the test attributes but leave the test to be compiled.

@mconnew  I have commented out the verification for Out and Ref in client. All test cases in this branch have passed.

PR #177, opened 2020-07-14
Create different service signatures and call from async client.

v-fangfc:
@mconnew @HongGit


mconnew:
It looks like some changes in this PR are adding types with the same names from a previously merged PR. These are the build errors:
```
src\CoreWCF.Http\tests\RequestReplyTests.cs(76,20): Error CS0101: The namespace 'CoreWCF.Http.Tests' already contains a definition for 'Startup'
src\CoreWCF.Http\tests\RequestReplyTests.cs(79,21): Error CS0111: Type 'Startup' already defines a member called 'ConfigureServices' with the same parameter types
src\CoreWCF.Http\tests\RequestReplyTests.cs(83,21): Error CS0111: Type 'Startup' already defines a member called 'Configure' with the same parameter types
```

PR #185, opened 2020-08-03
Add TypedMessageTypedMethod test cases

ZhaodongTian:
@mconnew @HongGit

### DIFF
--- a/src/CoreWCF.Http/tests/ServiceContract/ITypedMessageTypedMethodMyService.cs
+++ b/src/CoreWCF.Http/tests/ServiceContract/ITypedMessageTypedMethodMyService.cs
@@ -1,0 +1,235 @@
+﻿using CoreWCF;
+using System;
+using System.Collections;
+using System.Runtime.Serialization;
+using System.Xml;
+using System.Xml.Serialization;
+
+namespace ServiceContract
+{
+    public static class ResultHelper
+    {
+        public static string fromMethod;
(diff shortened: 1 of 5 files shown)


src/CoreWCF.Http/tests/ServiceContract/ITypedMessageTypedMethodMyService.cs
```diff
@@ -0,0 +1,235 @@
...
+        void MyOperation(FooMessage6 request);
+
+        [OperationContract]
+        DateTime Method6(int num, Foo foo);
+    }
+
+    [MessageContract]
+    public class FooMessage6
+    {
+        // [MessageProperty(Name = "Property1")] MessageProperty don't support
```
mconnew, line 171: I'm not seeing any test code which would have exercised the MessageProperty attribute. The type of code I would expect to see to test this attribute would be some extensibility point being used, probably an IDispatchMessageInspector which checks that the message property has been added. Are these types reused in other tests which aren't included in this PR? Basically my concern is the MessageProperty attributes seem to be purposefully added but I'm not seeing any consequences of them being commented out. If they were never used in the original tests, just remove the attributes instead of commenting them out.

ZhaodongTian: I checked the original tests ,which never used, I will remove it

PR #188, opened 2020-08-05
Test the RemoteEndpointMessageProperty.

imcarolwang:
The RemoteEndpointMessageProperty allows one to retrieve the IP and port of the sender of a message. 

@HongGIT, @mconnew .


mconnew:
I am absolutely amazed that this works on linux!! But apparently the build hosts have netstat on the path and the output and the command line parameters of netstat on linux are sufficiently compatible to make this work.

PR #189, opened 2020-08-06
Fix HTTP binding to use all async IO in BufferedMode scenario

mconnew:
Embarrassingly I had left the HTTP transport to read from the incoming request stream synchronously. This fixes that faux pas. It still will sometimes use sync reads and writes in the streamed mode scenario as Xml[Dictionary]Reader/Xml[Dictionary]Writer aren't entirely async and neither are XmlSerializer or DataContractSerializer.

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/AspNetCoreReplyChannel.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/AspNetCoreReplyChannel.cs
@@ -97,8 +97,7 @@ namespace CoreWCF.Channels
 
             var requestContext = HttpRequestContext.CreateContext(_httpSettings, context);
             var httpInput = requestContext.GetHttpInput(true);
-            Exception requestException;
-            Message requestMessage = httpInput.ParseIncomingMessage(out requestException);
+            (Message requestMessage, Exception requestException) = await httpInput.ParseIncomingMessageAsync();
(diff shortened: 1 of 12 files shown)


mconnew:
Skipping CR as no logic change, just sync -> async

src/CoreWCF.Http/src/CoreWCF/Channels/AspNetCoreReplyChannel.cs
```diff
@@ -97,8 +97,7 @@ internal async Task HandleRequest(HttpContext context)
 
             var requestContext = HttpRequestContext.CreateContext(_httpSettings, context);
             var httpInput = requestContext.GetHttpInput(true);
-            Exception requestException;
-            Message requestMessage = httpInput.ParseIncomingMessage(out requestException);
+            (Message requestMessage, Exception requestException) = await httpInput.ParseIncomingMessageAsync();
             requestMessage.Properties.Add("Microsoft.AspNetCore.Http.HttpContext", context);
```
Daniel-Svensson, line 101: I think this can throw a NRE if the method returns an exception and null request

mconnew: You are right. Although that wasn't introduced by this change. I'll fix this asap.

PR #211, opened 2020-08-25
Empty HTTP requests to active endpoints should respond per WCF

danielcrenna:
This PR adds `HttpClient` based integration testing capability, and uses it to verify a patch to make CoreWCF respond the same way as WCF when an HTTP request to an active service endpoint is empty. It currently attempts to parse a SOAP message and crashes, rather than send a terminal `400 Bad Request`, as per WCF.


mconnew:
It looks like this has some fixes from other PR's (TestServer stuff). Can you fix the conflicts and revert the channel close logic I'll merge this one.

PR #212, opened 2020-08-25
nit: IServiceBuilder chaining

danielcrenna:
nit: Change `IServiceBuilder` interface to match other official .NET Core APIs that use the builder to "chain" method calls configuring the same scope of services.

### DIFF
--- a/src/CoreWCF.Primitives/tests/ServiceBuilderTests.cs
+++ b/src/CoreWCF.Primitives/tests/ServiceBuilderTests.cs
@@ -12,6 +12,7 @@ namespace CoreWCF.Primitives.Tests
         public void ServiceBuilderCanChain()
         {
             var services = new ServiceCollection();
+            services.AddServiceModelServices();
             var serviceProvider = services.BuildServiceProvider();
             var builder = new ServiceBuilder(serviceProvider);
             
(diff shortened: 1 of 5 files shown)


src/CoreWCF.Primitives/tests/ServiceBuilderTests.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class ServiceBuilderTests
+    {
+        [Fact]
+        public void ServiceBuilderCanChain()
+        {
+            var services = new ServiceCollection();
+            services.AddServiceModelServices();
+            var serviceProvider = services.BuildServiceProvider();
+            var builder = new ServiceBuilder(serviceProvider);
```
mconnew, line 17: I'm not a fan of having product code referencing a test assembly with an InternalsVisibleTo attribute. I believe that if one assembly is strong name signed, then the reference to the other one must be too. Once I get all the signing stuff working, this will break this test unless the tests are also signed.  
It looks like you only need this to be able to reference and construct an instance of ServiceBuilder. The AddServiceModelServices adds this to the service collection as well as a factory which returns the same singleton instance when requesting an IServiceBuilder. So you should be able to replace this line of code with the following and remove the InternalsVisibleTo attribute.
```suggestion
            var builder = serviceProvider.GetRequiredService<IServiceBuilder>();
```

mconnew:
Once the InternalsVisibleTo stuff is fixed I'll merge this.

PR #213, opened 2020-08-25
Fix builder errors caused by `Microsoft.Test.Sdk` version `16.4.0` in VS 2019

danielcrenna:
- In this SDK, MSBuild parameter `GenerateProgramFile` is not test to `true` by default, but recent builds of VS2019 expect it
- End result is `Program does not contain a static 'Main' method suitable for an entry point`
- updating the SDK results in no other changes required to fix the build in recent updates of VS2019

### DIFF
--- a/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
+++ b/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
@@ -17,7 +17,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
     <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
     <PackageReference Include="System.ServiceModel.Http" Version="4.7.0" />
     <PackageReference Include="System.ServiceModel.Primitives" Version="4.7.0" />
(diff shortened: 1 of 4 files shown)


danielcrenna:
@mconnew Just checking in on this PR -- without it I can't build the test projects in VS2019 without re-applying a patch, which makes it harder to manage all the open threads I have against this project

danielcrenna:
@mconnew I see I have a few things in this PR that you didn't want though they cause intermittent crashes -- that said the only commit that is relevant is https://github.com/CoreWCF/CoreWCF/pull/213/commits/693c5bee901a55489735ec478bde299e598d6afa which is just to update `16.4.0` with it's build bug in latest VS2019, to the latest version.

mconnew:
@danielcrenna, can you pull the change to the test cleanup code and then I'll merge this PR.

danielcrenna:
@mconnew done - hopefully the CI/CD will luck out and pass

mconnew:
The failure is due to some test hanging. Here's the test log output:
```
"C:\Program Files\dotnet\dotnet.exe" test D:\a\1\s\src\CoreWCF.Primitives\tests\CoreWCF.Primitives.Tests.csproj --logger trx --results-directory D:\a\_temp --no-restore --no-build --configuration Release --framework netcoreapp2.1
Test run for D:\a\1\s\bin\Release\CoreWCF.Primitives.Tests\netcoreapp2.1\CoreWCF.Primitives.Tests.dll(.NETCoreApp,Version=v2.1)
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
##[error]The task has timed out.
```

I will kick it off again, it normally runs the second time. There's an option I can add to the test run to make it produce a blame log to find the test which caused problems, but I need to switch to the vstest runner. I seem to remember there's some way to get a process dump when a test times out. I think that would be useful anyway as the devops test machines appear to run a lot slower than my dev machine and I think there's some timing issues (sometimes test issues) which only show up on slow hosts.

danielcrenna:
@mconnew I've seen the timeout one but the logs I linked earlier show that this PR failed prior due to "machine actively refused" a connection, suggesting a binding race condition. Blame log is probably a good idea if the goal is for the tests not to exhibit magic behaviour -- for me it's harder because I don't have the ability to "bounce" the build if it fails due to a timeout or race condition error.

danielcrenna:
@mconnew at any rate, this mini-PR seems to be passing all checks now.

PR #219, opened 2020-08-27
Fix #218: Add port auto-assignment feature for `net.tcp` and use it for unit tests

### LINKED ISSUE #218: Intermittent `AddressInUseException` for nettcp false test failure on during CI/CD process
Builds often fail intermittently due to an `AddressInUseException` preventing Kestrel from starting.

Example log: https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=37970&view=logs&j=df77fcc7-169b-539a-4b1a-0702f867ecad&t=5eb216d8-81f4-5458-f745-21d97554dbca

danielcrenna:
This PR adds a feature to pass `0` to `UseNetTcp` which will properly handle a deferred/auto-assigned port, rather always require an explicit port. The port is assigned "implicitly" via the TCP stack, there is no "hunt and peck" required.

The feature is then used to convert `net.tcp` tests to use auto-assigned ports, to fix issue #218, where the build server will intermittently fail on `AddressInUseException`.

### DIFF
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
@@ -8,6 +8,9 @@ using CoreWCF.Channels.Framing;
 using System;
 using System.Collections.Generic;
 using System.Net;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.Logging;
 
 namespace CoreWCF.Configuration
 {
@@ -41,6 +44,17 @@ namespace CoreWCF.Configuration
                 o.EndPoints.Add(endPoint);
             });
 
+            //
+            // In cases where we're only binding to net.tcp, we still need this feature,
+            // in order to support automatically assigning TCP ports at runtime
+            //
+            services.TryAddSingleton<IFeatureCollection>(r =>
+            {
+                var features = new FeatureCollection();
+                features.Set<IServerAddressesFeature>(new ServerAddressesFeature());
+                return features;
+            });
+
             return services;
         }
     }
@@ -53,10 +67,12 @@ namespace CoreWCF.Configuration
 
     internal class NetTcpFramingOptionsSetup : IConfigureOptions<KestrelServerOptions>
     {
+        private readonly ILogger<NetTcpFramingOptions> _logger;
         private readonly NetTcpFramingOptions _options;
 
-        public NetTcpFramingOptionsSetup(IOptions<NetTcpFramingOptions> options)
+        public NetTcpFramingOptionsSetup(IOptions<NetTcpFramingOptions> options, ILogger<NetTcpFramingOptions> logger)
         {
+            _logger = logger;
             _options = options.Value;
         }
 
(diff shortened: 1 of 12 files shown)


src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
```diff
@@ -8,6 +8,9 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
+using Microsoft.AspNetCore.Hosting.Server.Features;
```
mconnew, line 11: Can you sort usings

src/CoreWCF.NetTcp/tests/TaskBasedOperationTests.cs
```diff
@@ -86,11 +88,12 @@ public async Task StreamedSynchronouslyCompletingTask()
                 System.ServiceModel.ChannelFactory<Contract.ITaskService> factory = null;
                 Contract.ITaskService channel = null;
                 host.Start();
+
                 try
```
mconnew, line 92: Nit: extra blank line introduced

mconnew:
This is fantastic, thank you. I had a look at the failures and it's a simple enough fix. Here's the error message:

> System.InvalidOperationException : Unable to resolve service for type 'Microsoft.Extensions.Logging.ILogger`1[CoreWCF.ServiceHostObjectModel`1[DependencyInjection.PerCallInstanceContextSimpleServiceAndBehavior]]' while attempting to activate 'CoreWCF.ServiceHostObjectModel`1[DependencyInjection.PerCallInstanceContextSimpleServiceAndBehavior]'.

This is caused by the test framework I've created for CoreWCF.Primitives. It's basically a System.ServiceModel binding to use with the client which uses the dispatcher to dispatch the calls. It means you don't need ASP.NET Core to host the dispatch pipeline to test non-transport related things. So I have to create an IServiceProvider with all the dependencies that the code uses. You've added usage of an ILogger\<T\> which doesn't exist in the IServiceProvider in those tests. [Here's where the code is](https://github.com/CoreWCF/CoreWCF/blob/master/src/CoreWCF.Primitives/tests/DispatcherClient/DispatcherChannelFactory.cs#L35) which populates the service provider. You just need to the ILogger stuff to DI there and this PR should pass.

src/CoreWCF.NetTcp/tests/TaskBasedOperationTests.cs
danielcrenna: Fixed. I think there is another PR for an `.EDITORCONFIG` file that would allow us to lint to your preferences, or you could create your own for that purpose.

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
danielcrenna: Fixed

mconnew:
It's still failing for the same reason. I'm going to pull your change down and debug it locally.

mconnew:
Just FYI, the build breaks were from some tests which tested creating a dispatcher and they now needed an ILogger, and a test which verified the server added the remote endpoint property correctly and had hard coded port 8080 into the test. I modified the test to use the dynamic port.

danielcrenna:
@mconnew great, thank you!

PR #221, opened 2020-09-08
Fix CompressionFormat for BinaryMessageEncoder

### LINKED ISSUE #220: Meet error when using Gzip compression format
I’m working on migrating our service from .Net Framework to .Net Core, using CoreWcf to support wcf service in .Net Core.
I set up an endpoint for gzip as shown below: 

### Client (With .Net Framework)
```
var binding = new CustomBinding();
binding.Elements.Add(new BinaryMessageEncodingBindingElement { CompressionFormat = CompressionFormat.Gzip });
binding.Elements.Add(new HttpsTransportBindingElement { MaxReceivedMessageSize = 200065536 });

var channelFactory = new ChannelFactory<IChannel>(binding, new EndpointAddress(endpoint));
var gzipChannel = channelFactory.CreateChannel();
var response = gzipChannel.MyMehod();
```

### Server (With .Net Core)
```
serverBuilder.AddServiceEndpoint<MyServer, IChannel>(
                       CoreWCFBindings.CreateMyBinaryCompressionBinding(),
                       "/MyEndpoint");

public static CustomBinding CreateMyBinaryCompressionBinding()
        {
            var binding = new CustomBinding();
            var encodingElement = new BinaryMessageEncodingBindingElement
            {
                CompressionFormat = CompressionFormat.Gzip
            };

            var httpElement = new HttpTransportBindingElement
            {
                MaxReceivedMessageSize = 200065536
            };

            binding.Elements.Add(encodingElement);
            binding.Elements.Add(httpElement);

            return binding;
        }
```

Everything works fine if we set CompressionFormat=None. But, if change it to Gzip, client side will receive exception below(server side will not throw any exception):
WebException: The underlying connection was closed: The connection was closed unexpectedly.

Looks like there is a problem with Gzip compression format in CoreWcf. Cloud you please take a look and point us why it go wrong? If you don’t have enough time to fix it, we can fix on a private branch to unblock our migration. Thanks in advance!

~ Ruoxi

mconnew:
A few fixes here. The most important is to enable copying the MaxMessageSize property from the transport to the BinaryMessageEncoder. This required stopping the binding from being cloned on a critical code path as the copying of the value would happen on the clone and the eventual encoder is created on one of the earlier copies.
There was also a bug where the HttpOutput object would be created twice resulting in a duplicate message property when using compression.
Fixes #220

### DIFF
--- a/src/Common/src/CoreWCF/Runtime/IOThreadScheduler.cs
+++ b/src/Common/src/CoreWCF/Runtime/IOThreadScheduler.cs
@@ -641,7 +641,11 @@ namespace CoreWCF.Runtime
                 {
                     throw Fx.AssertAndThrowFatal("Cleanup called on an overlapped that is in-flight.");
                 }
-                Overlapped.Free(_nativeOverlapped);
+
+                if (s_isWindows)
+                {
+                    Overlapped.Free(_nativeOverlapped);
+                }
             }
         }
     }
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
@@ -228,7 +228,6 @@ namespace CoreWCF.Channels
             try
             {
                 bool closeOutputAfterReply = PrepareReply(ref responseMessage);
-                httpOutput = GetHttpOutput(message);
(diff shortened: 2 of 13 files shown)


StephenMolloy (commented):
lgtm

PR #223, opened 2020-09-14
CSProj changes for making NUGET ready

birojnayak:
starting the PR for making dotnet pack ready...


src/CoreWCF.Http/src/CoreWCF.Http.csproj
```diff
@@ -4,6 +4,17 @@
     <AssemblyName>CoreWCF.Http</AssemblyName>
     <PackageId>CoreWCF.Http</PackageId>
     <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <VersionPrefix>0.0.0</VersionPrefix>
+    <VersionSuffix>pre-beta</VersionSuffix>
+    <Owners>dotnetfoundation</Owners>
+    <Authors>mconnew, biroj, .NET Foundation and Contributors</Authors>
```
mconnew, line 10: This is going to be difficult to maintain. I don't like the idea that package metadata like this needing to be updated with each release. I did a very quick look at what other packages in the dotnet foundation do and the one which made most sense to me would be "CoreWCF Contributors". We'll make sure there's a contributors section in the release notes and we might even be able to create a stable URL for it to include here as you can create GitHub url's to content which is dynamic in that you can point to the latest version of a document of a specific branch. For now just use "CoreWCF contributors" and we'll try to make something which is more explicit before RTM.

src/CoreWCF.Http/src/CoreWCF.Http.csproj
```diff
@@ -4,6 +4,17 @@
...
+    <Copyright>Copyright (c) .NET Foundation and Contributors.</Copyright>
+    <Description>Core WCF is a port of Windows Communication Framework (WCF) to .NET Core. The goal of this project is to enable existing WCF projects to move to .NET Core</Description>
+    <NeutralLanguage>en-US</NeutralLanguage>
+    <PackageTags>WCF .net core, CoreWCF</PackageTags>
+    <PackageProjectUrl> https://github.com/CoreWCF/CoreWCF </PackageProjectUrl>
```
mconnew, line 15: NIT, trailing space on the URL

src/CoreWCF.Http/src/CoreWCF.Http.csproj
mconnew: I was wrong, according to the documentation, the Authors field is about the nuget account which owns the package so that people can search for other packages from the same owner. From the [,nuget spec](https://docs.microsoft.com/en-us/nuget/reference/nuspec#authors):

> A comma-separated list of packages authors, matching the profile names on nuget.org. These are displayed in the NuGet Gallery on nuget.org and are used to cross-reference packages by the same authors.

So that field is yet to be worked out.

mconnew:
For the items which are common in all the packages, can you use Directory.Build.props to define them. You can either create new named properties in the props file and reference them here, or just define the properties in the props file directly. Another option would be to create a new file called nuspec.props which is imported at the top of each csproj which defines them.  

I just did a quick browse through the .nuspec documentation and it would be good to be able to populate the repository metadata. It should include the branch and commit that the package was built from. Documentation for how to specify those values is [at this page](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target). Can you look into whether GitVersion populates any properties with the values which can be used for `RepositoryBranch` and `RepositoryCommit`. `RepositoryUrl` and `RepositoryType` can already be populated (that one needs the .git suffix on the URL as it should be the path used with git).  

Can you also set `PackageRequireLicenseAcceptance` to true as it defaults to false.  

What happened to CoreWCF.Primitives?

src/CoreWCF.Http/src/CoreWCF.Http.csproj
```diff
@@ -4,6 +4,17 @@
     <AssemblyName>CoreWCF.Http</AssemblyName>
     <PackageId>CoreWCF.Http</PackageId>
     <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <VersionPrefix>0.0.0</VersionPrefix>
+    <VersionSuffix>pre-beta</VersionSuffix>
```
mconnew, line 8: We should use one of the recommended suffixes [listed here](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions), As the api is still changing slightly, as you identified beta isn't appropriate. -alpha seems the best match.

mconnew:
For the repository branch and version, if GitVersion doesn't supply values we can use, dotnet pack allows for passing these on the command line so those two values could be handled in the pipeline yaml by passing them.

src/CoreWCF.Http/src/CoreWCF.Http.csproj
birojnayak: How about for the time being let's keep ".NET Foundation and Contributors"

mconnew: As other packages use `%ProjectName% Contributors`, I'm fine with "CoreWCF Contributors" for now as the .NET Foundation didn't author anything.

birojnayak:
> For the items which are common in all the packages, can you use Directory.Build.props to define them. You can either create new named properties in the props file and reference them here, or just define the properties in the props file directly. Another option would be to create a new file called nuspec.props which is imported at the top of each csproj which defines them.
> 
> I just did a quick browse through the .nuspec documentation and it would be good to be able to populate the repository metadata. It should include the branch and commit that the package was built from. Documentation for how to specify those values is [at this page](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target). Can you look into whether GitVersion populates any properties with the values which can be used for `RepositoryBranch` and `RepositoryCommit`. `RepositoryUrl` and `RepositoryType` can already be populated (that one needs the .git suffix on the URL as it should be the path used with git).
> 
> Can you also set `PackageRequireLicenseAcceptance` to true as it defaults to false.
> 
> What happened to CoreWCF.Primitives?

Moved to separate prop file(nuspec.props)... it seems dotnet pack getting all the repository info by itself(pasted below)

<licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
    <projectUrl>https://github.com/CoreWCF/CoreWCF</projectUrl>
    <description>Core WCF is a port of Windows Communication Framework (WCF) to .NET Core. The goal of this project is to enable existing WCF projects to move to .NET Core</description>
    <copyright>CoreWCF contributors</copyright>
    <tags>WCF .net core, CoreWCF</tags>
    <repository url="https://github.com/CoreWCF/CoreWCF" branch="nuget-pre-beta-package" commit="71117bd51105dfe567efb788da2168eee1dba24a" />
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="CoreWCF.Primitives" version="0.1.0-nuget-pre-beta-package.117" exclude="Build,Analyzers" />
        <dependency id="Microsoft.AspNetCore.WebSockets" version="2.1.7" exclude="Build,Analyzers" />
        <dependency id="Microsoft.Extensions.DependencyInjection.Abstractions" version="2.1.1" exclude="Build,Analyzers" />
        <dependency id="Microsoft.Extensions.Logging.Abstractions" version="2.1.1" exclude="Build,Analyzers" />
      </group>
    </dependencies>

birojnayak:
updated review comments

nuspec.props
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <PackageVersion>0.0.0-alpha</PackageVersion>
+    <Owners>dotnetfoundation</Owners>
+    <Authors>CoreWCF contributors</Authors>
+    <Copyright>CoreWCF contributors</Copyright>
+    <Description>Core WCF is a port of Windows Communication Framework (WCF) to .NET Core. The goal of this project is to enable existing WCF projects to move to .NET Core</Description>
```
mconnew, line 8: This needs to be .NET Foundation

PR #224, opened 2020-09-22
Fix race condition where connection reuse would happen before ReadAsync completed

mconnew:
On channel close, a pending read by NegotiateStream/SslStream is being delayed until the send part of the channel close has completed. What wasn't happening and needed to happen was to wait for that pending read to complete before preparing the socket for connection reuse. There was an incredibly difficult to hit race condition which would cause a connection failure due to two concurrent reads.

### DIFF
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
@@ -1,7 +1,6 @@
 ﻿using CoreWCF.Runtime;
 using System;
 using System.Buffers;
-using System.Diagnostics;
 using System.IO;
 using System.IO.Pipelines;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
```diff
@@ -11,16 +11,19 @@ namespace CoreWCF.Channels.Framing
 {
     public class RawStream : Stream
     {
+        private readonly FramingConnection _connection;
```
Treit, line 14: This field isn't used after the class is constructed, curious why it was added?

Treit (approved):

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/RawStream.cs
mconnew: To help in debugging. I can see if the FramingConnection has been reset. I'll stick an `#if DEBUG` around it as it's not needed for normal runs.

PR #228, opened 2020-09-29
Fix dependency on ServerAddressFeatures

### LINKED ISSUE #226: Server needs http endpoint, although client uses only net.tcp
I’m using net.tcp client and my server logic is reduced to use only net.tcp:

//Program.cs
            .UseNetTcp(8808)
            .UseStartup<Startup>();

//Startup.cs
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            {
                builder
                    .AddService<EchoService>()
                    .AddServiceEndpoint<EchoService, Contract.IEchoService>(new NetTcpBinding(), "/nettcp");
            });
        }

It was working fine inclusive version 0.1.0-alpha.115.
Since the next available version 0.1.0-alpha.121 I’m getting exception:

System.ArgumentException
  HResult=0x80070057
  Message=This collection already contains an address with scheme net.tcp.  There can be at most one address per scheme in this collection. If your service is being hosted in IIS you can fix the problem by setting 'system.serviceModel/serviceHostingEnvironment/multipleSiteBindingsEnabled' to true or specifying 'system.serviceModel/serviceHostingEnvironment/baseAddressPrefixFilters'. (Parameter 'item')
  Source=CoreWCF.Primitives
  StackTrace:
   at CoreWCF.UriSchemeKeyedCollection.InsertItem(Int32 index, Uri item) in C:\Users\me\Source\Repos\CoreWCF\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\UriSchemeKeyedCollection.cs:line 35
   at CoreWCF.Collections.Generic.SynchronizedCollection`1.Add(T item) in C:\Users\me\Source\Repos\CoreWCF\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\Collections\Generic\SynchronizedCollection.cs:line 96
   at CoreWCF.UriSchemeKeyedCollection..ctor(Uri[] addresses) in C:\Users\me\Source\Repos\CoreWCF\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\UriSchemeKeyedCollection.cs:line 20
   at CoreWCF.ServiceHostObjectModel`1..ctor(IServiceProvider serviceProvider, IServer server, IServiceBuilder serviceBuilder, ILogger`1 logger) in C:\Users\me\Source\Repos\CoreWCF\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\ServiceHostObjectModel.cs:line 36

After a little bit experimenting, I found, that adding http endpoint solves my exception.
The issue can be easy reproduced with samples: NetCoreClient & NetCoreServer.

mconnew:
Fixes #226 
This change removes the need for UseUrl and fixes an early change which ended up requiring an HTTP endpoint. It replaces using ServerAddressFeature to obtain the BaseAddresses url's with a mechanism so that the transports can register a BaseAddress more directly.

### DIFF
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
@@ -11,6 +11,8 @@ using CoreWCF.Channels.Framing;
 using System;
 using System.Collections.Generic;
 using System.Net;
+using System.Linq;
+using CoreWCF.Runtime;
 
 namespace CoreWCF.Configuration
 {
@@ -38,21 +40,10 @@ namespace CoreWCF.Configuration
         private static IServiceCollection AddNetTcpServices(this IServiceCollection services, IPEndPoint endPoint)
         {
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<KestrelServerOptions>, NetTcpFramingOptionsSetup>());
-
+            services.TryAddSingleton<NetMessageFramingConnectionHandler>();
             services.Configure<NetTcpFramingOptions>(o =>
             {
                 o.EndPoints.Add(endPoint);
-            });
-
-            //
-            // In cases where we're only binding to net.tcp, we still need this feature,
-            // in order to support automatically assigning TCP ports at runtime
-            //
-            services.TryAddSingleton<IFeatureCollection>(r =>
-            {
-                var features = new FeatureCollection();
-                features.Set<IServerAddressesFeature>(new ServerAddressesFeature());
-                return features;
             });
 
             return services;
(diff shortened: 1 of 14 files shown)


src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceModelServiceCollectionExtensions.cs
```diff
@@ -17,6 +19,25 @@ public static IServiceCollection AddServiceModelServices(this IServiceCollection
...
+                    }
+                    services[i] = ServiceDescriptor.Singleton<IServer>((provider) =>
+                        {
+                            var originalIServer = (IServer)provider.GetRequiredService(implType);
+                            var wrappingServer = provider.GetRequiredService<WrappingIServer>();
+                            wrappingServer.InnerServer = originalIServer;
+                            return wrappingServer;
+                        });
+                }
+            }
```
halter73, line 40: This is probably necessary to run something that causes the legacy WebHost.StartAsync() to throw after the server binds to an endpoint. But if you can create an extension method on IHostBuilder (which configures the generic host which exists in 2.1 and is the only host getting new features) that runs after ConfigureWebHost(), you could add an IHostedService that starts after the server and causes Host.StartAsync() to throw.

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
```diff
@@ -69,38 +60,41 @@ internal class NetTcpFramingOptionsSetup : IConfigureOptions<KestrelServerOption
...
-                    // Explicit port: can safely add directly to the serviceBuilder (default behavior)
-                    _logger.LogDebug($"Adding base address {baseAddress} to serviceBuilder");
-                    serviceBuilder.BaseAddresses.Add(baseAddress);
-                }
+        private void OnServiceBuilderOpening(object sender, EventArgs e)
+        {
+            foreach (var listenOptions in ListenOptions)
+            {
+                var endpoint = listenOptions.IPEndPoint;
+                var baseAddress = new Uri($"net.tcp://localhost:{endpoint.Port}/");
```
halter73, line 95: Using IServerAddressesFeature will get you every address Kestrel binds to, not just the ones added via NetTcpFramingOptionsSetup.Configure(). IServerAddressesFeature is also supported by non-Kestrel servers. I'm not sure if either of those things matter.

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceModelServiceCollectionExtensions.cs
mconnew: I opened an issue so this doesn't get missed. The second release will target .NET 5+ (probably want to settle on a LTS release, but aiming to work on 5.0+) and I think that's the time to do this as I plan to make some api changes anyway.

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
mconnew: For HTTP, we don't care about the which ports are used. The behavior of WCF is that when you configure using the HTTP transport, we really don't care about which port it arrived on or for what hostname. The endpoint selection code converts http and https to just http for the purposes of routing the request so if an Http binding is used, To maintain the same behavior in CoreWCF, I just add "http://localhost/" to the base addresses and don't care about what's in IServerAddressesFeature.  
This means I only care about the NetTcp listening port I've added which means I need to have a list of the ListenOptions instances which were only used for NetTcp which is why I stash. Then in the initialization of CoreWCF, I use the list of ListenOptions to add the relevant addresses to IServerAddressesFeature.  
IServerAddressesFeature is useless for my use case as it claims everything is an http endpoint when that's not the case.

PR #231, opened 2020-09-30
Fix PublishStage.yml

mconnew:


mconnew:
Skipping waiting for tests to run as not a code fix and doesn't change behavior of PR pipeline

Kralizek:
@mconnew
What do you think about leaving the `/templates` folder to some `dotnet new` template to be created later on?

mconnew:
Anything which is product code which would be shipped should live in the /src folder. Although I do understand your point about ambiguity. Next time I'm making some code changes I'll move them to /eng/templates as I'd like to get tests running with docker to be able to do enterprise windows authentication testing so all that infrastructure stuff needs somewhere to live other than the root folder.

PR #248, opened 2020-11-04
Change visibility of types public in official WCF

danielcrenna:
What should the approach be for types marked internal in CoreWCF but public in WCF?

This PR contains visibility changes that, if not applied, prevent my current upgrade project from continuing due to build failures, as the project utilizes custom filters.

This PR is particular to my port and most assuredly doesn't call all type visibility differences between the two projects.


mconnew:
@danielcrenna, the reason anything public in WCF is internal is because of lack of tests. I know the message filter stuff works as it's used extensively internally so I'm comfortable exposing those. I have no idea if the IChannelInitializer stuff is functioning. I honestly haven't even tried a duplex scenario so don't know if that's even working which is the scenario where ClientRuntime.ChannelInitializers is used. I'm okay with making them public as long as you understand they might not work right now and might need some effort to get it working as this is completely untested. I can't promise any kind of timeframe to get it working if it isn't. Having said that, my priority has been to get real life usage working ahead of the "someone might want to use this" issues so it would be higher up my priority list. If you're okay with that ambiguity, I'll merge this.

danielcrenna:
@mconnew I'm OK with the ambiguity. 

We're moving a very large, highly customized (in terms of custom behaviours, message inspectors, service host factories, etc.) WCF service to .NET Core, via CoreWCF.

I need to make them public for the sole reason of unblocking the porting, knowing I'll likely run into issues, but the intention is to do further work to fix things that come up as part of this project.

PR #256, opened 2020-11-17
Fix DnsCache.MachineName on Docker Linux container

mconnew:
Remove dead code from DnsCache
Add another stage to pipeline to sign symbol packages

### DIFF
--- a/src/Common/src/CoreWCF/Channels/DnsCache.cs
+++ b/src/Common/src/CoreWCF/Channels/DnsCache.cs
@@ -1,26 +1,20 @@
 ﻿using System;
+using System.Diagnostics;
+using System.IO;
 using System.Net;
 using System.Net.Sockets;
-using CoreWCF.Collections.Generic;
+using System.Runtime.InteropServices;
 
 namespace CoreWCF.Channels
 {
-    static class DnsCache
+    internal static class DnsCache
     {
-        const int mruWatermark = 64;
-        static MruCache<string, DnsCacheEntry> resolveCache = new MruCache<string, DnsCacheEntry>(mruWatermark);
-        static readonly TimeSpan cacheTimeout = TimeSpan.FromSeconds(2);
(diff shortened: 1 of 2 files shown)


templates/CodeSignStage.yml
```diff
@@ -42,6 +35,19 @@ stages:
               --descriptionUrl "https://github.com/CoreWCF/CoreWCF"
             displayName: Sign packages
 
+          - pwsh: |
```
AAATechGuy, line 38: don't have much expertise on this change.

src/Common/src/CoreWCF/Channels/DnsCache.cs
```diff
@@ -1,26 +1,20 @@
...
+    internal static class DnsCache
     {
-        const int mruWatermark = 64;
-        static MruCache<string, DnsCacheEntry> resolveCache = new MruCache<string, DnsCacheEntry>(mruWatermark);
-        static readonly TimeSpan cacheTimeout = TimeSpan.FromSeconds(2);
+        private const string LinuxProcHostnamePath = "/proc/sys/kernel/hostname";
 
         // Double-checked locking pattern requires volatile for read/write synchronization
-        static volatile string machineName;
+        private static volatile string machineName;
```
AAATechGuy, line 15: not sure, if this was needed in C#. 
https://stackoverflow.com/questions/1964731/the-need-for-volatile-modifier-in-double-checked-locking-in-net

AAATechGuy (approved):
signing off with comments.

src/Common/src/CoreWCF/Channels/DnsCache.cs
AAATechGuy: nvm, the comment was specific to x86/x64 processor architecture

mconnew: Yep, the JIT is smart enough to no-op it on x86 and x64 platforms and to do a volatile read on platforms which need it. I think ARM has a weaker memory model.

PR #269, opened 2020-12-11
Fix race condition when new request is waiting for the instance just as it's being released

mconnew:
Basically the queue isn't being created under the lock so when the instance context is being released under lock, it checks and sees the queue either doesn't exist yet or doesn't have an item in the queue. This fix updates the queue under lock and only await's the returned task outside the lock.


mconnew:
@birojnayak, could you review this? It's what's been causing the occasional pipeline builds to fail from time to time.

birojnayak (approved):
one line outside lock probably would have taken 2 days to figure out the issue... :)

PR #292, opened 2021-02-02
Code cleanup

mconnew:


birojnayak:
Ensure all checks are passing before merging and just run one round of stress test.

birojnayak (approved):
Ensure all checks are passing before merging and just run one round of stress test.

mconnew:
I've run many hours of the one failing test on repeat and not been able to get it to fail again. I believe there was CPU contention on the test machine which caused the close code path to take too long and the connection was aborted. WebSockets spec says when one end sends a close message, it is supposed to wait a reasonable amount of time for the other end to send a close message, and if it doesn't arrive in a reasonable amount of time, it may abort the connection. I believe this is what happened in this case as the spec doesn't say what's reasonable. I think things just took too long to complete on the close handshake. Merging as this doesn't look to be an issue, and if it is, it's only when closing a connection so impact is minimal.

PR #294, opened 2021-02-05
Don't send Connection header if not using HTTP/1.1

mconnew:


mconnew:
@Treit, this is the fixes for HTTP/2 support in CoreWCF. Can you take a look?

Treit (approved):

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -353,13 +354,16 @@ protected override HttpInput GetHttpInput()
...
+                if (StringComparer.OrdinalIgnoreCase.Equals(Http11ProtocolString, _aspNetContext.Request.Protocol))
                 {
-                    _aspNetContext.Response.Headers["Connection"] = "keep-alive";
-                }
-                else
-                {
-                    _aspNetContext.Response.Headers["Connection"] = "close";
+                    if (HttpTransportSettings.KeepAliveEnabled)
+                    {
+                        _aspNetContext.Response.Headers["Connection"] = "keep-alive";
```
halter73, line 361: @mconnew Nit: `keep-alive` is the default behavior if the `Connection` header is omitted for HTTP/1.1, so it shouldn't be necessary to set it here.

PR #301, opened 2021-02-10
Integration tests

### LINKED ISSUE #216: Request will crash if no remote IP address is found
Simple missing null check will cause `NullReferenceException` when there is no RemoteIPAddress. This was discovered when setting up the ASP.NET Core MVC `TestServer`, which does not have a RemoteIPAddress as it is in memory and does not bind to an address. Presumably, other technology will induce this, such as masking proxies.

The fix is in https://github.com/CoreWCF/CoreWCF/pull/215.

mconnew:
This replaces PR #215. I rebased that PR and made a few fixes where the code in CoreWCF had changed since the PR and was broken for the integration test scenario. Plus a little bit of tidying up.
Fixes #216 
@danielcrenna, can you take a look at make sure this has everything in it that you wanted?


mconnew:
@danielcrenna, if you see any problems, feel free to reopen the original issue and comment there. I'm going to merge these changes. Thanks again for contributing this and doing the hard work to get it working across multiple netcoreapp versions. I didn't know about the test server feature.

PR #326, opened 2021-03-12
Fix HttpSys and IIS hosting, added HttpSys tests

mconnew:
IIS Host adds non-http bindings to IServerAddresses, this skip's those.
HttpSys has a base path which broke address mapping, this fixes that.
HttpSys can use + or * wildcards in server address, this fixes that.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceBuilder.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceBuilder.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using CoreWCF.Channels;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace CoreWCF.Configuration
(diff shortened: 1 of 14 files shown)


mconnew:
@birojnayak, can you take a look at this?

birojnayak (approved):
Make sure tests are passing...

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceBuilder.cs
```diff
@@ -15,9 +16,16 @@ internal class ServiceBuilder : CommunicationObject, IServiceBuilder
...
 
-        public ServiceBuilder(IServiceProvider serviceProvider)
+        public ServiceBuilder(IServiceProvider serviceProvider, IApplicationLifetime appLifetime)
         {
             ServiceProvider = serviceProvider;
+            appLifetime.ApplicationStarted.Register(() =>
+            {
+                if (State == CommunicationState.Created)
+                {
+                    OpenAsync().GetAwaiter().GetResult();
```
davidfowl, line 26: Do you need to call OpenAsync here or can it be started and awaited elsewhere?

mconnew: Thinking about it, it doesn't actually need to be waited here at all. This is for the case when IIS hosted and there isn't a wrapping IServer to ensure OpenAsync is called. So this is a fallback so that if nothing previously called OpenAsync, we make sure it gets called but we won't see the exception. I'll take out the waiting and use a discard.

mconnew:
Test failures are because I forgot to mark HttpSys test as Windows only.

Issue #335, opened 2021-03-26
Enabling Asp.net core Authorize attribute (for rolebased permission) instead of principalpermissionattribute


birojnayak:
Enabling Asp.net core Authorize attribute (for rolebased permission) at API level, rather using principalpermissionattribute https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/principalpermissionattribute-obsolete

Issue #265, opened 2020-12-02
For Windows Auth (WsHTTP), get the users permission group in Claimset for unix users...


birojnayak:
This needs little bit research.. although most of the API from System.DirectoryServices.Protocols we can use, still need to investigate how open up the ldap conn with client username, fetch domain name and eventually do ldap search...

something in the line of https://github.com/dotnet/aspnetcore/blob/3fdbc2e90d9762324ef3809098c9ca25b0cb3284/src/Security/Authentication/Negotiate/src/Internal/LdapAdapter.cs#L13

Issue #91, opened 2020-01-03
Port ServiceAuthorizationBehavior class


mconnew:
Original code can be found [here](https://github.com/microsoft/referencesource/blob/master/System.ServiceModel/System/ServiceModel/Description/ServiceAuthorizationBehavior.cs).  
This will also need `AuthorizationBehavior` (found [here](https://github.com/microsoft/referencesource/blob/master/System.ServiceModel/System/ServiceModel/Dispatcher/AuthorizationBehavior.cs)) to be ported and the relevant calls added to the dispatcher.

dgxhubbard:
I can volunteer for #91 and #92.

dgxhubbard:
Note AuthorizationBehavior exists in src\CoreWCF.Primitives, but it does handle performance counters. Does that code need to be ported or is the one in CoreWCF ok?

mconnew:
Performance counters are a Windows only technology and requires some P/Invoke's to do correctly. There's some work happening in .NET Core to come up with a cross platform way of doing performance counters but I'm not up to date on where the current state is with it. I am strongly against having platform specific code (although I had to for IOThreadScheduler, but there's no P/Invokes there and it's all managed API calls and run-time switched so no multi-targeting required) as there should be Managed classes which provide a cross platform way to do things.

mconnew:
ServiceAuthorizationBehavior provides multiple capabilities. Let's take each property in turn.  

**ExternalAuthorizationPolicies**
[WCF Docs](https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.description.serviceauthorizationbehavior.externalauthorizationpolicies?view=netframework-4.8)  
Basically you create a List<IAuthorizationPolicy> and add an IAuthorizationPolicy instance to it. Then you wrap the list in a ReadOnlyCollection<IAuthorizationPolicy>. So something like this:
```c#
var authPolicies = new List<IAuthorizationPolicy>();
authPolicies.Add(new MyTestAuthPolicy());
var externalAuthPolicies = new ReadOnlyCollection<IAuthorizationPolicy>(authPolicies);
```
The documentation for IAuthorizationPolicy can be found [here](https://docs.microsoft.com/en-us/dotnet/api/system.identitymodel.policy.iauthorizationpolicy?view=netframework-4.8). Basically in the Evaluate method you create a new ClaimSet, add your own Claim (it's just metadata so make up a claim type) then add the ClaimSet to the EvaluationContext. 

[...]

PR #340, opened 2021-04-08
Enabling roles based permission for Linux and Windows (NetTCp and WsHttp)

birojnayak:
Fix for issues #335 #265 #91 (PrincipalPermissionMode alternative)

This is forward looking PR to enable authorization feature in CoreWCF. In WCF PrincipalPermissionMode (for roles based authorization) is used for same,  which behind the scene uses CAS. As per https://docs.microsoft.com/en-us/dotnet/framework/misc/code-access-security-basics , MS is not going to support and CAS has also limitation on across platforms.

Based on above we are suggesting another API replacement for PrincipalPermissionMode ( as below ), it uses the System.DirectoryServices.Protocol to retrieve user roles. 

 [AuthorizeRole(AllowedRoles = "CoreWCFGroupAdmin")]
 public string EchoForPermission(string echo)
        {
            return echo;
        }

This implements IAuthorizeOperation, which made public to have your custom authorization on operation basis.


src/CoreWCF.NetTcp/tests/ServiceContract/ITestService.cs
```diff
@@ -42,5 +42,13 @@ public interface ITestService
...
         string EchoForImpersonation(string echo);
+
+        [OperationContract(Name = "EchoForAuthorizarionOneRole", Action = Constants.OPERATION_BASE + "EchoForAuthorizarionOneRole",
+        ReplyAction = Constants.OPERATION_BASE + "EchoForAuthorizarionOneRoleResponse")]
+        string EchoForAuthorizarionOneRole(string echo);
+
+        [OperationContract(Name = "EchoForAuthorizarionNoRole", Action = Constants.OPERATION_BASE + "EchoForAuthorizarionNoRole",
+        ReplyAction = Constants.OPERATION_BASE + "EchoForAuthorizarionNoRoleResponse")]
+        string EchoForAuthorizarionNoRole(string echo);
     }
```
Eruanion, line 53: These appear to match exactly with ClientContract/ITestServices.cs, can we derive a base class to derive from to eliminate the same code or do you foresee these diverging in the future?

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        private readonly ITestOutputHelper _output;
+
+        public AuthorizationTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Fact(Skip ="Skipped in pipeline run")]
+        public void AuthorizationBasedonRolesTest()
```
mconnew, line 32: Nit: formatting

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                builder.AddService<Services.TestService>();
+                builder.AddServiceEndpoint<Services.TestService, ServiceContract.ITestService>(new CoreWCF.NetTcpBinding(), WindowsAuthRelativePath);
+                builder.AddServiceEndpoint<Services.TestService, ServiceContract.ITestService>(new CoreWCF.NetTcpBinding(SecurityMode.None), NoSecurityRelativePath);
+                Action<ServiceHostBase> serviceHost = host => ChangeHostBehavior(host);
+                builder.ConfigureServiceHostBase<Services.TestService>(serviceHost);
+            });
+        }
+        public void ChangeHostBehavior(ServiceHostBase host)
+        {
```
mconnew, line 89: Nit: formatting. Need blank line between methods

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        {
+            app.UseServiceModel(builder =>
+            {
+                builder.AddService<Services.TestService>();
+                builder.AddServiceEndpoint<Services.TestService, ServiceContract.ITestService>(new CoreWCF.NetTcpBinding(), WindowsAuthRelativePath);
+                builder.AddServiceEndpoint<Services.TestService, ServiceContract.ITestService>(new CoreWCF.NetTcpBinding(SecurityMode.None), NoSecurityRelativePath);
+                Action<ServiceHostBase> serviceHost = host => ChangeHostBehavior(host);
+                builder.ConfigureServiceHostBase<Services.TestService>(serviceHost);
```
mconnew, line 85: You don't need an intermediate variable. You should just be able to pass the method directly to ConfigureServiceHostBase.

src/CoreWCF.NetTcp/tests/ClientContract/ITestService.cs
```diff
@@ -43,5 +43,13 @@ public interface ITestService
         [OperationContract(Name = "EchoForImpersonation", Action = Constants.OPERATION_BASE + "EchoForImpersonation",
         ReplyAction = Constants.OPERATION_BASE + "EchoForImpersonationResponse")]
         string EchoForImpersonation(string echo);
+
+        [OperationContract(Name = "EchoForAuthorizarionOneRole", Action = Constants.OPERATION_BASE + "EchoForAuthorizarionOneRole",
+        ReplyAction = Constants.OPERATION_BASE + "EchoForAuthorizarionOneRoleResponse")]
```
mconnew, line 48: Typo, Autoriazarion -> Authorization

src/CoreWCF.NetTcp/tests/Services/TestService.cs
```diff
@@ -93,5 +93,17 @@ public string EchoForImpersonation(string value)
         {
             return value;
         }
+
+        [AuthorizeRole(AllowedRoles = "InvalidRole, CoreWCFGroupAdmin")]
+        public string EchoForAuthorizarionOneRole(string value)
```
mconnew, line 98: Is this the only way to provide multiple roles? Or can I apply the attribute multiple times? Ie:
```c#
        [AuthorizeRole(AllowedRoles = "InvalidRole")]
        [AuthorizeRole(AllowedRoles = "CoreWCFGroupAdmin")]
        public string EchoForAuthorizarionOneRole(string value)
        {
            return value;
        }
```

src/CoreWCF.Primitives/src/CoreWCF.Primitives.csproj
```diff
@@ -9,9 +9,11 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Server.Abstractions" Version="2.1.1" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.1.1" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
```
mconnew, line 13: What is this being used for?

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+using CoreWCF.Dispatcher;
+using CoreWCF.IdentityModel.Claims;
+
+namespace CoreWCF
+{
+    [AttributeUsage(CoreWCFAttributeTargets.OperationContract, Inherited = true)]
+    public class AuthorizeRoleAttribute : Attribute, IAuthorizeOperation
+    {
+        public string AllowedRoles { get; set; }
+
```
mconnew, line 17: I'm not 100% on the exact api shape here. It seems odd to have a property and not a constructor. Generally attribute properties are for optional value and there's normally a constructor or two for common scenarios. Only having a property seems to be going against convention. Having a constructor with a params argument would make the multi-role scenario simpler. So if you did this:
```c#
private string[] _allowedRoles;
public AuthorizeRoleAttribute(params string[] allowedRoles)
{
    _allowedRoles = allowedRoles;
}
```
Then in BuildClaim you can iterate on the array. Usage would be:
```c#
        [AuthorizeRole("InvalidRole", "CoreWCFGroupAdmin")]
        public string EchoForAuthorizarionOneRole(string value) { return value; }
```

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (String.IsNullOrEmpty(AllowedRoles))
+            {
+                return;
+            }
+            string[] roles = AllowedRoles.Split(',');
+            List<Claim> allClaims = new List<Claim>();
+            foreach(string role in roles)
+            {
+                allClaims.Add(new Claim(ClaimTypes.Role, role.Trim(), Rights.Identity));
+            }
```
mconnew, line 38: Either need a check with string.IsNullOrEmpty or use the overload of string.Split which takes a StringSplitOptions and specify RemoveEmptyEntries.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -397,9 +420,9 @@ private IEnumerable<Type> GetKnownTypes(object[] knownTypeAttributes, CustomAttr
             return knownTypes;
         }
 
-        private KeyedByTypeCollection<IOperationBehavior> GetIOperationBehaviorAttributesFromType(OperationDescription opDesc, Type targetIface, Type implType)
+        private KeyedByTypeCollection<IOperation> GetIOperationAttributesFromType<IOperation>(OperationDescription opDesc, Type targetIface, Type implType)
         {
```
mconnew, line 424: Generic types generally start with T. One line 425 it looks like there's an interface called IOperation that you are specifying. It makes the code a little harder to read.

src/CoreWCF.Primitives/src/CoreWCF.Primitives.csproj
birojnayak: This is used to keep the roles for same user https://github.com/CoreWCF/CoreWCF/pull/340/files#diff-0453087f6201cfcca1f04e109115308aebbb6dfb46fb806660071a9ef1c11d14R40

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
birojnayak: I like your idea... it's easy.. will shape accordingly

src/CoreWCF.NetTcp/tests/Services/TestService.cs
birojnayak: Not allowing multiple attributes at this time... just trying to make it similar to asp.net core

src/CoreWCF.NetTcp/tests/ServiceContract/ITestService.cs
birojnayak: that's right Jonathan ! But the objective here is to recreate real world scenario, where both contracts remain same across client/server... this gives us a good readability... in real world, customers just uses(mostly) the wsdl to generate client and they don't write much code for getting client contract..

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+            else
+            {
+                userAccountName = user.Substring(0, userAccountNameIndex);
+            }
+
+            List<Claim> roleClaims = new List<IdentityModel.Claims.Claim>();
+            if (settings.ClaimsCache == null)
+            {
+                settings.ClaimsCache = new MemoryCache(new MemoryCacheOptions { SizeLimit = settings.ClaimsCacheSize });
+            }
```
mconnew, line 38: This code feels like it should live inside LdapSettings. The settings owns the ClaimsCacheSize, but isn't currently responsible for instantiating the MemoryCache with the correct size. It feels like the wrong ownership of responsibility when a different class can instantiate the cache with a limit different than settings specify.

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            var distinguishedName = settings.Domain.Split('.').Select(name => $"dc={name}").Aggregate((a, b) => $"{a},{b}");
+            var retrievedClaims = new List<string>();
+            if (!string.IsNullOrEmpty(settings.OrgUnit))
+            {
+               distinguishedName = "OU=" + settings.OrgUnit + "," + distinguishedName;
+            }
+            var filter = $"(&(objectClass=user)(sAMAccountName={userAccountName}))"; // This is using ldap search query language, it is looking on the server for someUser
+            var searchRequest = new SearchRequest(distinguishedName, filter, SearchScope.Subtree, null);
```
mconnew, line 57: Is there a potential for escaping problems here? Is the value part of the query quotable at all? If so, is the quote value a valid account name character which means that would need to be escaped?

mconnew, line 57: This query isn't correct. See [this article](https://social.technet.microsoft.com/wiki/contents/articles/52250.active-directory-user-principal-name.aspx) for details. If the original passed in username had an @ in it, you should first search for userPrincipalName=userName. The userPrincipalName does NOT have a uniqueness constraint so if more than 1 result comes back, fail and return zero claims as you can't disambiguate. The userPrincipalName attribute isn't required, so zero results might come back. If that happens, then do the search that you have here.

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                settings.ClaimsCache = new MemoryCache(new MemoryCacheOptions { SizeLimit = settings.ClaimsCacheSize });
+            }
+
+            if (settings.ClaimsCache.TryGetValue<IEnumerable<string>>(user, out var cachedClaims))
+            {
+                foreach (var claim in cachedClaims)
+                {
+                    roleClaims.Add(new Claim(ClaimTypes.Role, claim, Rights.Identity));
+                }
```
mconnew, line 45: Is ClaimTypes.Role the type of claim that gets populated on Windows? I think you probably want the cache to be for `IEnumerable<Claim>` as you want the claimset to include things like the Upn is the ldap record has a userPrincipalName attribute, and a Name claim to hold the sAMAccountName attribute.

mconnew: Thinking about this a little more, instead of `IEnumerable<Claim>`, it should be a `ClaimSet`.

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
mconnew: From the article I listed elsewhere, single quotes are valid values inside of a username. Will that potentially cause a problem for the query?

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+                    var groupDN = $"{Encoding.UTF8.GetString((byte[])group)}";
+                    var groupCN = groupDN.Split(',')[0].Substring("CN=".Length);
+                    retrievedClaims.Add(groupCN);
+                }
+
+                var entrySize = user.Length * 2; //Approximate the size of stored key in memory cache.
+                foreach (var claim in retrievedClaims)
+                {
+                    roleClaims.Add(new Claim(ClaimTypes.Role, claim, Rights.Identity));
+                    entrySize += claim.Length * 2; //Approximate the size of stored value in memory cache.
```
mconnew, line 93: Look at WindowsClaimSet.InitializeClaimsCore to see what set of claims is expected

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/WindowsSecurityTokenAuthenticator.cs
```diff
@@ -31,6 +38,14 @@ protected override ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(Se
         {
             var windowsToken = (WindowsSecurityToken)token;
             var claimSet = new WindowsClaimSet(windowsToken.WindowsIdentity, windowsToken.AuthenticationType, _includeWindowsGroups, windowsToken.ValidTo);
+            if (_ldapSettings != null && claimSet.Count > 0)
+            {
```
mconnew, line 42: On .NET Framework, it's WindowsClaimSet which handles retrieving all the claims. I think it would be worth adding an overload to the constructory to pass the _ldapSettings instead and have that do the work. That way the code to call RetrieveClaims doesn't need to be duplicated.

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapSettings.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// Validates the <see cref="LdapSettings"/>.
+        /// </summary>
+        public void Validate()
+        {
+            if (EnableLdapClaimResolution)
+            {
+                if (Servers == null || Servers.Length == 0)
+                {
+                    throw new ArgumentException($"{nameof(EnableLdapClaimResolution)} is set to true but ldap server not set.");
+                }
```
mconnew, line 160: Add this to string resources

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            var distinguishedName = settings.Domain.Split('.').Select(name => $"dc={name}").Aggregate((a, b) => $"{a},{b}");
+            var retrievedClaims = new List<string>();
+            if (!string.IsNullOrEmpty(settings.OrgUnit))
+            {
+               distinguishedName = "OU=" + settings.OrgUnit + "," + distinguishedName;
+            }
+            var filter = $"(&(objectClass=user)(sAMAccountName={userAccountName}))"; // This is using ldap search query language, it is looking on the server for someUser
+            var searchRequest = new SearchRequest(distinguishedName, filter, SearchScope.Subtree, null);
+            SearchResponse searchResponse = null;
```
mconnew, line 58: This should be executed async as it's going off box. There's a BeginSendRequest and EndSendRequest api pair that you should be able to use with Task.Factory.FromAsync.

src/CoreWCF.NetTcp/tests/ClientContract/ITestService.cs
birojnayak: can't find

mconnew: Sorry, I should have copy and pasted it instead of trying to retype it. The incorrect text says Authorizarion

OperationContract(Name = "EchoForAuthori**zarion**OneRole

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
birojnayak: If you are asking in aspnet core, yes https://github.com/dotnet/aspnetcore/blob/52eff90fbcfca39b7eb58baad597df6a99a542b0/src/Identity/Extensions.Core/src/ClaimsIdentityOptions.cs#L16 .. but not in WCF... WCF handles via CAS.. while aspnetcore handles via CalimsIdentity for both os

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/WindowsSecurityTokenAuthenticator.cs
birojnayak: I went through the code with approach discussed but found the problem is WindowsClaimSet is very much tied with WindowsIdentity, and we can't initialize WindowsIdentity with SPN, we can do by UPN only. I can make changes inside WindowsClaimSet to act differently when it's WIndows/Linux, but that would lead to lot of if/else code for various OS and to me it won't be nice... so I would suggest to have GenericIdentity as we had already decided and in the similar line in .net core I believe.. if you think otherwise, we can discuss and close on this... keeping open to discuss..

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
birojnayak: Not sure... I am following what's done in aspnetcore https://github.com/dotnet/aspnetcore/blob/3fdbc2e90d9762324ef3809098c9ca25b0cb3284/src/Security/Authentication/Negotiate/src/Internal/LdapAdapter.cs#L41

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
birojnayak: It would be difficult to completely replicate as we will hit CAS.. so I am taking guidance from aspnetcore to make it more generic as discussed... let me know your thought

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -189,13 +256,13 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
                 {
                     if (right == null || Rights.Identity == right)
                     {
-                        yield return new Claim(ClaimTypes.Sid, _windowsIdentity.User, Rights.Identity);
+                        yield return new Claim(ClaimTypes.Sid, _windowsIdentity, Rights.Identity);
                     }
```
mconnew, line 260: claimType of Sid is only valid if it's a WindowsIdentity object. If a ClaimTypes.Sid is requested, this will return the wrong thing. You need to make the SupportedClaimType return false is it's not a WindowsIdentity and requesting Sid or DenyOnlySid. Then in this code it needs to be cast as WindowsIdentity and fetch the User property which is the Sid.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -189,13 +256,13 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
...
                 }
 
                 if (right == null || Rights.PossessProperty == right)
                 {
-                    if (TryCreateWindowsSidClaim(_windowsIdentity, out Claim sid))
+                    if (_windowsIdentity is WindowsIdentity && TryCreateWindowsSidClaim((WindowsIdentity)_windowsIdentity, out Claim sid))
                     {
```
mconnew, line 266: If you make SupportedClaimType return false when it's not a WindowsIdentity, then you don't need the is check here.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -122,29 +175,37 @@ public void Dispose()
...
+                {
+                    claims.Add(claim);
+                }
+                claims.Add(Claim.CreateNameClaim(_windowsIdentity.Name));
+            }
+            else
             {
-                claims.Add(claim);
+                claims.Add(Claim.CreateNameClaim(_windowsIdentity.Name));
             }
```
mconnew, line 204: This doesn't need to be in an else block as you also do it at the end of the true if block. Just move this outside of the if block and don't have an else.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -252,5 +319,42 @@ public static bool TryCreateWindowsSidClaim(WindowsIdentity windowsIdentity, out
...
+                                base.Add(new Claim(ClaimTypes.Role, group, Rights.Identity));
+                            }
+                        }
+                    }
+                }
+                else if(ldapSettings !=null)
+                {
+                    List<Claim> allClaims = LdapAdapter.RetrieveClaimsAsync(ldapSettings, claimsIdentity.Name).GetAwaiter().GetResult();
+                    //should we make above sync -- this would be in the Open path..
+                    foreach(Claim roleClaim in allClaims)
```
mconnew, line 351: This is a difficult one. Most of the time we don't need to execute this code, whic is why Groups is lazily created. We don't want to do the work and allocate the memory if it's not needed. But as this requires off box work, if needed it would be a good idea to kick off fetching this as soon as possible in anticipation. I see 2 possibilities:  
1. Leave it as is and let it block on async work
2. Initially have this behavior where it blocks on async work, but set a static boolean field recording that this feature was used in the process so one the second usage of this class we pre-emptively kick this off asynchronously and have the result available as a Task so that hopefully it's completed by the time it's actually needed.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -12,11 +14,13 @@ namespace CoreWCF.IdentityModel.Claims
     internal class WindowsClaimSet : ClaimSet, IIdentityInfo, IDisposable
     {
         internal const bool DefaultIncludeWindowsGroups = true;
-        private readonly WindowsIdentity _windowsIdentity;
+        private readonly ClaimsIdentity _windowsIdentity;
         private readonly bool _includeWindowsGroups;
         private IList<Claim> _claims;
         private bool _disposed = false;
         private readonly string _authenticationType;
+        GroupSidClaimCollection groups;
```
mconnew, line 22: groups should be named _groups

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -49,20 +53,57 @@ internal WindowsClaimSet(WindowsIdentity windowsIdentity, string authenticationT
...
+        {
+            if (claimsIdentity == null)
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(claimsIdentity));
+            }
+            _windowsIdentity = claimsIdentity;
+        }
+
+       internal WindowsClaimSet(ClaimsIdentity claimsIdentity, bool includeWindowsGroups,  LdapSettings ldapSettings)
+        : this(claimsIdentity, null, includeWindowsGroups, DateTime.UtcNow.AddHours(10), false, null)
```
mconnew: Nit: identation

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -49,20 +53,57 @@ internal WindowsClaimSet(WindowsIdentity windowsIdentity, string authenticationT
...
+                foreach (Claim claim in _fromClaims)
+                {
+                    allClaims.Add(claim);
+                }
+                _claims = allClaims;
+            }
         }
 
+
         private WindowsClaimSet(WindowsClaimSet from)
```
mconnew, line 105: Extra blank line

mconnew:
>                     // _includeWindowsGroups is true; don't throw here or we bust the default case on UWP/.NET Core

It looks like this was copied from the WCF client. There's code missing here which wasn't important on client but is more so here. Can you restore that code?

---
Refers to: src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs:277 in 52c54c4. [](commit_id = 52c54c428f6142133a0ff6b9fa8ce1a0c40aaa43, deletion_comment = False)

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
birojnayak: going with option 1

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/NetMessageFramingConnectionHandler.cs
```diff
@@ -127,7 +128,13 @@ private static HandshakeDelegate BuildHandshakeDelegateForDispatcher(IServiceDis
...
             {
-                var bindingContext = new BindingContext(new CustomBinding(dispatcher.Binding), new BindingParameterCollection());
+                var parameter = new BindingParameterCollection();
+                SecurityCredentialsManager credentialsManager = dispatcher.Host.Description.Behaviors.Find<SecurityCredentialsManager>();
+                if (credentialsManager != null)
+                {
+                    parameter.Add(credentialsManager);
+                }
+                var bindingContext = new BindingContext(new CustomBinding(dispatcher.Binding), parameter);
                 streamUpgradeProvider = upgradeBindingElements[0].BuildServerStreamUpgradeProvider(bindingContext);
```
mconnew, line 138: Blank line after closing brace at end of if block

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                        new System.ServiceModel.EndpointAddress(host.GetNetTcpAddressInUse() + Startup.WindowsAuthRelativePath));
+                    channel = factory.CreateChannel();
+                    ((IChannel)channel).Open();
+                    string result = channel.EchoForAuthorizarionOneRole(testString);
+                    Assert.Equal(testString, result);
+                    try
+                    {
+                        channel.EchoForAuthorizarionNoRole(testString);
+                    }catch(Exception ex)
+                    {
```
mconnew, line 54: Formatting

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                        Assert.Contains("Access is denied", ex.Message);
+                    }
+                    ((IChannel)channel).Close();
+                    factory.Close();
```
mconnew, line 58: Blank line after closing brace of try/catch

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                    ((IChannel)channel).Open();
+                    string result = channel.EchoForAuthorizarionOneRole(testString);
+                    Assert.Equal(testString, result);
+                    try
+                    {
+                        channel.EchoForAuthorizarionNoRole(testString);
+                    }catch(Exception ex)
+                    {
+                        Assert.Contains("Access is denied", ex.Message);
+                    }
```
mconnew, line 56: Do you always expect this exception to be thrown? If the method doesn't throw an exception, then the Assert doesn't get hit and the test fails. Should this be Assert.ThrowsAny?

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                builder.AddServiceEndpoint<Services.TestService, ServiceContract.ITestService>(new CoreWCF.NetTcpBinding(SecurityMode.None), NoSecurityRelativePath);
+                Action<ServiceHostBase> serviceHost = host => ChangeHostBehavior(host);
+                builder.ConfigureServiceHostBase<Services.TestService>(serviceHost);
+            });
+        }
+        public void ChangeHostBehavior(ServiceHostBase host)
+        {
+            var srvCredentials = new CoreWCF.Description.ServiceCredentials();
+            LdapSettings _ldapSettings = new LdapSettings("ldapserver.test.local", "test.local", "your_own_top_org");
+            srvCredentials.WindowsAuthentication.LdapSetting = _ldapSettings;
```
mconnew, line 92: Add comment specifying that this needs to be configured for local LDAP config when doing local testing.

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
mconnew, line 32: Formatting (need space after =), also can you update Skip comment to say needs to be running on domain joined machine.

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(operationDescription));
+            }
+
+            if (dispatchOperation == null)
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(dispatchOperation));
+            }
+            if (_allowedRoles == null || _allowedRoles.Length ==0)
+            {
```
mconnew, line 35: I think you can do _allowedRoles?.Length == 0. Should have space after ==. Also blank line after if blocks and after foreach block. Also extra blank line at end of method.

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+                return;
+            }
+            List<Claim> allClaims = new List<Claim>();
+            foreach(string role in _allowedRoles)
```
mconnew, line 39: Use constructor overload which accepts an initial size and set to _allowedRoles.Length

src/CoreWCF.Primitives/src/CoreWCF/AuthorizeRoleAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                allClaims.Add(new Claim(ClaimTypes.Role, role.Trim(), Rights.Identity));
+            }
```
mconnew, line 42: Check that role is not null or the empty string.

src/CoreWCF.Primitives/src/CoreWCF/IAuthorizeOperation.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using CoreWCF.Description;
+using CoreWCF.Dispatcher;
+
+namespace CoreWCF
+{
+    public interface IAuthorizeOperation //Any good name would be helpful
+    {
```
mconnew, line 10: Remove comment. This name is fine. I think this interface should be in the CoreWCF.Description namespace as should the AuthorizeRoleAttribute class as they are involved in annotating and describing the service.

src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -49,8 +50,12 @@ internal class WindowsStreamSecurityUpgradeProvider : StreamSecurityUpgradeProvi
                 //}
             }
 
-
-            _securityTokenManager = credentialProvider.CreateSecurityTokenManager();
+           if(credentialProvider is ServiceCredentials)
+            {
```
mconnew, line 54: Identation

src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -85,6 +90,8 @@ public override EndpointIdentity Identity
 
         private NetworkCredential ServerCredential { get; set; }
 
+        protected LdapSettings LdapSettings { get; private set; }
+
```
mconnew, line 94: Why protected? This class in internal.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -647,6 +647,11 @@ private static void BindOperations(ContractDescription contract, ClientRuntime p
                             IOperationBehavior behavior = operation.Behaviors[j];
                             behavior.ApplyDispatchBehavior(operation, dispatchOperation);
                         }
+                        for (int k = 0; k < operation.AuthorizeOperation.Count; k++)
+                        {
```
mconnew, line 651: Blank line between for blocks. Also I think foreach would result in more compact code

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -83,7 +83,7 @@ private ContractDescription LoadContractDescriptionHelper(Type contractType, obj
                         }
                         UpdateOperationsWithInterfaceAttributes(contractDescription, reflectionInfo);
                         AddBehaviors(contractDescription, false, reflectionInfo);
-
+                        AddAuthorizeOperations(contractDescription,false, reflectionInfo);
                         _contracts.Add(actualContractType, contractDescription);
```
mconnew, line 87: Formatting, space after comma

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -204,6 +204,28 @@ private ContractDescription LoadRequestChannelContractDescription()
             return contract;
         }
 
+        //This needs proper review. 
+        private void AddAuthorizeOperations(ContractDescription contractDesc, bool implIsCallback, ContractReflectionInfo reflectionInfo)
+        {
+
+            for(int i=0; i< contractDesc.Operations.Count; i++)
```
mconnew, line 211: Extra blank line

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -204,6 +204,28 @@ private ContractDescription LoadRequestChannelContractDescription()
...
+            {
+                OperationDescription opDesc = contractDesc.Operations[i];
+                Type targetIface = implIsCallback ? opDesc.DeclaringContract.CallbackContractType : opDesc.DeclaringContract.ContractType;
+                ApplyServiceInheritance(
+                    opDesc.AuthorizeOperation,
+                    delegate (Type currentType, KeyedByTypeCollection<IAuthorizeOperation> behaviors)
+                    {
```
mconnew, line 218: This could benefit from using a local function. See [here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions) for details.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -204,6 +204,28 @@ private ContractDescription LoadRequestChannelContractDescription()
             return contract;
         }
 
+        //This needs proper review. 
+        private void AddAuthorizeOperations(ContractDescription contractDesc, bool implIsCallback, ContractReflectionInfo reflectionInfo)
```
mconnew, line 208: Remove comment before this is merged

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -204,6 +204,28 @@ private ContractDescription LoadRequestChannelContractDescription()
...
+        {
+
+            for(int i=0; i< contractDesc.Operations.Count; i++)
+            {
```
mconnew, line 212: Formatting around mathematical operators. Would a foreach work better here?

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/GenericSecurityTokenAuthenticator.cs
```diff
@@ -21,7 +31,46 @@ protected override bool CanValidateTokenCore(SecurityToken token)
         protected override ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(SecurityToken token)
         {
             var genericToken = (GenericSecurityToken)token;
-            return CoreWCF.Security.SecurityUtils.CreatePrincipalNameAuthorizationPolicies(genericToken.Name);
+            string principalName = genericToken.Name;
+            if (principalName == null)
+            {
```
mconnew, line 36: Should this be String.IsNullOrEmpty?

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/GenericSecurityTokenAuthenticator.cs
```diff
@@ -21,7 +31,46 @@ protected override bool CanValidateTokenCore(SecurityToken token)
...
+                identityClaim,
+                primaryPrincipal
+            };
+
+
+            if (_ldapSettings != null)
+            {
+                List<Claim> allCaims = LdapAdapter.RetrieveClaimsAsync(_ldapSettings, genericToken.GenericIdentity.Name).Result;
+                // if this is made async, many other API changes has to happen. COnsidering this is one of the scenario, ok to take the hit ?
+                foreach (Claim claim in allCaims)
```
mconnew, line 64: This is one of the improvements we should make to the api's. Can you open an issue to make SecurityTokenAuthenticator async?

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                if (upnIndex > 0)
+                {
+                    searchResponse = (SearchResponse)await Task<DirectoryResponse>.Factory.FromAsync(
+                    settings.LdapConnection.BeginSendRequest, settings.LdapConnection.EndSendRequest,
+                    upnSearchRequest, PartialResultProcessing.NoPartialResultSupport, null);
+                    if(searchResponse !=null && searchResponse.Entries != null && searchResponse.Entries.Count > 1)
+                    {
+                        throw new Exception(SR.DuplicateUPN); //resource
+                    }
```
mconnew, line 67: We need to find a better exception than just Exception. Maybe SecurityNegotiationException? Maybe a new exception type needs to be created?

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                var userFound = searchResponse.Entries[0]; //Get the object that was found on ldap
+                var memberof = userFound.Attributes["memberof"]; // You can access ldap Attributes with Attributes property
+
+                foreach (var group in memberof)
+                {
+                    // Example distinguished name: CN=TestGroup,DC=KERB,DC=local
+                    var groupDN = $"{Encoding.UTF8.GetString((byte[])group)}";
+                    var groupCN = groupDN.Split(',')[0].Substring("CN=".Length);
+                    retrievedClaims.Add(groupCN);
```
mconnew, line 99: There should be a bit more validation here. Make sure group isn't null, make sure groupDn isn't the empty string. Make sure the first entry does begin with `CN=`. Also I believe you can have leading or trailing spaces in many places, so `CN = TestGroup, DC = KERB, DC = local` is technically a valid distinguished name. You need to split the groupDN on a comma, then split the first field on the equals sign and take the second half (after checking it split into exactly 2 pieces) and Trim() it to remove and leading or trailing spaces.

src/CoreWCF.Primitives/src/CoreWCF/Security/SecurityUtils.cs
```diff
@@ -1098,7 +1098,8 @@ public static void ValidateAnonymityConstraint(WindowsIdentity identity, bool al
                     new SecurityTokenValidationException(SR.AnonymousLogonsAreNotAllowed));
             }
         }
-        internal static ReadOnlyCollection<IAuthorizationPolicy> CreatePrincipalNameAuthorizationPolicies(string principalName)
+
+        /*internal static ReadOnlyCollection<IAuthorizationPolicy> CreatePrincipalNameAuthorizationPolicies(string principalName, LdapSettings ldapSettings)
         {
```
mconnew, line 1103: If this is dead code, please remove it

src/CoreWCF.Primitives/src/Resources/Strings.resx
```diff
@@ -8038,4 +8038,16 @@
   <data name="TimeSpanMustBeGreaterThanTimeSpanZero" xml:space="preserve">
     <value>TimeSpan must be greater than TimeSpan.Zero.</value>
   </data>
+    <data name="EnableLdapClaimResolutionServer" xml:space="preserve">
+    <value>'{0}' is set to true but ldap server not set.</value>
```
mconnew, line 8042: Nit: indentation is incorrect. Tabs vs spaces?

src/CoreWCF.NetTcp/tests/AuthorizationTest.cs
birojnayak: This is expected to be thrown. I wanted to validate the message.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
birojnayak: need little more explanation on how ?

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/GenericSecurityTokenAuthenticator.cs
birojnayak: https://github.com/CoreWCF/CoreWCF/issues/375

src/CoreWCF.Primitives/src/CoreWCF/Security/LdapAdapter.cs
birojnayak: SecurityNegotiationException

mconnew:
@birojnayak, something went wrong with your merging. This PR effectively has zero changes.

mconnew:
Closing in preference of #382

PR #348, opened 2021-04-17
Add XML Docs to CoreWCF.Channels

Grauenwolf:
Added XML Docs to the public classes in CoreWCF.Channels


mconnew:
Thank you

Grauenwolf:
You're welcome. I'll try to do more as I have time to read the code.

PR #349, opened 2021-04-19
Use AsyncManualResetEvent in ChannelHandler, remove ResettableAsyncWaitable

BradBarnich:
In doing some load testing for a migration, I found that a very small percentage of requests timed out. Catching it in a debugger, the long request was stuck on the await on https://github.com/CoreWCF/CoreWCF/compare/main...BradBarnich:deadlock?expand=1#diff-b59ad09365ce977d3b5d54beb9f8c4396253e434b320f7bcb7805e58f755f1c6L1018

I think something is not quite correct in the CompareExchange code of ResettableAsyncWaitable. Luckily there is another implementation of an async reset event in CoreWCF, https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Primitives/src/CoreWCF/Runtime/AsyncManualResetEvent.cs 

The issue when away when I switched to that.


mconnew:
Thank you

PR #357, opened 2021-05-10
Adding IPAddress overload for net.tcp

### LINKED ISSUE #342: IP address selection with NetTcp binding
IWebHostBuilder.UseNetTcp() listens on the specified port on all IP addresses (IPAddress.Any), it would be nice if overload methods are added to allow selection of IP addresses.

ovebastiansen:
Added an overload to be able to specify IPAddress when adding net.tcp setup.

Because of the change in OnServiceBuilderOpening I am not sure if IPAddress.Any is the correct default, since it was Loopback before?

closes #342


ovebastiansen:
@mconnew any feedback on this?

mconnew:
The test specifying the specific port number is going to be fragile as sometimes that port will just happen to be in use. Can you also add an additional overload with this signature and use it in the test. Sometimes you want to specify the ip address such as localhost to prevent external hosts from connecting, but you just don't care what port it's running on. This should only be a minor change to your PR.
```c#
public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder, IPAddress ipAddress);
```

ovebastiansen:
What port should that new overload use, default 808 or something random?

mconnew:
Yes, port 808 as that's the default.

ovebastiansen:
added that overload

mconnew:
@ovebastiansen, closing this in preference of #383 as this branch has conflicts. I cherry-picked the changes so they maintained your authorship on the commits.

PR #359, opened 2021-05-11
Pretty sure this is wrong, but cannot reproduce in a test

ovebastiansen:
I tried adding a test, but cannot get into that method with an address that looks like any of those. So fix is either to totally remove this code, if it is not needed or just update that 8 -> 9 in substring

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
@@ -251,7 +251,7 @@ namespace CoreWCF
                      address.StartsWith("https://*:", StringComparison.OrdinalIgnoreCase) ||
                      address.StartsWith("https://*/", StringComparison.OrdinalIgnoreCase))
             {
-                address = "https://localhost" + address.Substring(8);
+                address = "https://localhost" + address.Substring(9);
             }
 
             return new Uri(address);


ovebastiansen:
@mconnew do you have any idea on how to get into this method with + or * to test this, or should I just wipe out the tests for this and fix the substring?

mconnew:
This code path gets hit when using the HttpSys based HTTP host. We have a test for this for HTTP but not HTTPS because of the extra out of proc work needed to configure Http.Sys for a service certificate. Basically need to use the commandline tool netsh to configure the certificate on a port.

ovebastiansen:
I tried all kinds of address settings in both kestrel and http.sys, shipping in a + in useurls and those things, but it was always replaced with [::], or 0.0.0.0 or localhost for instance when that code was running. So could not reproduce the need for that function at all. happy to take pointers on how to get into that function with a + or * in the url? So getting into that function is quite easy in tests, but not with those values

mconnew:
I was going to try to get this in for the upcoming release, but there's some complications. It looks like for HttpSys we do actually need BasicHttp**s**Binding as the address is passed along as an https:// address and we look up base addresses based on the binding declared Scheme, which is http and not https.

ovebastiansen:
the problem is the substring(8) in the https case, you are also substring(8) in http case, so that means the address returned from the else would be wrong
https://localhost + "https://+/method".Substring(8) => https://localhost/+method and not https://localhost/method
right?

mconnew:
I understand the issue. The problem you were having with writing tests is reaching this code. I reached out to a colleague in the asp.net core team to find the scenario where + and * end up in the base addresses. He said there needs to be a request queue registered in HTTP.SYS using the netsh command line tool and the + or * has to be used in that registration. The next problem is that the scheme for BasicHttpBinding is HTTP and not HTTPS. We haven't ported BasicHttpsBinding because the S part of HTTPS was only configured in Kestrel (as we don't set the listening address) so it didn't matter, we could just treat it as HTTP. The HttpSys server implementation is different and I do need to make BasicHttpsBinding available so the scheme matches. Until that happens, the fix is technically unreachable code. Because of the need to register a urlacl in HTTP.SYS (which needs to be done as an admin), it can't be an always run test. The test will need to be manual run only and have some comments explaining what commands to run to set up and clean up the request queue registration.

ovebastiansen:
ok, I understand now, just your previous post made me wonder ;)
So what should we do now? pretty sure anybody else stumbling onto that code will start wondering. I can kill the tests and just have the simple code fix in, since like you said, currently dead, but misleading code?

mconnew:
Yes, please do that.

ovebastiansen:
finally got around to drop the tests for this, also rebased to make sure there are no conflicts on against main

Issue #353, opened 2021-04-29
Using an existing instance of an object as a service should default to InstanceBehavior.Single


pgrawehr:
I want the server to serve an existing instance of an object (because the object has a state already). Let's try it with an example:

In `Services.EchoService` add a ctor and change the implementation as follows:

```csharp
public class EchoService : ServiceContract.IEchoService
    {
        private readonly string _prefix;

        public EchoService(string prefix)
        {
            _prefix = prefix;
        }
        public string EchoString(string echo)
        {
            return _prefix + echo;
        }
        // snip rest of class
   }
```

If I run the `BasicHttp.SimpleTest` now I get an error `System.InvalidOperationException: The service type provided could not be loaded as a service because it does not have a default (parameter-less) constructor. To fix the problem, add a default constructor to the type, or pass an instance of the type to the host.` This is expected. Now let's do the second suggestion. 

If in the test I change 
```csharp
 [Fact]
        public void BasicHttpRequestReplyEchoString()
        {
            string testString = new string('a', 3000);
            var builder = ServiceHelper.CreateWebHostBuilder<Startup>(_output);
            EchoService echoService = new EchoService("Prefix ");
            var descriptor = new ServiceDescriptor(typeof(EchoService), echoService);
            builder.ConfigureServices(a => a.Add(descriptor));
            IWebHost host = builder.Build();
            using (host)
            {
                host.Start();
                System.ServiceModel.BasicHttpBinding httpBinding = ClientHelper.GetBufferedModeBinding();
                var factory = new System.ServiceModel.ChannelFactory<ClientContract.IEchoService>(httpBinding,
                    new System.ServiceModel.EndpointAddress(new Uri("http://localhost:8080/BasicWcfService/basichttp.svc")));
                ClientContract.IEchoService channel = factory.CreateChannel();
                string result = channel.EchoString(testString);
                Assert.Equal(testString, result);
            }
        }
```

The test now fails the "Assert.Equal()" because the strings don't match. So far, so good, exactly what we need. 

But: Now I add `IDisposable` to EchoService:

```csharp
public void Dispose()
        {
            _prefix = "Disposed";
        }
```

and find that this is called _during host construction_. As a result, the `result` of our `EchoString()` request now contains the text "Disposed". Bad. Looking trough the call stack at the time of the Dispose call, I find that our EchoService has `InstanceContextMode.PerSession` set. 

### Expected Behavior
Services registered with 
```csharp
[...]

PR #362, opened 2021-05-19
Cleanup DI handling, fix allowing ServiceBehaviorAttribute via DI

### LINKED ISSUE #355: DI is checked for instances of IServiceBehavior after we ensure there's a ServiceBehaviorAttribute
This is preventing injecting a ServiceBehaviorAttribute via DI. The logic needs to be changed to check for IServiceBehavior instances in DI before we check if we have an instance of ServiceBehaviorAttribute in the set of behaviors.
This was discovered as part of #353

mconnew:
Fixes #355
Fixes other issues brought up in #353

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
@@ -4,7 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using CoreWCF.Collections.Generic;
+using CoreWCF.Dispatcher;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace CoreWCF.Description
 {
@@ -77,10 +80,23 @@ namespace CoreWCF.Description
 
         public Type ServiceType { get; set; }
 
-        internal static void AddBehaviors<TService>(ServiceDescription serviceDescription) where TService : class
+        internal static void AddBehaviors<TService>(ServiceDescription serviceDescription, IEnumerable<IServiceBehavior> injectedBehaviors = null) where TService : class
         {
             TypeLoader<TService>.ApplyServiceInheritance<IServiceBehavior, KeyedByTypeCollection<IServiceBehavior>>(
                 serviceDescription.Behaviors, GetIServiceBehaviorAttributes);
+
+            if (injectedBehaviors != null)
+            {
+                // Only add IServiceBehavior from DI if concrete type not already added
+                // TODO: Add logging to state when an injected behavior has been skipped
+                foreach(var behavior in injectedBehaviors)
+                {
+                    if (!serviceDescription.Behaviors.Contains(behavior.GetType()))
+                    {
+                        serviceDescription.Behaviors.Add(behavior);
+                    }
+                }
+            }
 
             ServiceBehaviorAttribute serviceBehavior = EnsureBehaviorAttribute(serviceDescription);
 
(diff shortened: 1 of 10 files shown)


src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
```diff
@@ -77,11 +80,24 @@ public string ConfigurationName
 
         public Type ServiceType { get; set; }
 
-        internal static void AddBehaviors<TService>(ServiceDescription serviceDescription) where TService : class
+        internal static void AddBehaviors<TService>(ServiceDescription serviceDescription, IEnumerable<IServiceBehavior> injectedBehaviors = null) where TService : class
         {
             TypeLoader<TService>.ApplyServiceInheritance<IServiceBehavior, KeyedByTypeCollection<IServiceBehavior>>(
                 serviceDescription.Behaviors, GetIServiceBehaviorAttributes);
 
+            if (injectedBehaviors != null)
```
birojnayak, line 88: how this is different from ConfigureServiceHostBase ?

mconnew: ConfigureServiceHostBase allows you to provide a delegate which is passed the ServiceHostBase instance for you to modify. This is the mechanism of allowing you to add an implementation of IServiceBehavior to DI and have that automatically added to the description. The problem was that the DI behaviors were being added too late which meant if you tried to add a ServiceBehaviorAttribute to DI, it would fail because the automatic add if missing logic had already ran.

src/CoreWCF.Http/tests/Services/TestBasicScenariosService.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Services
 {
-    [ServiceBehavior]
+    //[ServiceBehavior]
```
birojnayak, line 10: you want this behavior to be there ?

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
```diff
@@ -231,6 +243,26 @@ private static void GetIServiceBehaviorAttributes(Type currentServiceType, Keyed
             }
         }
 
+        internal static void SetupSingleton<TService>(ServiceDescription serviceDescription, IServiceProvider services) where TService : class
+        {
+            ServiceBehaviorAttribute serviceBehavior = serviceDescription.Behaviors.Find<ServiceBehaviorAttribute>();
+            Debug.Assert(serviceBehavior != null, "EnsureServiceBehavior should have ensured the serviceBehavior");
```
birojnayak, line 249: should we these message to resource.xml or some format... or you want to keep as it it

src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
```diff
@@ -52,69 +54,44 @@ internal override object DisposableInstance
...
-                else
-                {
-                    serviceBehavior.InstanceProvider = new DependencyInjectionInstanceProvider(_serviceProvider, typeof(TService));
-                    if (serviceInstanceUsedAsABehavior == null && instance is IDisposable disposable)
-                    {
-                        disposable.Dispose();
-                    }
-                }
+                // If using Single, then ServiceBehavior fetched/created the instance and need to set on SingletonInstance
+                Debug.Assert(serviceInstanceUsedAsABehavior != null, "Service behavior should have created a singleton instance");
```
birojnayak, line 84: same as above.. for these message, should we leave as is

birojnayak (approved):

src/CoreWCF.Http/tests/Services/TestBasicScenariosService.cs
mconnew: Oops, that was left over from before I had explicit tests. I'll fix that.

Issue #129, opened 2020-04-23
Support ConfigurationManager based config for sharing with .NET Framework


twsouthwick:
I put together a POC a while back on incorporating ConfigurationManager to work for configuration of clients so that code could be shared between .NET Framework and Core easier. I'm opening this to see if there is interest. I put this together before CoreWCF was a thing and not sure if it would be an easy integration or not.

Here's the link: https://github.com/twsouthwick/ServiceModel.Configuration

The API looks like this:

```csharp
services.AddServiceModelClient()
  .AddConfigurationManagerFile("wcf.config")
  .AddDefaultChannel<IRoleService>("roles");
```

It uses ASP.NET Core IOptions and builds up the services so they can be directly injected. I only know the basics of WCF, though, so not sure how well this would extend to other use cases.

Feel free to close if there's no interest in it. Much of the configuration code for this came from ReferenceSource and has been `#if/#def`'d for types that didn't exist when I was playing with it.

mconnew:
@twsouthwick, the api's are stable around the bindings, would you be interested in making this work for CoreWCF and contributing a PR?

Ximik87:
Don't you mind me to do this task, using code based on this repository?

mconnew:
Consider it yours!! The first thing you do is an api proposal. This would include the class and method definitions with no implementation filled out and a code snippet showing how this would be used. Some questions that need to be answered are:  
1. Is the feature opt in by adding something to the ConfigureServices method or will it piggyback on the AddServiceModelServices?  
2. How much is going to be implicit? For example, in WCF you can new up a ServiceHost for your service type and all the bindings can automatically be populated from config. Is that going to be the model or will AddServiceEndpoint still be used with a Binding factory to provide the binding? Or will it be more deeply integrated and support named configuration constructor on the Binding?  

I would like to have the Binding itself not know anything about configuration and it should be decoupled. The reason being long term I want there to be multiple configuration options, including pulling it from JSON using the .NET Core configuration object binding feature. If you want to support passing a configuration name to the binding, this could still be done decoupled. For example, there's a Binding.Name string property which a constructor could set. Then you could do a post process step where that name is used to lookup the binding configuration. This would keep a similar experience to WCF but still keep it decoupled.  

I'm not going to be super dogmatic about how the api should look, but it should fit in with the ASP.NET Core general patterns.

[...]

PR #366, opened 2021-05-28
Adding support ConfigurationManager based config

Ximik87:
This feature adding support ConfigurationManager config files from .Net Framework. (Issues #129 )
Now is supported (basic scenario):
BindingsSection - BasicHttpBinding, NetHttpBinding, NetTcpBinding, WSHttpBinding
ServicesSection - Service and endpoint


src/CoreWCFWithSamples.sln
```diff
@@ -40,6 +40,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "StandardClient", "Samples\StandardClient\StandardClient.csproj", "{C43E715A-5E6D-40B1-AAA4-A91A1CBA3524}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CoreWCF.ConfigurationManager", "CoreWCF.ConfigurationManager\src\CoreWCF.ConfigurationManager.csproj", "{8BA271E4-0DCE-4D1B-8373-B07D35F8C09E}"
+EndProject
```
mconnew, line 44: Can you also add to CoreWCF.sln

src/CoreWCF.ConfigurationManager/src/ConfigurationManagerExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
mconnew, line 2: There shouldn't be any source files at the root of the project. Can you move the collection of source files to the correct folder?

src/CoreWCF.ConfigurationManager/src/WrappedConfigurationFile.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public WrappedConfigurationFile(string path)
+        {
+            _config = new Lazy<string>(() => WrapFile(XDocument.Load(path)), true);
+        }
+
+        private static string WrapFile(XDocument original)
+        {
+            string configPath = Path.GetTempFileName();
+            // for debug
```
mconnew: We can't save this to a temp file. Often the temporary file path is readable by any process on the machine. The passed in config could have security keys included inline. I see this eventually used with ConfigurationFileMap which is meant to represent machine.config. This isn't the optimal way to do this. On .NET Framework, you have machine.config, and you have app.config (or web.config). When you use ConfigurationManager to look for things, it will effectively merge the two in the in-memory model. The correct thing to do would be to have the machine.config pieces in one file, and leave the input config file as is unaltered and in it's original location. That way the security of the app config is the responsibility of the service owner.  
There are a couple of options. You could ship a static machine.config file as part of the package and reference it. Or you could add a static machine config file as an assembly file resource and access that. Or you could dynamically create one in a temporary folder. I'm not a fan of the temporary file creation as you aren't guaranteed it will be deleted either.

src/CoreWCF.ConfigurationManager/src/WrappedConfigurationFile.cs
```diff
@@ -0,0 +1,61 @@
...
+            using (var writer = XmlWriter.Create(configPath, new XmlWriterSettings { Indent = true }))
+            {
+                doc.WriteTo(writer);
+            }
+
+            return configPath;
+        }
+
+        internal void Dispose()
+        {
```
mconnew: If we stick with a temporary file, class should implement IDisposable. Also should have a finalizer to make sure the file gets deleted if Dispose isn't called.

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ICreateBinding.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using CoreWCF.Channels;
+
+public interface ICreateBinding
+{
```
mconnew, line 8: IBindingFactory would probably be a better name

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/NotFoundEndpointException.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace CoreWCF.Configuration
+{
+    public class NotFoundEndpointException : Exception
+    {
```
mconnew, line 9: We already have an EndpointNotFoundException, would it make sense to use that here? In CoreWCF that exception is thrown when a message arrives with a To header for an address that the service isn't configured to accept message for. On WCF client, it's thrown when the server indicates the endpoint doesn't exist, eg with a 404 HTTP response. So it isn't exactly the same scenario, but I don't see an ambiguity with how this would get thrown. I believe it would be different code paths.

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ServiceDefaults.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace CoreWCF.Configuration
+{
+    static class ServiceDefaults
+    {
```
mconnew, line 9: There are several places where we have missing accessors (internal in this case). If you place the cursor in the word static, press ctrl + ., choose "Add acessibility modifiers" and select to fix it in the entire project, you should be able to fix them all in one go.

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ConfigurationManagerServiceModelOptions.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public ConfigurationManagerServiceModelOptions(IServiceProvider builder, string path, bool isOptional)
+        {
+            _holder = builder.GetRequiredService<IConfigurationHolder>();
+            _file = new WrappedConfigurationFile(path);
+
+            _section = new Lazy<ServiceModelSectionGroup>(() =>
+            {
+                System.Configuration.Configuration configuration = ConfigurationManager.OpenMappedMachineConfiguration(new ConfigurationFileMap(_file.ConfigPath));
+                var section = ServiceModelSectionGroup.GetSectionGroup(configuration);
```
mconnew, line 28: I found the api's you need to use so you don't need to merge the app.config with the machine.config.
```c#
ExeConfigurationFileMap fileMap = new ExeConfigurationFileMap
{
    ExeConfigFilename = path,
    MachineConfigFilename = machineConfigFilePath
};
System.Configuration.Configuration configuration = ConfigurationManager.OpenMappedExeConfiguration(fileMap, ConfigurationUserLevel.None);
```
This way you don't need to manually merge them.

src/CoreWCF.Primitives/src/CoreWCF/Configuration/IConfigurationHolder.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using CoreWCF.Channels;
+
+namespace CoreWCF.Configuration
+{
+    public interface IConfigurationHolder
+    {
```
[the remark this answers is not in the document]
mconnew: I've been going back and forth on how to approach this. This configuration mechanism using IConfigureOptions kind of mirrors the UseServiceModel apis which takes a delegate which is passed a IServiceBuilder instance where you define the service. The difference is the convention with IConfigureOptions is to have the options type be a concrete class implementation. So I took a look at how the KestrelServerOptions is used. In `NetTcpFramingOptionsSetup.Configure` we configure an instance of KestrelServerOptions. We've previously populated an endpoint list on our own internal configuratopm class NetTcpFramingOptions via calls to UseNetTcp. Basically if you call UseNetTcp with two different ports, it will have to endpoints listed in NetTcpFramingOptions.Endpoints. We iterate over the endpoints and call KestrelServerOptions.Listen for each one in turn. The code looks like this:
```
            foreach (IPEndPoint endpoint in _options.EndPoints)
            {
                options.Listen(endpoint, builder =>
                {
                    builder.UseConnectionHandler<NetMessageFramingConnectionHandler>();
                    NetMessageFramingConnectionHandler handler = builder.ApplicationServices.GetRequiredService<NetMessageFramingConnectionHandler>();
                    // Save the ListenOptions to be able to get final port number for adding BaseAddresses later
                    ListenOptions.Add(builder);
                });
            }
```

We should be going for a similar model. The method ConfigurationManagerServiceModelOptions.Configure should have some way to parse the configuration into some kind of object model. The ConfigurationHolder class looks like a good candidate. This class should be internal though. It could either be done in Configure with a method call, or via something that is injected. Then the this object model is iterated over and ServiceModelOptions populated with a more typical set of objects which reflect closely to what IServiceBuilder does. At service startup, CoreWCF should get the ServiceModelOptions and iterate over it's data to set things up. So ConfigurationManagerServiceModelOptions.Configure would look something like this:
```
public void Configure(ServiceModelOptions options)
{
    var configHolder = ParseConfig();
    foreach (var endpointName in configHolder.Endpoints.Keys) // expose ConfigurationHolder._endpoints as property
    {
        var configEndpoint = configHolder.GetXmlConfigEndpoint(endpointName);
        options.ConfigureService(configEndpoint.Service, serviceConfig =>
        {
            serviceConfig.ConfigureServiceEndpoint(endpoint.Service, endpoint.Contract, endpoint.Binding, endpoint.Address, null);
        });
    }
}
```
Having ServiceModelOptions and another class ServiceConfigurationBuilder be public types which the ServiceBuilder can pull from DI and use to build the service allows a future capability to add other config options such as JSON. The classes would look like this and live in primitives.
```
public class ServiceModelOptions
{
    private Dictionary<Type, ServiceConfigurationBuilder> _config = new Dictionary<Type, ServiceConfigurationBuilder>();
    public void ConfigureService(Type serviceType, Action<ServiceConfigurationBuilder> configure)
    {
        ServiceConfigurationBuilder configBuilder;
        if (!_config.TryGetValue(serviceType, out configBuilder))
        {
            configBuilder = new ServiceConfigurationBuilder(serviceType);
            _config[serviceType] = configBuilder;
        }
        configBuilder.AddConfigureDelegate(configure);
    }
    internal void ConfigureServiceBuilder(IServiceBuilder serviceBuilder)
    {
        foreach (var serviceConfigBuilder in _config.Values)
        {
            serviceConfigBuilder.ConfigureServiceBuilder(serviceBuilder);
        }
    }
}

public class ServiceConfigurationBuilder
{
    private List<Action<ServiceConfigurationBuilder>> _configDelegates = new List<Action<ServiceConfigurationBuilder>>();
    private Dictionary<Type, List<(Type implementedContract, Binding binding, Uri address, Uri listenUri)>> _endpoints;
    internal Type ServiceType { get; }
    public ServiceConfigurationBuilder(Type serviceType)
    {
        ServiceType = serviceType;
    }
    public void AddConfigureDelegate(Action<ServiceConfigurationBuilder> configDelegate)
    {
        _configDelegates.Add(configDelegate);
    }
    public void ConfigureServiceEndpoint(Type service, Type implementedContract, Binding binding, Uri address, Uri listenUri)
    {
        List<(Type implementedContract, Binding binding, Uri address, Uri listenUri)> serviceEndpoints;
        if (!_endpoints.TryGetValue(service, out serviceEndpoints))
        {
            serviceEndpoints = new List<(Type implementedContract, Binding binding, Uri address, Uri listenUri)>();
            _endpoints[service] = serviceEndpoints;
        }
        serviceEndpoints.Add((implementedContract, binding, address, listenUri));
    }
    internal void ConfigureServiceBuilder(IServiceBuilder serviceBuilder)
    {
        _endpoints = new Dictionary<Type, List<(Type implementedContract, Binding binding, Uri address, Uri listenUri)>>();
        foreach (var configDelegate in _configDelegates)
        {
            configDelegate(this);
        }
        foreach (var service in _endpoints)
        {
            serviceBuilder.AddService(service.Key);
            serviceBuilder.AddServiceEndpoint(service.Key, service.Value.implementedContract, service.Value.binding, service.Value.address, service.Value.listenUri);
        }
        _endpoints = null;
    }
}
```
Then in ServiceBuilder where is creates the service, right at the top it grabs from DI ServiceModelOptions and calls `ServiceModelOptions.ConfigureServiceBuilder(this)`.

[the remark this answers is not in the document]
mconnew: Btw, I wrote this code in the GitHub comment box so there might be typos or some errors in there.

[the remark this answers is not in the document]
birojnayak: @mconnew @Ximik87 should we expose another method inside ServiceConfigurationBuilder as ConfigureBehavior(), which would be called from ConfigurationManagerServiceModelOptions.Configure (if behaviors present) and set up the ServiceBehaviors and EndPointBehaviors.. this would enable the TransportWithMessageCred etc... let me know what you both think...

birojnayak:
`  <system.serviceModel>
    <services>
      <service name="Service" behaviorConfiguration="ServiceBehavior">
        
        <endpoint address="" binding="basicHttpBinding" contract="IService">
          
          
        </endpoint>
        <endpoint address="mex" binding="mexHttpBinding" contract="MetadataExchange"/>
      </service>
    </services>
    <behaviors>
      <serviceBehaviors>
        <behavior name="ServiceBehavior">
          
          <serviceMetadata httpGetEnabled="true"/>
          
          <serviceDebug includeExceptionDetailInFaults="false"/>
        </behavior>
      </serviceBehaviors>
    </behaviors>
  </system.serviceModel>`
I was doing some testing with the changes by creating a default WCF service. The endpoint  doesn't have "identity" (please create an Issue, to separately handle) and importantly since there are 2 end points (with empty names), the ConfigurationManager is dying saying  the entry " has already been added... see if that can be fixed.

birojnayak:
@Ximik87 are you working on this PR ? Or if you want I can take it forward from here to bring closure and incorporate review comments... let me know..

Ximik87:
@birojnayak yes, i'm working, don't close this PR,  Matt gave a lot of comments about the code, and it will take some time

src/CoreWCF.ConfigurationManager/src/WrappedConfigurationFile.cs
Ximik87: Yes, I understand that a hack with a temporary file is a bad idea, but this is just to force the ConfigurationManager to read a config like <system.serviceModel>,
if there is a way to do it InMemory (without creating your own custom sections) - let me know, I saw you provided the code below, I have not tested it yet.

src/CoreWCF.Primitives/src/CoreWCF/Configuration/IConfigurationHolder.cs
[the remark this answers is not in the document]
Ximik87: yes, I have little experience with IConfigureOptions and I did not fully understand how it works correctly.
@mconnew , I did as you described, and it works except ServiceModelOptions.ConfigureServiceBuilder (this)
because - AddService in Configure (app, env) calls ServiceModelOptions.ConfigureServiceBuilder (this),
this calls ServiceConfigurationBuilder.onfigureServiceBuilder () and this calls AddService again and so recursively ...
I made an extension method for IServiceBuilder that explicitly calls ServiceModelOptions.ConfigureServiceBuilder and it seems to work, but check it out.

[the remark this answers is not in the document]
Ximik87: @birojnayak, Yes, as I understand it Matt - all ServiceBehaviors and EndPointBehaviors settings will be read in ConfigurationManagerServiceModelOptions.Configure
but we will use them in the ServiceConfigurationBuilder, make a separate method for this - I don't understand yet, I have to start trying coding

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/NotFoundEndpointException.cs
Ximik87: I think this is semantically correct, there is no reason to create a new type for exception

paulomorgado:
I'm a bit confused about the why and what for of this PR.

Is this to consume a web.config/app.config of the running process or just consume any file with a `system.serviceModel` section?

Why just not [`AddXmlFile`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.xmlconfigurationextensions.addxmlfile?view=dotnet-plat-ext-5.0) or [`AddXmlStream`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.xmlconfigurationextensions.addxmlstream?view=dotnet-plat-ext-5.0) the significant part?

paulomorgado:
Ideally, something like this would be enough:

```csharp
var configXml = @"<configuration>
  <system.serviceModel>
    <services>
      <service name=""Service"" behaviorConfiguration=""ServiceBehavior"">
        <!-- Service Endpoints -->
        <endpoint address="""" binding=""basicHttpBinding"" contract=""IService"">
          <!-- 
              Upon deployment, the following identity element should be removed or replaced to reflect the 
              identity under which the deployed service runs.  If removed, WCF will infer an appropriate identity 
              automatically.
          -->
          <!-- <identity>
            <dns value=""tlocalhost""/>
          </identity>-->
        </endpoint>
        <endpoint address=""mex"" binding=""mexHttpBinding"" contract=""MetadataExchange""/>
      </service>
    </services>
    <behaviors>
      <serviceBehaviors>
        <behavior name=""ServiceBehavior"">
          <!-- To avoid disclosing metadata information, set the value below to false before deployment -->
          <serviceMetadata httpGetEnabled=""true""/>
          <!-- To receive exception details in faults for debugging purposes, set the value below to true.  Set to false before deployment to avoid disclosing exception information -->
          <serviceDebug includeExceptionDetailInFaults=""false""/>
        </behavior>
      </serviceBehaviors>
    </behaviors>
  </system.serviceModel>
</configuration>";

var config = new ConfigurationBuilder()
    .AddXmlStream(new MemoryStream(Encoding.UTF8.GetBytes(configXml)))
    .Build();
```

But the first problem here is the `services/endpoint` element. It needs to be unique.

paulomorgado:
This is a naïf and not performant way of sourcing the configuration from `/configuration/system.serviceModel´:

```csharp
var configItems = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);

var xml = new XmlDocument();
xml.LoadXml(configXml);

var serviceModel = xml.SelectSingleNode("//system.serviceModel");

foreach (XmlNode service in serviceModel.SelectNodes("services/service"))
{
    var serviceConfigurationRoot = $"serviceModel{ConfigurationPath.KeyDelimiter}services{ConfigurationPath.KeyDelimiter}{service.Attributes["name"]!.Value}{ConfigurationPath.KeyDelimiter}";

    foreach (XmlAttribute attribute in service.Attributes)
    {
        if (attribute.Name != "name")
        {
            configItems.Add($"{serviceConfigurationRoot}{attribute.Name}", attribute.Value);
        }
    }

    foreach (XmlNode endpoint in service.SelectNodes("endpoint"))
    {
        var endpointConfigurationRoot = $"{serviceConfigurationRoot}endpoints{ConfigurationPath.KeyDelimiter}{endpoint.Attributes["address"]!.Value}{ConfigurationPath.KeyDelimiter}";

        foreach (XmlAttribute attribute in endpoint.Attributes)
        {
            if (attribute.Name != "address")
            {
                configItems.Add($"{endpointConfigurationRoot}{attribute.Name}", attribute.Value);
            }
        }
    }
}

var config = new ConfigurationBuilder()
    .AddInMemoryCollection(configItems)
    .Build();
```

The result will be:

|Key|Value|
|-|-|
|serviceModel|_null_|
|serviceModel:services|_null_|
|serviceModel:services:Service|_null_|
|serviceModel:services:Service:endpoints|_null_|
|serviceModel:services:Service:endpoints:mex|_null_|
|serviceModel:services:Service:endpoints:mex:contract|MetadataExchange|
|serviceModel:services:Service:endpoints:mex:binding|mexHttpBinding|
|serviceModel:services:Service:endpoints:|_null_|
|serviceModel:services:Service:endpoints::contract|IService|
|serviceModel:services:Service:endpoints::binding|basicHttpBinding|
|serviceModel:services:Service:behaviorConfiguration|ServiceBehavior|

Loading this JSON configuration would yield the same result:

```JSON
{
  "ServiceModel" : {
    "Services" : {
      "Service": {
        "Endpoints": {
          "mex": {
            "Contract": "MetadataExchange",
            "Binding": "mexHttpBinding",
          }
          "": {
            "Contract": "IService",
            "Binding": "ServiceBehavior",
          }
        }
      }
    }
  }
}
```

Now, what's missing is an options model.

Ximik87:
@paulomorgado, firstly, this is not a refinement of the ConfigurationBuilder as you described, secondly, it is not as simple as you gave in the example implementation, because the serviceModel
a lot of parameters to configure via app.config - bindings, timeouts, transportMode, securityMode and so on ...

paulomorgado:
It's not a refinement of `ConfigurationBuilder` when it's using the configuration system as intended. So, I didn't describe that.

I did say that the implementation was "naïf and not performant".

paulomorgado:
Here is the logging set up from configuration: https://source.dot.net/#Microsoft.Extensions.Logging.Configuration/ILoggerProviderConfigurationFactory.cs

And where it is used in the host builder: https://source.dot.net/#Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs,238

birojnayak:
@Ximik87 I have started porting Behavior elements and eventually add in the ServiceConfigurationBuilder by defining method .. let me know if you are doing that... @mconnew could you please review this PR ?

src/CoreWCF.ConfigurationManager/src/WrappedConfigurationFile.cs
mconnew: This turned out to be more complicated than I expected, there were multiple hurdles to overcome to get it to work. If you look at my [commit here](https://github.com/mconnew/CoreWCF/commit/080a51a33d0ebe83ee3b3faca75277bbf6481c3d) I have added a machine.config file to the nuget package and added a targets file to copy it into the application output folder. Then in the code I specify that as the machine.config file.

Ximik87: @mconnew Great! thank you so much, I applied these changes in PR, it works, though it took a small fix for .net core 2.1 on linux

Ximik87:
@birojnayak yes, this PR lacks support for behavior, because I don't understand much how to do it in the code. If you start doing it well, I think you can handle it better

birojnayak:
> @birojnayak yes, this PR lacks support for behavior, because I don't understand much how to do it in the code. If you start doing it well, I think you can handle it better

@Ximik87 I will create an issue once this is merged for behavior support

src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
```diff
@@ -0,0 +1,34 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\CoreWCF.Primitives\src\CoreWCF.Primitives.csproj" />
+    <ProjectReference Include="..\src\CoreWCF.ConfigurationManager.csproj">
+      <Private>true</Private>
```
mconnew, line 30: Why did you need to put `<Private>true</Private>` on this project reference?

Ximik87: this code generated of visual studio, i'm fixed it

mconnew:
I have a few more tweaks to clean up the code and api design a little and then this will be good. I'll try and push that in the next hour or so.

mconnew:
@Ximik87, I have another commit for you to add to your branch. You can find it [here](https://github.com/mconnew/CoreWCF/commit/f70bd670ce9f879f67adefeb29d17b2064f88b86).  

Basically I made it so you add using the config file in the ConfigureServices method and simply use app.UseServiceModel() in your Configure method. I removed any concept of Xml based configuration from Primitives. I made concrete implementations of interfaces internal. I also modified a few method names to make their intent a little clearer.  

Feel free to comment on any of my changes if you think it should be a different way.

mconnew:
@Ximik87, this needs to be rebased on top of main as it looks like there's some conflict now.

Ximik87:
@mconnew  Of course, here's an example blog post

Added support for configuration loaded from the classic app.config. Of course, not all options are supported, but this should help when migrating WCF applications.
An example migration guide looks like this:
Copy the system.serviceModel section into a separate file, for example wcf.config, which should be located in the application folder. Example of file content:
```
 <system.serviceModel>
    <bindings>
      <netTcpBinding>
        <binding name="netTcpBindingConfig" receiveTimeout="00:10:00" />       
      </netTcpBinding>
    </bindings>
    <services>
      <service name="Services.ISomeContact" >
        <endpoint address="net.tcp://localhost:8750/Service" binding="netTcpBinding" bindingConfiguration="netTcpBindingConfig" contract="ISomeContact"  />       
      </service>
    </services>
  </system.serviceModel>
```
Next, we create an ASP .Net Core application, add the CoreWCF.Primitives, CoreWCF.NetTcp and CoreWCF.ConfigurationManager nuget packages. Configure services and applications:
```
public void ConfigureServices(IServiceCollection services)
{           
     services.AddServiceModelConfigurationManagerFile("wcf.config");
}

public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
      app.UseServiceModel();
}
```
Note that if NetTcpBinding is used, then the ports of endpoints using this binding must be additionally specified when configuring WebHostBuilder:
```
  	  .UseNetTcp(8750)
          .UseStartup<Startup>();
```
Ready! If there are any configuration problems in the wcf.config file, this will be visible when the application starts.

List of supported attributes:
BindingSection:  
BasicHttpBinding, NetHttpBinding, NetTcpBinding, WSHttpBinding
For all specified bindings - SecurityMode, MaxReceivedMessageSize, ReceiveTimeout, CloseTimeout, OpenTimeout, SendTimeout and XmlReaderQuotas. As well as specific settings for specific bindings:
BasicHttpBinding - MaxBufferSize,TransferMode,TextEncoding
NetHttpBinding - MaxBufferSize,TransferMode,TextEncoding,MessageEncoding
NetTcpBinding - MaxBufferSize,MaxBufferPoolSize,MaxConnections, TransferMode,HostNameComparisonMode
WSHttpBinding  - MaxBufferPoolSize
ServiceSection:
Endpoint - Address,Binding,BindingConfiguration,Contract

PR #368, opened 2021-06-04
Samples include UserName and Password validation, and Fix in CoreWCF to allow it to work 

JohnLeyva:
First commit has a fix to allow UserNameCredentialsValidation to work.

Samples UserName and Password validation

Common URLs in a shared Settings project that facilitates keeping the paths used in the sample projects


src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
```diff
@@ -488,7 +488,7 @@ public Task OpenAsync()
 
         public Task OpenAsync(CancellationToken token)
         {
-            throw new NotImplementedException();
+            return Task.CompletedTask;
```
birojnayak, line 491: looks good.. bdw this never gets called, did you hit a situation where this was getting called ?

birojnayak (commented):
looks ok to me... @mconnew any concern ?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
JohnLeyva: Yes, Indeed it was called consistenly from the sample application and a few times tests also called it, but the behaviour was not deterministic.

Wondering if it is related to having the debugger attached.

After, trying to follow it, in the end, I decided to implement the method. But still curios about what is really happening there.

CustomUserNamePasswordValidator.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.ServiceModel;
+
+namespace DesktopServer
+{
+    internal class CustomUserNamePasswordValidator : System.IdentityModel.Selectors.UserNamePasswordValidator
```
mconnew, line 5: This file and the Core file need to be in the relevant project folders and not at the root.

src/Samples/DesktopServer/Program.cs
```diff
@@ -1,25 +1,42 @@
...
-            host.AddServiceEndpoint(contract, new WSHttpBinding(SecurityMode.None), "/wsHttp.svc");
-            host.AddServiceEndpoint(contract, new WSHttpBinding(SecurityMode.Transport), "/wsHttp.svc");
+            string hostname = "localhost"; //"localtest"
+
+            IEnumerable<string> baseAddressList = new[] {
+                "net.tcp://localhost:8089/",
+                "http://localhost:8088/",
+                "https://localhost:8443/" };
+            baseAddressList = baseAddressList.Select(i => i.Replace("//localhost:", $"//{hostname}:"));
+            Uri[] baseUriList = baseAddressList.Select(a => new Uri(a)).ToArray();
```
mconnew, line 22: This replacing of the hostname isn't needed as you added an optional parameter to SetDefaults to do this for you.

src/Samples/StandardCommon/Settings.cs
```diff
@@ -0,0 +1,30 @@
...
+        public bool UseHttps { get; set; } = true;
+        public Uri basicHttpAddress { get; set; }
+        public Uri basicHttpsAddress { get; set; }
+        public Uri wsHttpAddress { get; set; }
+        public Uri wsHttpAddressValidateUserPassword { get; set; }
+        public Uri wsHttpsAddress { get; set; }
+        public Uri netTcpAddress { get; set; }
+
+        public Settings SetDetaults(string hostname = "localhost")
+        {
```
mconnew, line 16: SetDefaults

src/Samples/gencert.ps1
```diff
@@ -40,8 +40,9 @@ function Set-SSL_WCF_ServiceHost {
     $certHash = "certhash=$($certificateObject.Thumbprint)"
     $appid = "appid={1d7985f5-e1b0-4fd4-bdeb-7e28590783da}"
 
-#    netsh http add sslcert ipport=127.0.0.1:8443 $certHash $appid certstorename=Root
+    netsh http add sslcert ipport=127.0.0.1:8443 $certHash $appid certstorename=Root
     netsh http add sslcert hostnameport=localhost:8443 $certHash $appid certstorename=Root
+    netsh http add sslcert hostnameport=localtest:8443 $certHash $appid certstorename=Root
 #    Netsh http delete sslcert ipport=127.0.0.1:8443
```
mconnew, line 46: Do we need localtest in here? Looks like it's specific to your dev environment. There's also a little bit of remnants of this still in some of the test code.

JohnLeyva: I'll remove it. 
I added it in order to be able to use fiddler.

JohnLeyva:
@mconnew, @birojnayak, please let me know if there is something pending

mconnew:
@JohnLeyva, thanks for making the requested changes. Merging now.

PR #383, opened 2021-07-27
Adding IPAddress overload for net.tcp

mconnew:
Replaces #357. This is the same changes cherry-picked and manually resolved.


ovebastiansen:
missed the closes #342 from the org pr

PR #384, opened 2021-07-27
Changes to ensure the NetCoreServer compiles

birojnayak:


mconnew:
There's a test which is failing because the wrong inner exception type was returned on Linux. The test doesn't say what the wrong type is, so I've created a PR to improve the test so that if the same thing happens again, we'll have a better chance of tracking it down. That's in PR #385

mconnew:
Reran and the test passed, so it must be some timing issue causing a different exception to be thrown.

PR #394, opened 2021-08-01
Fix return output params

AlexanderSemenyak:
Always null output params resolving


mconnew:
This looks correct to me, but I'm going to manually verify this change against the disabled tests in PR #171 before I merge.

mconnew:
I've verified it fixes those broken tests.

Issue #318, opened 2021-02-26
NetTcp and Duplex Channel


MemoTrap:
CoreWCF 0.1.0

I tried a little example with NetTcp binding and a callback channel, based on:
https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/duplex-services.

It crashes in the CoreWCF based server, throwing InvalidCast, but works fine in classic .Net Framework.
```
    ICalculatorDuplexCallback Callback {
      get {
        return OperationContext.Current.GetCallbackChannel<ICalculatorDuplexCallback> ();
      }
    }
```
`System.InvalidCastException: 'Unable to cast object of type 'CoreWCF.Channels.ServiceChannel' to type 'CalculatorCommon.ICalculatorDuplexCallback'.'`

Same effect with both clients, Core or Framework.

Is there anything I have to do differently for CoreWCF?

The entire sample is here:
[CalculatorDuplex.zip](https://github.com/CoreWCF/CoreWCF/files/6048651/CalculatorDuplex.zip)

_On a sideline:_
All operation contracts are one way, as in the original example. The client first calls the "add" method, then "subtract". 
```
      var factory = new
                    DuplexChannelFactory<ICalculatorDuplex> (context, binding, address);
      ICalculatorDuplex channel = factory.CreateChannel ();

      channel.AddTo (1234);
      channel.SubtractFrom (123);
```
However, on the CoreWCF server, "subtract" is invoked first.

mconnew:
This has been merged and will be released as part of 0.3.0

jaubin333:
> I would be happy to contribute a duplex channel example once working.

Were you ever able to get a working duplex example? I'd love to grab it and take a look.

jaubin333:
> Never mind. The problem was invoking the GetCurrentCallbackChannel with the service interface rather than the callback interface.

@CloudwatcherGit  Can you show what, exactly, you fixed? I am new to this and looking to use this CalculatorDuplex project as a sample and I am seeing the same invalid cast exception but unsure where you made the tweak to fix it. Thanks!

Issue #264, opened 2020-11-19
Add StrongName 


Danielku15:
Many applications and other libraries in the market still use Strong Names, so do we for our products (historically). All CoreWCF packages currently are not signed with any solution key and therefore do not have a strong name. 

In order to support smooth migration for any products still requiring strong names, it would be great if also CoreWCF would add it. 

Please note I am requesting this purely for compatibility with products/libraries which cannot simply remove the SNK due to again transitive dependencies expecting it. So basically the first point of ["Why strong-name your assemblies"](https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named) 

> You want to enable your assemblies to be referenced by strong-named assemblies, or you want to give friend access to your assemblies from other strong-named assemblies.

mconnew:
That constraint is only enforced when running on .NET Framework. In .NET Core/.NET 5, an assembly can be strong names and reference a non-strong named assembly. With this in mind, is this still something you need?

Danielku15:
Yes. Unfortunately we are forced to stick still to .net 4.8. We have a rather long product roadmap to finally hop to .net core / .net 5 where we need to adapt a lot of changes (ASP.net -> ASP.net core, EF -> EF core, WCF -> CoreWCF/Rest/gRPC). This migration needs to be done for multiple services, libraries and products. 

Having a strong name on CoreWCF would allow as a more step-wise transition/migration of our products.

mconnew:
Do you need a full strong-name signed assembly or is public signed an option? Public signed assemblies basically have the public key in the assembly to use as part of their identity but the assembly is missing the actual signature. So the assembly load binding process can validate the identity of the assembly, but there's no actual signature. It makes creating a private build and dropping it in place to test something locally easier and it's the method used by .NET Core OOB packages. There are 3 scenarios it doesn't work for, you can read more info [here](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/public-signing.md#:~:text=NET%20Core%20projects%20build%20using%20public%20signing.%20Public,context%20that%20doesn%27t%20require%20validating%20the%20strong-name%20signature.).

Daniel-Svensson:
Just a quick question: are there any drawbacks with using normal strong singing and publishing the snk in the repo so anyone can use it?

I think that was the recommendation a while back and it should have the mentioned advantages of public signed.

mconnew:
[...]

PR #399, opened 2021-08-04
Minor patch to make happy path of duplex channel working

Danielku15:
Relates to #318, #398
This PR adds a simple test for a duplex service where the server is broadcasting messages to clients. 

To get it actually green and working there was only a very small change needed. During some proxy class generation an exception was raised because the base class for the proxy was sealed and internal. By making it public the proxy generation works and the calls start to succeed. 

Please note that this is just a very basic minor patch to make a happy path working. I am not aware yet of certain paths which might not work.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.ComponentModel;
 using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Reflection;
@@ -15,6 +16,9 @@ using CoreWCF.Runtime;
 
 namespace CoreWCF.Channels
 {
+    [Browsable(false)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [Obsolete("This API supports the .NET Framework infrastructure and is not intended to be used directly from your code.")]
     public class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
     {
         private const string activityIdSlotName = "E2ETrace.ActivityID";
@@ -45,7 +49,7 @@ namespace CoreWCF.Channels
         //Workaround is to set the activityid in remoting call's LogicalCallContext
 
         // Override ToString() to reveal only the expected proxy type, not the generated one
-        public override string ToString()
+        public sealed override string ToString()
         {
             return _proxiedType.ToString();
         }
@@ -115,7 +119,7 @@ namespace CoreWCF.Channels
             return _serviceChannel;
         }
 
-        protected override object Invoke(MethodInfo targetMethod, object[] args)
+        protected sealed override object Invoke(MethodInfo targetMethod, object[] args)
         {
             if (args == null)
             {
@@ -659,14 +663,14 @@ namespace CoreWCF.Channels
             return _serviceChannel.RequestAsync(message, token);
         }
 
-        public Task CloseOutputSessionAsync(CancellationToken token)
+        Task IDuplexContextChannel.CloseOutputSessionAsync(CancellationToken token)
         {
             return ((IDuplexContextChannel)_serviceChannel).CloseOutputSessionAsync(token);
         }
 
         EndpointAddress IRequestChannel.RemoteAddress
         {
-            get { return ((IContextChannel)_serviceChannel).RemoteAddress; }
+            get { return _serviceChannel.RemoteAddress; }
         }
 
         Uri IRequestChannel.Via
@@ -679,7 +683,7 @@ namespace CoreWCF.Channels
             get { return _serviceChannel.ListenUri; }
         }
 
-        public bool AutomaticInputSessionShutdown
+        bool IDuplexContextChannel.AutomaticInputSessionShutdown
         {
             get
             {
@@ -692,7 +696,7 @@ namespace CoreWCF.Channels
             }
         }
 
-        public InstanceContext CallbackInstance
+        InstanceContext IDuplexContextChannel.CallbackInstance
         {
             get
             {
@@ -705,7 +709,7 @@ namespace CoreWCF.Channels
             }
         }
 
-        public IServiceChannelDispatcher ChannelDispatcher { get => ((IChannel)_serviceChannel).ChannelDispatcher; set => ((IChannel)_serviceChannel).ChannelDispatcher = value; }
+        IServiceChannelDispatcher IChannel.ChannelDispatcher { get => ((IChannel)_serviceChannel).ChannelDispatcher; set => ((IChannel)_serviceChannel).ChannelDispatcher = value; }
         #endregion // Channel interfaces
     }
 }
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
@@ -15,7 +15,7 @@ using CoreWCF.Runtime;
 
 namespace CoreWCF.Channels
 {
-    internal sealed class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
+    public class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
     {
         private const string activityIdSlotName = "E2ETrace.ActivityID";
         private Type _proxiedType;
--- a/src/CoreWCF.NetTcp/tests/CoreWCF.NetTcp.Tests.csproj
+++ b/src/CoreWCF.NetTcp/tests/CoreWCF.NetTcp.Tests.csproj
@@ -19,6 +19,7 @@
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
     <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
+    <PackageReference Include="System.ServiceModel.Duplex" Version="4.8.1" />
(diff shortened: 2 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/CoreWCF/CoreWCF?pullRequest=399) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/CoreWCF/CoreWCF?pullRequest=399) before we can accept your contribution.<br/><br/>:x: Danielku15 [sign now](https://cla.dotnetfoundation.org/CoreWCF/CoreWCF?pullRequest=399)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/CoreWCF/CoreWCF?pullRequest=399) it.</sub>

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelHandler.cs
```diff
@@ -615,6 +615,7 @@ private Task InitializeServiceChannel(ServiceChannel channel)
                 _serviceDispatcher.ChannelDispatcher.InitializeChannel((IClientChannel)channel.Proxy);
             }
 
+            // TODO: shouldn't this open be awaited? -> problem with lock outside
```
Danielku15, line 618: I stumbled over this potentially bug/issue. I came to notice that calls to InitializeServiceChannel are NOT awaited:

https://github.com/CoreWCF/CoreWCF/blob/9bdd86153f04790cea42147da1e47ba3ede2f2a7/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelHandler.cs#L534
https://github.com/CoreWCF/CoreWCF/blob/9bdd86153f04790cea42147da1e47ba3ede2f2a7/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelHandler.cs#L557

Awaits are not easily possible there because awaits are not supported within locks. And there are also out parameters which cannot be used with async-await. 

I would guess the channel creation could be reworked to use some SemaphoreSlim instead of a lock and the out parameters could be replaced with a named-tuple return. I didn't dare to touch this area because it seems very general overall CoreWCF and not specific to handling Duplex Channels, but I wanted to raise this concern.

birojnayak:
> Relates to #318, #398
> This PR adds a simple test for a duplex service where the server is broadcasting messages to clients.
> 
> To get it actually green and working there was only a very small change needed. During some proxy class generation an exception was raised because the base class for the proxy was sealed and internal. By making it public the proxy generation works and the calls start to succeed.
> 
> Please note that this is just a very basic minor patch to make a happy path working. I am not aware yet of certain paths which might not work.



> Relates to #318, #398
> This PR adds a simple test for a duplex service where the server is broadcasting messages to clients.
> 
> To get it actually green and working there was only a very small change needed. During some proxy class generation an exception was raised because the base class for the proxy was sealed and internal. By making it public the proxy generation works and the calls start to succeed.
> 
> Please note that this is just a very basic minor patch to make a happy path working. I am not aware yet of certain paths which might not work.

We were going to look into this issue, thanks for taking the initiative. We will get it reviewed soon.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelHandler.cs
mconnew: You're right, this is a potential race condition bug. I had a look and the code in GetSessionChannel is safe to replace with a SemaphoreSlim. This is the session creation code path, and the only other place ThisLock is used is in the session close code path. There isn't going to be any interaction between those two code paths, but even it's safe to replace the ThisLock with a SemaphoreSlim and use it in both locations.
The same is true of the _serviceDispatcher.ThisLock, it's only used in this one location so you can replace it with a SemaphoreSlim safely.

The problem you will hit is those methods use out parameters, so you will need to modify the method to use a Task of named tuple return value to be able to return multiple values. E.g.:
```c#
private async Task<(ServiceChannel serviceChannel, EndpointDispatcher endpoint, bool addressMatched)>  GetDatagramChannel(Message message)
```

Would you be able to fix this and make the async flow through correctly?

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace CoreWCF.Channels
 {
-    internal sealed class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
+    public class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
```
mconnew, line 18: This change makes sense, but I'm not really happy with it. The DispatchProxy class that we use to create the proxied type needs to derive from this class via ref emit, which if why it can't be sealed and needs to be public. The thing I don't like is making this type public. Are you familiar with the `InternalsVisibleTo` attribute? [Docs are here](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-5.0). Could you see if adding an InternalsVisibleTo attribute for the DispatchProxy assembly would allow us to keep this type internal?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ChannelHandler.cs
Danielku15: Sure, I'll prepare the change later today,l and add it to this PR, should not take very long.

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
Danielku15: I will double check this change and try to keep the class internal. Actually I just went the quick path to align it with the [System.ServiceModel.ServiceChannelProxy](https://github.com/dotnet/wcf/blob/367d64dc79a9b23acc15a3ef549996cf05e10e72/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannelProxy.cs#L18), `sealed` is definitely an issue, but maybe `internal` can even stay.

Danielku15: @mconnew I fear that `public` cannot be avoided for the currently targeted runtimes unless we remove the strong name. The issue is partly related to https://github.com/dotnet/runtime/issues/30917  and some limitations with the `InternalsVisibleTo`:

I tried following things: 

- Add `[assembly: InternalsVisibleTo("ProxyBuilder")]` -> Leads to a compiler error because CoreWCF has a strong name: 

> Error CS1726 : Friend assembly reference 'ProxyBuilder' is invalid. Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations.

- Add `[assembly: InternalsVisibleTo("ProxyBuilder, PublicKey=null")]` or `[assembly: InternalsVisibleTo("ProxyBuilder, PublicKey=")]` or `[assembly: InternalsVisibleTo("ProxyBuilder, PublicKey=<public-key of corewcf.snk>")]`
 -> Leads to a runtime error (not always but sometimes, likely related to https://github.com/dotnet/runtime/issues/30917): 
> Type 'generatedProxy_6' from assembly 'ProxyBuilder, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' is attempting to implement an inaccessible interface.

I was the initiator for requesting strong name via #264 Things changed already for us, and nowadays we would also be fine to not have a strong name anymore. 

I think we have only two choices: 

a) Make it public
b) Remove the strong name and add a `[assembly: InternalsVisibleTo("ProxyBuilder")]`

Danielku15:
@mconnew I stayed for now with the approach to keep it public to keep the changes in this PR small. Let me know your thoughts on your preferred approach. I can also adapt to the approach to remove the strong name again. 

I found the `AsyncLock` which already abstracts away the complexity of locking via SemaphoreSlim. So I used it at the affected places and changed the out params to named-tuple returns. It would likely be a bigger task for an own PR to replace all classical locks with this one to stay unified and async-compatible. I want to keep this change small with a clear focus 😉

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannelProxy.cs
Danielku15: I fiddled a bit around with some tricks to maybe bypass some checks of the runtime but unfortunately I failed. I was not able to work around the requirement that the base class of proxies currently need to be public. 😢 

I added the Browsable, EditorBrowsable and Obsolete attribute to hide it as good as possible. Unfortunately also here we cannot activate the error flag because there is no way to supress an error level Obsolete attribute. 

I also tried to minimize the public facing API surface to zero by sealing some methods and using explicit interface implementations where possible.

mconnew:
I just saw your comment about the AsyncLock. Don't use it, it's broken and leaks memory over time. Can you use a SemaphoreSlim like was originally discussed?

Danielku15:
@mconnew Yes, will rework to use directly SemaphoreSlim, but the memory leak is actually an important concern. I saw neither ServiceDispatcher nor ChannelHandler have an obvious resource cleanup in place so I am not sure exactly where to place the disposal. 

I think I will revert the lock/async changes within this PR and start a fresh PR for this concern. This way we can get the concern about the duplex channel closed and focus on the lock rework separately. What do you think?

mconnew:
You don't need to dispose the semaphore slim. Disposing it just releases any waiters, but as it's going to be used like a lock that isn't a problem. Just just need to make sure the code inside the "lock" is in a try/finally block with the release in the finally so that you guarantee any exceptions will still result in releasing the "lock". SemaphoreSlim doesn't leak, my AsyncLock implementation leaks. But having that as a separate PR makes sense anyway.

Danielku15:
@mconnew Thanks for the hint. Your remark made me double check again the implementation of SemaphoreSlim in detail. I was worried because it holds a ManualResetEvent as handle and in the dispose method is cleans it. From this I assumed it would cause a handle leak if we don't dispose it. But you're right: upon SemaphoreSlim creation there is no wait handle created. It is only leaking if you use AvailableWaitHandle from within your code. But also SafeHandle should be finalizable so GC should take care of it in worst case.

PR #400, opened 2021-08-06
Fix HTTPS endpoints

mconnew:
No longer need dummy http endpoints
Now allows multiple https server addresses


mconnew:
This will be released as a servicing fix for 0.2

birojnayak (approved):

PR #404, opened 2021-08-08
Add net5.0 unit test target

g7ed6e:
As CoreWCF 0.2.0 is released and supports .net5 unit tests should also be ran targeting .net5


g7ed6e:
I think in unit test csproj at least we should, starting from netcoreapp3.1 target runtime, introduce FrameworkReference to Microsoft.AspNetCore.App. I will try to add these references.

PR #405, opened 2021-08-09
Fix BasicHttpBinding Transport security properties

mconnew:
When using security mode Transport, MaxReceivedMessageSize and MaxBufferSize weren't propagated to the transport binding element.


mconnew:
Merging without review as trivial change and added unit tests for the change.

PR #412, opened 2021-08-12
Use compiled lambda to access aspnetcore 5.0 remote endpoint

### LINKED ISSUE #409: RemoteEndpointMessageProperty broken for NetTcp @ net5.0
It seems there were some changes in the ASP.net core connection handling and Kestrel as part of net5.0 which break the access to the remote address of requests. 

**Situation Before**
In v2.x the `Microsoft.AspNetCore.Connections.ConnectionContext` did not expose any local or remote endpoints. The feature collection contained always an `IHttpConnectionFeature` which could be used for accessing the endpoint. The CoreWCF FramingConnection goes with this assumption an decodes the endpoint accordingly: 
https://github.com/CoreWCF/CoreWCF/blob/e16e5e5e8763c9d04f9de943b12a4804a27c0d86/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingConnection.cs#L70-L80

**Situation in net5.0**
In net5.0 the ConnectionContext does **not** have an `IHttpConnectionFeature` anymore. Instead the `ConnectionContext` inherits from `BaseConnectionContext` and this base class exposes the endpoints accordingly: 

The property is defined here: 
https://github.com/dotnet/aspnetcore/blob/c663adee8e64ba5d379fa0edfb8201984a7df7d0/src/Servers/Connections.Abstractions/src/BaseConnectionContext.cs#L43
And filled here by Kestrel (`SocketConnection` extends `TransportConnection` extends `ConnectionContext`): 
https://github.com/dotnet/aspnetcore/blob/952ccb4532cc6287ca49b91992b2912144d4d752/src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs#L57

This is currently blocking us of using CoreWCF because we are transitioning to net5.0.

Danielku15:
Fixes #409 

As requested [here ](https://github.com/CoreWCF/CoreWCF/issues/409#issuecomment-897792477) I added a reflection access to the new ASP.net core 5.x RemoteEndPoint property on the ConnectionContext. 

I did a bit an optimization already: 

* I use `Lazy<>` to avoid access to the data unless really needed. 
* I use a compiled lambda instead of always using reflection. 

To actually test it, I needed to adapt a bit the targeting of the test projects because of the version differences. ASP.net core became a framework reference. 

It partly duplicates what was done in #404 but in extension to it, it really pulls ASP.net core 5.0 instead of running ASP.net core 2.x under .net 5

### DIFF
--- a/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
+++ b/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
@@ -28,13 +28,15 @@
     <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' != 'net5.0'">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'netcoreapp2.1' ">
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing">
-      <Version Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472'">2.2.0</Version>
-      <Version Condition="!('$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472')">3.1.17</Version>
-    </PackageReference>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys" Version="2.1.12" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' ">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.1.17" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0'">
--- a/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
+++ b/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
@@ -23,16 +23,25 @@
     </PackageReference>
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
+    <PackageReference Include="System.ServiceModel.Http" Version="4.8.1" />
+    <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' != 'net5.0'">
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing">
       <Version Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472'">2.2.0</Version>
       <Version Condition="!('$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472')">3.1.17</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys" Version="2.1.12" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" />
-    <PackageReference Include="System.ServiceModel.Http" Version="4.8.1" />
-    <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
   </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0'">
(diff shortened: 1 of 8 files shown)


src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -23,16 +23,25 @@
...
       <Version Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472'">2.2.0</Version>
       <Version Condition="!('$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'net472')">3.1.17</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys" Version="2.1.12" />
-    <PackageReference Include="System.Net.Http" Version="4.3.4" />
-    <PackageReference Include="System.ServiceModel.Http" Version="4.8.1" />
-    <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
   </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0'">
```
mconnew, line 40: Could you do the same for ASP.NET Core 3.1 so that tests are actually running against the correct implementation there too?

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingConnection.cs
```diff
@@ -22,6 +23,7 @@ public class FramingConnection
     {
         private readonly ConnectionContext _context;
         private IDuplexPipe _transport;
+        private Lazy<IPEndPoint> _remoteEndpoint;
```
mconnew, line 26: Use of Lazy isn't needed as we always generate a RemoteEndpointMessageProperty for every request on NetTcp so it's inevitable this will always be used. The only benefit of a Lazy is to make the cost happen at first use rather than connection. Although this has given me an idea for a new feature. What do you think about adding a feature where if you add a `Lazy<T>` message property, then getting the property causes MessageProperties to evaluate the Lazy and return the value. This way things like RemoteEndpointMessageProperty would only get constructed if application code used it.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingConnection.cs
```diff
@@ -210,6 +200,57 @@ public async Task SendFaultAsync(string faultString, TimeSpan sendTimeout, int m
...
+            // https://github.com/dotnet/aspnetcore/blob/v5.0.9/src/Servers/Connections.Abstractions/src/BaseConnectionContext.cs
+            var property =
+                typeof(ConnectionContext).GetProperty("RemoteEndPoint", BindingFlags.Instance | BindingFlags.Public);
+            if (property == null)
+            {
+                return null;
+            }
+
+            // context => context.RemoteEndPoint as IPEndpoint
+            var contextParam = Expression.Parameter(typeof(ConnectionContext), "context");
```
mconnew, line 247: FYI, property.GetGetMethod returns a MethodInfo and you can use Delegate.CreateDelegate to create an open delegate for getting the property value. I'm fine with your approach too though. An interesting thing is your general approach used for fields is significantly faster than using reflection as the methods to access fields result in around 70 lines of native code running in the clr when using reflection, but expressions compile to IL and very directly access the field in memory. Expressions are also slower in AOT scenarios (at least they used to be for UWP) as they are interpreted at runtime and just end up using the reflection method in the end. I don't think that's a concern here though.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingConnection.cs
Danielku15: The lazy evaluation of properties might be beneficial if there can go something wrong during creation. The benefit I see there, is that if this really can happen the exception would be properly handed over to the caller instead of happening somewhere within the CoreWCF infrastructure during context creation. 

I remember the case in some old ASP.net Web API times where access to some network level properties could lead to an exception if the client closed the connection already. 

Unfortunately I fear that I'm lacking of the insight into the possible MessageProperties within WCF to judge whether such a lazy evaluation would be worth the effort. Might be a micro optimization without much benefit. For real critical paths, I would assume the lazy evaluation happens within the C# properties of the related MessageProperty class.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingConnection.cs
Danielku15: I used quite some expression compiling in the past due to this benefit of runtime performance. Of course it comes at some initial cost for compiling the lamdba. I think this dynamically emitted lambdas even have a chance to get inlined by the runtime which makes them almost as fast as if there would be a normal c# access. 

I would say also for AOT cases where the expression tree is still interpreted instead of emitted to IL, the complexity of the tree is simple enough to not cause real troubles.

mconnew:
It's failing with the error:
```
/templates/TestStage.yml (Line: 86, Col: 9): 'Windows_net5.0' is already defined
```
Which doesn't make any sense as everything else is defined twice, once under debug and again under release.

mconnew:
@Danielku15, I think I know what the problem is. The pipeline change has already been committed and this change is being merged with main and I think that's resulting in Windows_net5.0 being in the pipeline twice. Can you rebase to main and update the PR, that should fix it.

mconnew:
Now that you've rebased, you do have Windows_net5.0 in there twice, which is exactly what the build system was seeing before with a merge.

mconnew (approved):

PR #415, opened 2021-08-13
Fix SOAP reflection importer 

### LINKED ISSUE #397: `OperationFormatUse.Encoded` use of `XmlSerializerFormatAttribute` throws PlatformNotSupportedException
I am trying to migrate an existing WCF service to .NET 5.0 via CoreWCF.
Starting the asp.net core application throws `PlatformNotSupportedException` when the `Use` property of the `XmlSerializerFormatAttribute` attribute is set. Removing it dont throws but the invocation always returns null (debugging the service there is a correct result returned in the implementation).

```csharp
[CoreWCF.OperationContractAttribute(Action = "uri:egovmsg/GetAllRecords")]
[CoreWCF.XmlSerializerFormatAttribute(Style = CoreWCF.OperationFormatStyle.Rpc, Use = CoreWCF.OperationFormatUse.Encoded)]
[return: CoreWCF.MessageParameterAttribute(Name = "return")]
EGovMessageDir GetAllRecords();
```

petarpetrovt:
Fixes #397

### DIFF
--- a/src/CoreWCF.Primitives/tests/Soap/ServiceModelSoapTests.cs
+++ b/src/CoreWCF.Primitives/tests/Soap/ServiceModelSoapTests.cs
@@ -48,7 +48,7 @@ namespace Soap
             var expected = new ComplexMessage
             {
                 Date = System.DateTime.Now,
-                Entities = new InnerComplexMessage[]
+                InnerMessages = new InnerComplexMessage[]
                 {
                     new InnerComplexMessage
                     {
@@ -73,7 +73,7 @@ namespace Soap
             var expected = new ComplexMessage
             {
                 Date = System.DateTime.Now,
-                Entities = new InnerComplexMessage[]
+                InnerMessages = new InnerComplexMessage[]
                 {
                     new InnerComplexMessage
                     {
--- a/src/CoreWCF.Primitives/tests/Soap/ServiceModelSoapTests.cs
+++ b/src/CoreWCF.Primitives/tests/Soap/ServiceModelSoapTests.cs
@@ -11,21 +11,7 @@ namespace Soap
     public class ServiceModelSoapTests
     {
         [Fact]
-        public static void OperationFormatUseNone()
-        {
-            System.ServiceModel.ChannelFactory<IEchoSoapService> factory = DispatcherHelper.CreateChannelFactory<EchoSoapService, IEchoSoapService>();
-            factory.Open();
-            IEchoSoapService channel = factory.CreateChannel();
-            ((System.ServiceModel.Channels.IChannel)channel).Open();
-            string echo = channel.Echo("hello");
-            Assert.Equal("hello", echo);
-            ((System.ServiceModel.Channels.IChannel)channel).Close();
-            factory.Close();
-            TestHelper.CloseServiceModelObjects((System.ServiceModel.Channels.IChannel)channel, factory);
-        }
-
-        [Fact]
-        public static void OperationFormatUseEncoded()
+        public static void Echo_OperationFormatUseEncoded()
         {
             System.ServiceModel.ChannelFactory<IEchoSoapService> factory = DispatcherHelper.CreateChannelFactory<EchoSoapService, IEchoSoapService>();
             factory.Open();
@@ -39,7 +25,7 @@ namespace Soap
         }
 
         [Fact]
-        public static void OperationFormatUseLiteral()
+        public static void Echo_OperationFormatUseLiteral()
         {
             System.ServiceModel.ChannelFactory<IEchoSoapService> factory = DispatcherHelper.CreateChannelFactory<EchoSoapService, IEchoSoapService>();
             factory.Open();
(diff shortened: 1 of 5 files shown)


petarpetrovt:
More complex custom SOAP types seem to not serialize correctly with `OperationFormatUse.Encoded` only on dotnet core.

```
Assert.Equal() Failure
Expected: ComplexMessage { Date = 2021-08-13T10:06:12.2896577+03:00, Entities = [InnerComplexMessage { ... }] }
Actual:   ComplexMessage { Date = 0001-01-01T00:00:00.0000000, Entities = null }
```

petarpetrovt:
It seems the problems is actually the request deserialization at `XmlSerializerOperationFormatter.cs` line `461`
> object[] bodyParameters = (object[])serializer.Deserialize(reader, _isEncoded ? GetEncoding(version.Envelope) : null);

returns default property values

When comparing the `serializer`
`NET472` type is `Microsoft.Xml.Serialization.GeneratedAssembly.ArrayOfObjectSerializer6`
`NET5.0|NET3.1|NET2.1` type is `System.Xml.Serialization.XmlSerializer`
both have the same mapping element: `GetComplexMessageEncoded`

petarpetrovt:
@mconnew Can you take a look, i can't find the problem.

mconnew:
An unreleated test failed. It's occasionally flakey with timing, I need to go fix that test. I've kicked off running again and hopefully it will get through this time. Your changes passed.

mconnew:
Actually I was wrong, there is a test failing, it's this one:
```
[xUnit.net 00:00:03.29]     Soap.ServiceModelSoapTests.ComplexMessage_OperationFormatUseEncoded [FAIL]
  Failed Soap.ServiceModelSoapTests.ComplexMessage_OperationFormatUseEncoded [18 ms]
  Error Message:
   Assert.Equal() Failure
Expected: ComplexMessage { Date = 2021-08-13T18:05:00.1472895+00:00, InnerMessages = [InnerComplexMessage { ... }] }
Actual:   ComplexMessage { Date = 0001-01-01T00:00:00.0000000, InnerMessages = null }
```
I'll take a look. It's passing on NetFx (which uses full framework version of WCF) so there might be a problem with the client.

mconnew:
It's not the client, the server is returning something different on NetFx than on .NET Core. This might be an XmlSerializer bug. I'll keep working on it to work out what's going on and I'll open a bug in the dotnet runtime repo if needed.

mconnew:
This turned out to be a WCF client bug. In the code here, the encoding string is calculated but never passed to the Serialize method. This causes the serializer to serialize for SOAP 1.1, but the inline dispatcher channel factory encodes to SOAP 1.2. On .NET Framework, it's correctly serializing using SOAP 1.2 which is why it's working, but on Core we end up serializing as SOAP 1.1 and trying to read it as SOAP 1.2. I modified the dispatcher channel factory to use SOAP 1.1 instead and the tests passed. I'm going to do a quick PR to modify it to accept an optional MessageVersion and then you can rebase and change your tests to explicitly specify SOAP 1.1. Meanwhile I'll try to get this bug fixed in the WCF client. Not sure which version we would ship the fix in considering .NET 6 is coming soon. Do you need this to work with SOAP 1.2? I suspect you are using SOAP 1.1 because this why of encoding is so old.

petarpetrovt:
I need it to work with SOAP 1.1 with client (NET48|NET50) and server (NET50)

mconnew:
@petarpetrovt, I've created a PR with the fix at #417. You can either wait for that to be merged and rebase, or you can cherry-pick my commit and add it to your PR. You need to pass MessageVersion.Soap11 to DispatcherHelper.CreateChannelFactory and then it should all work. I'm going to be busy for a little while so won't get this merged for a few hours.

mconnew:
As you only need Soap 1.1, you don't need the client fix. Do you want to cherry-pick my commit or wait until I've merged the fix?

src/CoreWCF.Primitives/tests/Soap/ServiceModelSoapTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            Assert.Equal("hello", echo);
+            ((System.ServiceModel.Channels.IChannel)channel).Close();
+            factory.Close();
+            TestHelper.CloseServiceModelObjects((System.ServiceModel.Channels.IChannel)channel, factory);
+        }
+
+        [Fact]
+        public static void ComplexMessage_OperationFormatUseEncoded()
+        {
+            System.ServiceModel.ChannelFactory<IEchoSoapService> factory = DispatcherHelper.CreateChannelFactory<EchoSoapService, IEchoSoapService>();
```
mconnew, line 44:
```suggestion
            System.ServiceModel.ChannelFactory<IEchoSoapService> factory = DispatcherHelper.CreateChannelFactory<EchoSoapService, IEchoSoapService>(System.ServiceModel.Channels.MessageVersion.Soap11);
```
If you rebase, this should fix your test.

petarpetrovt: Done

PR #419, opened 2021-08-14
Introduce SecurityTokenAuthenticator.ValidateTokenAsync api

### LINKED ISSUE #375: make SecurityTokenAuthenticator async
This is one of the improvements we should make to the api's. Can you open an issue to make SecurityTokenAuthenticator async?

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/340#discussion_r658333874_

g7ed6e:
Resolves #375.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
@@ -283,7 +283,7 @@ namespace CoreWCF.Channels
                 _clientCertificate = certificate2;
                 try
                 {
-                    SecurityToken token = new X509SecurityToken(certificate2, true);
+                    SecurityToken token = new X509SecurityToken(certificate2, false);
                     ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies;
                     var validationValueTask = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token);
                     authorizationPolicies = validationValueTask.IsCompleted
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
@@ -298,7 +298,12 @@ namespace CoreWCF.Channels
                 try
                 {
                     SecurityToken token = new X509SecurityToken(certificate2, true);
-                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
+                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies;
+                    var validationValueTask = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token);
+                    authorizationPolicies = validationValueTask.IsCompleted
+                        ? validationValueTask.Result
+                        : validationValueTask.AsTask().GetAwaiter().GetResult();
+
                     _clientSecurity = new SecurityMessageProperty
                     {
                         TransportToken = new SecurityTokenSpecification(token, authorizationPolicies),
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
@@ -297,8 +297,8 @@ namespace CoreWCF.Channels
                 _clientCertificate = certificate2;
                 try
                 {
-                    SecurityToken token = new X509SecurityToken(certificate2, false);
-                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateToken(token);
+                    SecurityToken token = new X509SecurityToken(certificate2, true);
+                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
                     _clientSecurity = new SecurityMessageProperty
                     {
                         TransportToken = new SecurityTokenSpecification(token, authorizationPolicies),
@@ -322,13 +322,6 @@ namespace CoreWCF.Channels
             }
             if (_clientCertificate != null)
             {
-                SecurityToken token = new X509SecurityToken(_clientCertificate);
-                ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = SecurityUtils.NonValidatingX509Authenticator.ValidateToken(token);
-                _clientSecurity = new SecurityMessageProperty
-                {
-                    TransportToken = new SecurityTokenSpecification(token, authorizationPolicies),
-                    ServiceSecurityContext = new ServiceSecurityContext(authorizationPolicies)
-                };
                 return _clientSecurity;
             }
             return base.GetRemoteSecurity();
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
@@ -149,21 +149,13 @@ namespace CoreWCF.Channels
         }
 
 
-        bool IChannelBindingProvider.IsChannelBindingSupportEnabled
-        {
-            get
-            {
-                return _enableChannelBinding;
-            }
-        }
+        bool IChannelBindingProvider.IsChannelBindingSupportEnabled => _enableChannelBinding;
 
         public override StreamUpgradeAcceptor CreateUpgradeAcceptor()
         {
             ThrowIfDisposedOrNotOpen();
             return new SslStreamSecurityUpgradeAcceptor(this);
         }
-
-
 
         protected override void OnAbort()
         {
@@ -244,13 +236,7 @@ namespace CoreWCF.Channels
             }
         }
 
-        internal bool IsChannelBindingSupportEnabled
-        {
-            get
-            {
-                return ((IChannelBindingProvider)_parent).IsChannelBindingSupportEnabled;
-            }
-        }
+        internal bool IsChannelBindingSupportEnabled => ((IChannelBindingProvider)_parent).IsChannelBindingSupportEnabled;
 
         protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgradeAsync(Stream stream)
         {
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/StreamUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/StreamUpgradeProvider.cs
@@ -36,7 +36,7 @@ namespace CoreWCF.Channels
 
         protected override TimeSpan DefaultOpenTimeout
         {
-            get { return _closeTimeout; }
+            get { return _openTimeout; }
         }
 
         public virtual T GetProperty<T>() where T : class
--- a/src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
@@ -168,11 +168,11 @@ namespace CoreWCF.Channels
                         SR.Format(SR.NegotiationFailedIO, ioException.Message), ioException));
                 }
 
-                SecurityMessageProperty remoteSecurity = CreateClientSecurity(negotiateStream, _parent.ExtractGroupsForWindowsAccounts);
+                SecurityMessageProperty remoteSecurity = await CreateClientSecurityAsync(negotiateStream, _parent.ExtractGroupsForWindowsAccounts);
                 return (negotiateStream, remoteSecurity);
             }
 
-            private SecurityMessageProperty CreateClientSecurity(NegotiateStream negotiateStream,
+            private async Task<SecurityMessageProperty> CreateClientSecurityAsync(NegotiateStream negotiateStream,
                           bool extractGroupsForWindowsAccounts)
             {
                 IIdentity remoteIdentity = negotiateStream.RemoteIdentity;
@@ -190,7 +190,7 @@ namespace CoreWCF.Channels
                     ClaimsIdentity claimsIdentity = new ClaimsIdentity(remoteIdentity);
                     token = new GenericSecurityToken(remoteIdentity.Name, SecurityUniqueId.Create().Value);
                 }
-                authorizationPolicies = authenticator.ValidateToken(token);
+                authorizationPolicies = await authenticator.ValidateTokenAsync(token);
                 SecurityMessageProperty clientSecurity = new SecurityMessageProperty
                 {
                     TransportToken = new SecurityTokenSpecification(token, authorizationPolicies),
--- a/src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
@@ -1,7 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.ObjectModel;
+using System.Threading.Tasks;
 using CoreWCF.IdentityModel.Policy;
 using CoreWCF.IdentityModel.Tokens;
 
@@ -22,6 +24,12 @@ namespace CoreWCF.IdentityModel.Selectors
 
         public ReadOnlyCollection<IAuthorizationPolicy> ValidateToken(SecurityToken token)
         {
+            // We don't call this internally, so we won't have problems with sync over async causing performance issues
+            return ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
+        }
+
+        public async ValueTask<ReadOnlyCollection<IAuthorizationPolicy>> ValidateTokenAsync(SecurityToken token)
+        {
             if (token == null)
             {
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(token));
@@ -31,7 +39,7 @@ namespace CoreWCF.IdentityModel.Selectors
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityTokenValidationException(SR.Format(SR.CannotValidateSecurityTokenType, this, token.GetType())));
             }
 
-            ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = ValidateTokenCore(token);
+            ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = await ValidateTokenCoreAsync(token);
             if (authorizationPolicies == null)
             {
                 string errorMsg = SR.Format(SR.CannotValidateSecurityTokenType, this, token.GetType());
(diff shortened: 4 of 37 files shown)


mconnew:
We can't remove api's in case someone is using them, at least not in a minor version release. The correct approach is to add the new api as a virtual method so it can be overridden and have the base type implementation call the original api in case someone has an implementation which derives it. So for example, this is the changes needed to be made to SecurityTokenAuthenticator:
```c#
    public abstract class SecurityTokenAuthenticator
    {
        protected SecurityTokenAuthenticator() { }

        public bool CanValidateToken(SecurityToken token)
        {
            if (token == null)
            {
                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(token));
            }
            return CanValidateTokenCore(token);
        }

        public ReadOnlyCollection<IAuthorizationPolicy> ValidateToken(SecurityToken token)
        {
            // We don't call this internally, so we won't have problems with sync over async causing performance issues
            return ValidateTokenAsync(token).GetAwaiter().GetResult();
        }

        public async Task<ReadOnlyCollection<IAuthorizationPolicy>> ValidateTokenAsync(SecurityToken token)
        {
            if (token == null)
            {
                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(token));
            }
            if (!CanValidateToken(token))
            {
                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityTokenValidationException(SR.Format(SR.CannotValidateSecurityTokenType, this, token.GetType())));
            }

            ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = await ValidateTokenCoreAsync(token);
            if (authorizationPolicies == null)
            {
                string errorMsg = SR.Format(SR.CannotValidateSecurityTokenType, this, token.GetType());
                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityTokenValidationException(errorMsg));
            }

            return authorizationPolicies;
        }

        protected abstract bool CanValidateTokenCore(SecurityToken token);
        protected abstract ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(SecurityToken token);
        protected virtual Task<ReadOnlyCollection<IAuthorizationPolicy>> ValidateTokenCoreAsync(SecurityToken token)
        {
            // Default to calling sync implementation to support existing derived types which haven't overridden this method
            return Task.FromResult(ValidateTokenCore(token));
        }
    }
```
This way we don't break any existing code, it's compatible with existing implementations and we can use the new async api in the CoreWCF codebase to get the benefits.

src/CoreWCF.Primitives/src/CoreWCF/Channels/StreamSecurityUpgradeAcceptor.cs
```diff
@@ -11,6 +12,6 @@ protected StreamSecurityUpgradeAcceptor()
         {
         }
 
-        public abstract SecurityMessageProperty GetRemoteSecurity(); // works after call to AcceptUpgrade
+        public abstract Task<SecurityMessageProperty> GetRemoteSecurityAsync(); // works after call to AcceptUpgrade
     }
```
mconnew, line 16: Keep `GetRemoteSecurity()` and add new api as virtual which calls GetRemoteSecurity. See main conversation explanation.

src/CoreWCF.Primitives/src/CoreWCF/Channels/StreamUpgradeProvider.cs
```diff
@@ -36,7 +36,7 @@ protected override TimeSpan DefaultCloseTimeout
 
         protected override TimeSpan DefaultOpenTimeout
         {
-            get { return _closeTimeout; }
+            get { return _openTimeout; }
         }
```
mconnew, line 40: Oops, thanks :)

src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -199,13 +199,13 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
                 return clientSecurity;
             }
 
-            public override SecurityMessageProperty GetRemoteSecurity()
+            public override Task<SecurityMessageProperty> GetRemoteSecurityAsync()
             {
```
mconnew, line 203: By keeping GetRemoteSecurity in the base class and making GetRemoteSecurityAsync virtual and have it call GetRemoteSecurity and wrap the result in Task.FromResult, that means implementations which actually are synchronous like this one can just implement GetRemoteSecurity.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -298,7 +298,7 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
                 try
                 {
                     SecurityToken token = new X509SecurityToken(certificate2, false);
-                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateToken(token);
+                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token).GetAwaiter().GetResult();
                     _clientSecurity = new SecurityMessageProperty
```
mconnew, line 302: This is actually a good example of why to keep the sync methods around. X409SecurityTokenAuthenticator doesn't actually do anything async in it's validation implementation, so we could avoid creating a Task, then getting the Awaiter and then getting the result and could directly call it here.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -314,7 +314,7 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
             return true;
         }
 
-        public override SecurityMessageProperty GetRemoteSecurity()
+        public override async Task<SecurityMessageProperty> GetRemoteSecurityAsync()
         {
```
mconnew, line 318: I did a quick deep dive into the implementation and usage here and this method doesn't need to be async. Other than the fact that an X509Authenticator actually has a sync implementation, even if it was async, this is duplicating effort from ValidateRemoteCertificate. The changes needed are to pass `true` to the second parameter of X509SecurityToken constructor in ValidateRemoteCertificate, and in the `if (_clientCertificate!=null)` code block here return the existing _clientSecurity that was created during ValidateRemoteCertificate.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
```diff
@@ -20,7 +21,7 @@ public bool CanValidateToken(SecurityToken token)
             return CanValidateTokenCore(token);
         }
 
-        public ReadOnlyCollection<IAuthorizationPolicy> ValidateToken(SecurityToken token)
+        public async Task<ReadOnlyCollection<IAuthorizationPolicy>> ValidateTokenAsync(SecurityToken token)
         {
```
mconnew, line 25: I've been thinking about needing to avoid sync over async and I think we'd actually benefit from using ValueTask instead. When the implementation is synchronous, it has zero extra allocation as ValueTask is a struct. When it goes async, it gets backed by a Task and has no additional allocation. Here's a [great blog post by Stephen Toub](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/) about ValueTask, how to use it, and why to use it.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/USerNameSecurityToken.cs
```diff
@@ -7,17 +7,17 @@
 
 namespace CoreWCF.IdentityModel.Tokens
 {
-    internal class USerNameSecurityToken : SecurityToken
+    internal class UserNameSecurityToken : SecurityToken
     {
```
mconnew, line 11: Can you fix the casing on the filename too?

src/CoreWCF.Primitives/src/CoreWCF/Channels/StreamSecurityUpgradeAcceptor.cs
g7ed6e: I removed the GetRemoteSecurityAsync API I intoduced before as you investigated and noticed this API does not need to be async.

src/CoreWCF.Primitives/src/CoreWCF/Channels/WindowsStreamSecurityUpgradeProvider.cs
g7ed6e: I removed the GetRemoteSecurityAsync API I intoduced before as you investigated and noticed this API does not need to be async.

src/CoreWCF.Primitives/src/CoreWCF/Security/SpnegoTokenAuthenticator.cs
```diff
@@ -156,10 +156,10 @@ private ReadOnlyCollection<IAuthorizationPolicy> GetAuthorizationPolicies(IIdent
             }
             else
             {
-                ClaimsIdentity claimsIdentity = new ClaimsIdentity(remoteIdentity);
+                ClaimsIdentity claimsIdentity = new ClaimsIdentity(remoteIdentity); // why is this ClaimsIdentity not used in the line below ?
```
g7ed6e, line 159: I don't know understand why here we create this identity as it is not used later

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
g7ed6e: update: I added a commit to remove sync over async from SpnegoTokenAuthenticator

src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiationTokenAuthenticatorState.cs
```diff
@@ -10,11 +11,9 @@ internal class NegotiationTokenAuthenticatorState : IDisposable
     {
         private bool _isNegotiationCompleted;
         private SecurityContextSecurityToken _serviceToken;
-        private readonly object _thisLock;
+        private readonly AsyncLock _asyncLock = new AsyncLock();
 
-        public NegotiationTokenAuthenticatorState() => _thisLock = new object();
-
-        public object ThisLock => _thisLock;
+        public AsyncLock AsyncLock => _asyncLock;
```
mconnew, line 17: You can actually do this:
```c#
public AsyncLock AsyncLock { get; } = new AsyncLock();
```
It saves having an explicit backing field.

g7ed6e: I removed the redundant backing field declaration

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
```diff
@@ -42,6 +50,14 @@ public ReadOnlyCollection<IAuthorizationPolicy> ValidateToken(SecurityToken toke
         }
 
         protected abstract bool CanValidateTokenCore(SecurityToken token);
-        protected abstract ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(SecurityToken token);
+
+        [Obsolete("Implementers should override ValidateTokenCoreAsync.")]
+        protected virtual ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(SecurityToken token) => throw new NotImplementedException(SR.SynchronousTokenValidationIsDeprecated);
```
g7ed6e, line 55: @mconnew do you still have any concern with this `abstract` made `virtual` ?

mconnew: I checked with some experts on the matter and abstract -> virtual is safe from a backwards compatibility standpoint.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -297,8 +297,8 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
                 _clientCertificate = certificate2;
                 try
                 {
-                    SecurityToken token = new X509SecurityToken(certificate2, false);
-                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateToken(token);
+                    SecurityToken token = new X509SecurityToken(certificate2, true);
+                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
                     _clientSecurity = new SecurityMessageProperty
```
mconnew, line 302: We want to avoid allocation if possible, which is why we're using ValueTask instead of Task. Do something like this instead which will not allocate if using the default built-in authenticator:
```c#
ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies;
var validationValueTask = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token);
if (validationValueTask.IsCompleted)
{
    authorizationPolicies = validationValueTask.Result; // Zero allocation
}
else
{
    authorizationPolicies = validationValueTask.AsTask().GetAwaiter().GetResult();
}
```

This gives the benefits of zero allocation synchronous completion while still supporting async completion.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -297,8 +297,8 @@ protected override async Task<(Stream, SecurityMessageProperty)> OnAcceptUpgrade
                 _clientCertificate = certificate2;
                 try
                 {
-                    SecurityToken token = new X509SecurityToken(certificate2, false);
-                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateToken(token);
+                    SecurityToken token = new X509SecurityToken(certificate2, true);
+                    ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = _parent.ClientCertificateAuthenticator.ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
```
mconnew, line 301: Any reason you told this to clone the certificate when creating an X509SecurityToken? We've already cloned it on line 296 when we make our own instance of X509Certificate2 from the passed in certificate.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
```diff
@@ -21,6 +23,12 @@ public bool CanValidateToken(SecurityToken token)
         }
 
         public ReadOnlyCollection<IAuthorizationPolicy> ValidateToken(SecurityToken token)
+        {
+            // We don't call this internally, so we won't have problems with sync over async causing performance issues
+            return ValidateTokenAsync(token).AsTask().GetAwaiter().GetResult();
+        }
```
mconnew, line 29: Try to avoid allocation when it completes synchronouslys.
```c#
var validationResult = ValidateTokenAsync(token);
if (validationResult.IsCompleted)
{
    return validationResult.Result;
}
else
{
    return validationResult.AsTask().GetAwaiter().GetResult();
}

File: src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/UserNameSecurityTokenAuthenticator.cs [hunk 14]
```diff
@@ -15,13 +16,13 @@ protected UserNameSecurityTokenAuthenticator()
...
         }
 
-        protected override ReadOnlyCollection<IAuthorizationPolicy> ValidateTokenCore(SecurityToken token)
+        protected override ValueTask<ReadOnlyCollection<IAuthorizationPolicy>> ValidateTokenCoreAsync(SecurityToken token)
         {
-            USerNameSecurityToken userNameToken = (USerNameSecurityToken)token;
-            return ValidateUserNamePasswordCore(userNameToken.UserName, userNameToken.Password);
+            UserNameSecurityToken userNameToken = (UserNameSecurityToken)token;
+            return new ValueTask<ReadOnlyCollection<IAuthorizationPolicy>>(ValidateUserNamePasswordCore(userNameToken.UserName, userNameToken.Password));
         }
```
mconnew, line 26: Create new virtual method ValidateUserNamePasswordCoreAsync which calls ValidateUserNamePasswordCore and returns the result wrapped in a ValueTask. Make ValidateUserNamePasswordCore virtual and throw NotImplementedException. Put Obsolete attribute on ValidateUserNamePasswordCore saying to implement ValidateUserNamePasswordCoreAsync. This method can just return the ValueTask returned from ValidateUserNamePasswordCoreAsync as it's the only async call and it's the last one in the method. This class is often implemented to look up the username/password in a SQL database and has been a pain point for blocking a thread when doing something which is async so I know this will help some companies.

src/CoreWCF.Primitives/src/CoreWCF/Security/SecuritySessionServerSettings.cs
```diff
@@ -2342,11 +2345,13 @@ internal async Task<(bool success, bool wasAborted)> WaitForOutputSessionCloseAs
             private async Task<(bool success, bool wasAborted)> WaitForInputSessionCloseAsync(TimeSpan timeout)
             {
                 TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
-                RequestContext context;
+                //RequestContext context;
                 bool wasAborted = false;
```
mconnew, line 2349: Remove commented variable

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/SecurityTokenAuthenticator.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/UserNameSecurityTokenAuthenticator.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Security/SecuritySessionServerSettings.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Channels/SslStreamSecurityUpgradeProvider.cs
g7ed6e: I followed this comment https://github.com/CoreWCF/CoreWCF/pull/419#discussion_r691691284 but i might misunderstood.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Selectors/UserNamePasswordValidator.cs
```diff
@@ -21,13 +24,18 @@ public static UserNamePasswordValidator None
             }
         }
 
-        public abstract void Validate(string userName, string password);
+        [Obsolete("Implementers should override ValidateAsync.")]
+        public virtual void Validate(string userName, string password) => throw new NotImplementedException(SR.SynchronousUserNamePasswordValidationIsDeprecated);
+
+        public virtual Task ValidateAsync(string userName, string password)
+        {
```
mconnew, line 31: Use ValueTask.

g7ed6e: done. please see https://github.com/CoreWCF/CoreWCF/pull/419/commits/532b9f93bffc269719c0a243c20c37d71ec11c62

Issue #337, opened 2021-03-31
performance benchmarking of CoreWCF


birojnayak:
Creating a line item for benchmark CoreWCF and finding potential performance improvements and sharing the result in CoreWCF community.

mconnew:
This will use the work from #273

mconnew:
We are working on both measuring the performance in a more formal way (we know it's in the same ballpark as WCF on .NET Framework) and have some proactive improvements we're already planning for 2.0 (switch from reading/writing a Stream to IDuplexPipe, switch Task -> ValueTask where appropriate to reduce allocations). That's going to take a bit of time to get ramped up so no immediate answers.  

You can break down the service into 3 parts. The IO part where the bytes are received and sent. We've built on top of asp.net core for this so there shouldn't be any significant difference between CoreWCF and gRPC there. The second part is serialization. I'll come back to that. The third part is dispatch of the service call. While CoreWCF has a lot more extensibility and capability here than gRPC, we've copied the general dispatch model from WCF on .NET and made it pay to play where ever possible. What that means is you don't pay a cost (or the cost is just a null check) if you don't use the feature. I suspect there won't be much overhead difference between CoreWCF and a gRPC solution in this part of the stack.  
[...]

PR #273, opened 2020-12-18
Port CoreWCF Perf Crank test

v-joesun:
Port CoreWCF Perf Crank test

PR #420, opened 2021-08-16
Introduce new AsyncLock and use it to await channel open

Danielku15:
This is the promised follow up of the discussion in #399 related to the channel opening not being awaited. 

I took some design decisions which I would like to still discuss, hence opening this PR as draft:

- I also went for the approach ot wrap the SemaphoreSlim to avoid wrong usage of it. My primary concerns were about the handle-leak which is possible when accessing some SemaphoreSlim members. Unless the implementation of SemaphoreSlim changes, we are really safe with this lock against memory or handle leaks. 
- As a result of the statement [here](https://github.com/CoreWCF/CoreWCF/pull/399#issuecomment-898045090), I temporarily flagged the `AsyncLock` as obsolete. I was thinking of extending this PR to replace the `AsyncLock` with the implementation of `SimpleAsyncLock` and introduce it in codebase where reasonable. But before going this step I would prefer the approval of @mconnew to address this topic. 

I have following proposals how we could proceed on this PR:  

a) We keep the new `SimpleAsyncLock` and the old `AsyncLock`, but remove again the `Obsolete` attribute. Like this we integrate the PR. 
**Pros:** changes are small and complexity of using the SemaphoreSlim as lock is hidden from the user. Locks can be replaced gradually without impacting existing codebase. 
**Cons:** there are already quite some wrappers and we increase the number of classes doing similar things. 

b) We fully inline the `SimpleAsyncLock` and keep the old `AsyncLock`. Hence we directly use `SemaphoreSlim` at the places where currently `SimpleAsyncLock` is used and apply a try-finally pattern. 
**Pros:** No additional wrapper and minimal change to the code. In comparison to the other approaches we also avoid 1 struct allocation of the disposal helper. 
**Cons:** Complexity of using `SemaphoreSlim` as lock duplicates across the codebase (try-finally pattern applied everywhere). There is a risk of leaks if somebody uses the `SemaphoreSlim` with the wait handle. If something related to locks will need to be fixed or changed in newer runtimes, we need to touch all the code places to fix the locks again. 

c) We replace the implementation of `AsyncLock` with the proposal of `SimpleAsyncLock`. 
**Pros:** Similar benefits as in a) plus: No code duplication regarding async locks. We also eliminate mentioned leaks of the old implementation.
**Con:** Impact on the codebase is bigger which implies a risk of breaking something. We will lose the lock tracking and other debug features which the old implementation currently offers. 

@mconnew Let me know which path you prefer for CoreWCF 😉


StevenBonePgh:
I like option 'c' of replacing implementation of AsyncLock and using it everywhere, though admittedly I have not reviewed usages. If necessary, additional diagnostics can be created in the replacement to cover some of what is being lost. After one spends days staring at a minidump where async locking went awry and trying to determine under which circumstances the lock was not released, any additional diagnostics would be awesome. This is only possible if a wrapper for async locking exists, so B is out. 

I'd recommend applying some enhancements where the happy path of no lock contention results in no allocations. Steven Toub, writing for the pfx team, has two blog posts I'd suggest adopting in-part:
Toub's implementation of [AsyncLock](https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-6-asynclock/) only differs from your `SimpleAsyncLock` in that it uses the uses the `AsyncSemaphore` rather than `SemaphoreSlim`, and only allocates the disposal helper struct if there is contention, which is quite clever.
 [AsyncSemaphore](https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-5-asyncsemaphore/) as an alternative to `SemaphoreSlim`. This uses a lock(Monitor) internally, and allocates a `TaskCompletionSource` only if contention exists. The benefit here is that it is pretty minimalistic and would act/perform exactly the same on all platforms. The disadvantage is that this implementation does not handle `CancellationToken` out of the box.

src/CoreWCF.Primitives/src/CoreWCF/Runtime/SimpleAsyncLock.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    /// <summary>
+    /// This class provides a locking mechanism compatible with task async patterns.
+    /// </summary>
+    /// <remarks>
+    /// Unlike the <see cref="AsyncLock"/> it does not have the complexity of lock
+    /// tracking and it ensures SemaphoreSlim does not leak memory or handles by hiding
+    /// the access to certain members which would create resources that require disposal.
+    /// </remarks>
+    internal sealed class SimpleAsyncLock
```
mconnew, line 18: Can you rename this to NonReentrantAsyncLock? The problem that AsyncLock was trying to solve was allowing you to continue to take a lock on the same task. Regular monitor locks tie the lock to the thread, but with async code you can hop threads which is why monitor locks don't work. This code can't be reentered and can deadlock. E.g. something like this will deadlock:
```c#
public async Task DoSomething()
{
    using(_asyncLock.TakeLockAsync())
    {
        await DoSomethingElse();
    }
}
public async Task DoSomethingElse()
{
    using(_asyncLock.TakeLockAsync())
    {
        Console.WriteLine("Doing something else");
    }
}
```
My initial approach in AsyncLock was to track lock taken in an AsyncLocal, but that only works if you have a linear set of Task methods. With that approach if you do something like this:
```c#
public async Task DoSomething()
{
    using(_asyncLock.TakeLockAsync())
    {
        var task1 = DoSomethingElse();
        var task2 = DoSomethingElse();
        await Task.WhenAll(task1, task2);
    }
}
public async Task DoSomethingElse()
{
    using(_asyncLock.TakeLockAsync())
    {
        Console.WriteLine("Doing something else");
    }
}
```
Then both task1 and task2 would be able to enter the lock at the same time, which defeats the purpose of the lock. I have an idea how to address this, but the book keeping code is quite complicated and I haven't had the 2+ days available to "do it right". So for now let's rename this to NonReentrantAsyncLock so anybody introducing new usage is very clear about it's limitations. What might be a nice addition is to add some DEBUG only code which detects reentrance usage and asserts. This would be a static AsyncLocal<bool> that you set to true when acquiring the lock, and set to false when releasing. Before acquiring the lock, check to see if the AsyncLocal is true and assert. The only problem with this approach is it would fire if you are using two different NonReentrantAsyncLock instances on the call path. I'm not requiring that for this PR, basically giving you my stream of thought. The only change required here is to rename the class.

src/CoreWCF.Primitives/src/CoreWCF/Runtime/SimpleAsyncLock.cs
```diff
@@ -0,0 +1,61 @@
...
+            private readonly SimpleAsyncLock _asyncLock;
+
+            public SafeSemaphoreRelease(SimpleAsyncLock asyncLock)
+            {
+                _asyncLock = asyncLock;
+            }
+
+            public void Dispose()
+            {
+                _asyncLock._semaphore.Release();
```
mconnew, line 57: This is vulnerable to a double release. Add a null check and after releasing the semaphore, null it out.
```suggestion
                _asyncLock?._semaphore.Release();
                _asyncLock = null;
```
Maybe add an Fx.Assert that _asyncLock is not null to catch this. Fx.Assert is DEBUG conditional so has zero overhead in release builds. I'm not adding Fx.Assert to the suggestion as it requires a `using System.Runtime;` at the top too.

Danielku15: A double release would imply a misusage of the class because you would really have to initiate a double dispose. It is not finalizable and it is expected to the caller to release the lock via dispose. And if we're concerned about double release, we would also need to be concerned about race conditions on the release. There could be still a double release in this case. 

Even though I think it might be a YAGNI, on thread concurrency related code like this is is usually better to be safe then sorry. Murphys Law is calling 😁 Therefore I would rather recommend to go for a Interlocked.Exchange approach here to be really safe for concurrency and misusages: 

```cs
var asyncLock = Interlocked.Exchange(ref _asyncLock, null);
Fx.Assert(asyncLock != null, "Double unlock, this must be avoided!")
asyncLock?._semaphore.Release();
```

src/CoreWCF.Primitives/src/CoreWCF/Runtime/SimpleAsyncLock.cs
Danielku15: Reentrancy is indeed a very strong argument to keep the versions separate. I will rename them both to `NonReentrantAsyncLock` and `ReentrantAsyncLock` to be explicit in both cases. 

This way the reentrant version also can then be tackled separately. I stumbled over the [ReentrantSemaphore of the VSSDK](https://github.com/microsoft/vs-threading/blob/main/src/Microsoft.VisualStudio.Threading/ReentrantSemaphore.cs). We might use it or adopt it into CoreWCF. It has quite a flexible design on the different reentrancy use cases.

Danielku15:
Updates added:
1. Lock classes were renamed and docs updated
2. NonReentrantAsyncLock was optimized not to allocate any new tasks or releasers if no wait was done. (remark of @StevenBonePgh)
3. NonReentrantAsyncLock was simplified to the actually needed cases to avoid handling all the complexity of cancellation token and timeout paths. Can be gratually added again once required. 
4. NonReentrantAsyncLock releaser was made secured against wrong usage scenarios (double release, concurrent releases).

src/Common/src/CoreWCF/Runtime/NonReentrantAsyncLock.cs
```diff
@@ -0,0 +1,94 @@
...
+                    this,
+                    CancellationToken.None,
+                    TaskContinuationOptions.ExecuteSynchronously,
+                    TaskScheduler.Default);
+        }
+
+        /// <summary>
+        /// This releaser does a release on every dispose.
+        /// </summary>
+        private readonly struct ReusableReleaser : IDisposable
```
Danielku15, line 53: I am a bit unsure about which path is best to follow. The remark of @StevenBonePgh to avoid allocations somehow requires us to have a disposable which does a release every time because this instance is reused for multiple unlocks. 

This opens again the path that some outside caller could also make a mistake and do too many releases which was raised as concern by @mconnew . The `OneTimeReleaser` would handle the single release for cases where we really did wait, but I guess it will not be used most of the time.

I think the mix we have right now is not good and we should clean this to one of the following options:

a) We keep the "ResusableReleaser" and use it everywhere. We neglect the risk that somebody misuses the lock, and by this gain the benefit that we can provide the same releaser instance in case no wait was done. 

b) We keep the "OneTimeReleaser" and use it everywhere. We always create new releaser instances and by this remove the benefit of saving some struct allocations. By this we keep the benefit that we prevent wrong usage of the Releasers.

birojnayak:
@Danielku15 @mconnew I would suggest to have a stress testing considering the change is going to impact every flow( to be safe). @Danielku15 here is a client  https://github.com/birojnayak/StressTestForCoreWCF you can use(or modify) to do some stress testing and update the PR, if things looks ok.

Danielku15:
@birojnayak Your client seems to be lacking a bit the insights and measurements that allow judging whether things are fine. I see it as quite some significant effort to still add the right metrics to do a before/after comparison. But you're concerns are legit and the current test suite does not give full indication about the impact of the change. 

I am not sure if there are good test suites already available and the current suite was also suitable for releasing a 0.1 and 0.2 releases. https://github.com/CoreWCF/CoreWCF/pull/273 seems also a good start.

birojnayak:
> @birojnayak Your client seems to be lacking a bit the insights and measurements that allow judging whether things are fine. I see it as quite some significant effort to still add the right metrics to do a before/after comparison. But you're concerns are legit and the current test suite does not give full indication about the impact of the change.
> 
> I am not sure if there are good test suites already available and the current suite was also suitable for releasing a 0.1 and 0.2 releases. #273 seems also a good start.
@Danielku15 you are true... for 0.1/0.2 we ensured this way to avoid memory leak and process not being hanged. Long term objective is to accept #273, and work on #337

mconnew:
@Danielku15, I modified that test client to significantly increase the concurrency and added a mix of long running clients and short running clients as well as the aborts. I ran it for about a day and checked for unexpected exceptions and memory leaks. It's a bit rough and ready and we need to massage it a bit to be configurable and have multiple predefined configurations for stress testing, and we do need automated measurements. Ideally we could have a system which takes the latest build and runs stress tests continuously and detects leaks and unexpected exceptions and reports them (with memory dump), and restarts with each new build. That's quite a bit of infrastructure which also requires ongoing resources. I'd like to get there, but one step at a time.

mconnew:
@Danielku15, I finally got around to writing a non-leaking reentrant async lock. You can find it's implementation [here](https://github.com/mconnew/wcf/blob/FixAsyncLock/src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs). I've pretty aggressively stress tested it on the WCF Core client. Feel free to pull that in, or you can leave you current implementation that you have here.

src/Common/src/CoreWCF/Runtime/NonReentrantAsyncLock.cs
mconnew: It might be worth returning a ValueTask. [Here's a good article](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/) by the designer of ValueTask about it. Reusable is a bit of a misnomer with a struct. It gets copied by value. So on line 28 ReusableReleaser is created on the local stack. When Task.FromResult is called, the stack pointer gets moved to create room for the calling parameters, including space for the entire struct. It then gets copied into that space on the stack. In the FromResult method, it then needs to box the struct as the field in the Task class which stores it is a reference type (IDisposable). With your current api you might as well leave it as a class to begin with as it's going to get boxed anyway. As the lock is internal, you could save the allocation by leaving it as a struct and making it internal, then defining TakeLockAsync to return Task<LockReleaser>. You wouldn't need to have a cached _releaser any more as there would be zero allocation. The using statement doesn't need an IDisposable, it needs an object which implements IDisposable. I just verified that the using statement doesn't cast to IDisposable so there wouldn't be any boxing. Returning a ValueTask provides the zero allocation for the no-wait scenario. Also I personally prefer when avoiding making the method async to place the alternative code into a separate method which is async. So something like this:
```c#
        public ValueTask<LockReleaser> TakeLockAsync()
        {
            var wait = _semaphore.WaitAsync();
            return wait.IsCompleted
                ? new ValueTask<LockReleaser>(new LockReleaser(this))
                : new ValueTask<LockReleaser>(InnerTakeLockAsync(wait);
        }
        public async Task<LockReleaser> InnerTakeLockAsync(Task wait)
        {
            await wait;
            return new LockReleaser(this);
        }
```
This way the only allocation is the _semaphore.WaitAsync. But you can do even better than that!! You can utilize the synchronous Wait method passing a timeout of 0ms. This will return immediately telling you whether it acquired the lock or if it couldn't, with zero waiting. So you can do this:
```c#
        public ValueTask<LockReleaser> TakeLockAsync()
        {
            bool acquired = _semaphore.Wait(0);
            return acquired 
                ? new ValueTask<LockReleaser>(new LockReleaser(this))
                : new ValueTask<LockReleaser>(InnerTakeLockAsync());
        }
        public async Task<LockReleaser> InnerTakeLockAsync()
        {
            await _semaphore.WaitAsync();
            return new LockReleaser(this);
        }
```
This is truly zero allocation when the lock is available.

src/Common/src/CoreWCF/Runtime/NonReentrantAsyncLock.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// Unlike the <see cref="ReentrantAsyncLock"/> it does not provide a reentrant (aka. recursive) locking
+    /// and can therefore only be used in scenarios where we have clear control over the execution path.
+    /// </remarks>
+    internal sealed class NonReentrantAsyncLock
+    {
+        // based of the ideas in https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-6-asynclock/
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
+        private readonly Task<IDisposable> _releaser;
+#if DEBUG
+        private static readonly AsyncLocal<bool> s_reentranceFlag = new AsyncLocal<bool>();
```
mconnew, line 23: This won't work. If you have 2 NonReentrantAsyncLock instances being used in the same call path, when Lock A is acquired, it will set the reentranceFlag, and when Lock B is acquired, it will think it's already been acquired. You need the AsyncLocal to be an instance variable. But then this is where you have problems with memory leaks. Once you set an AsyncLocal to a value, it gets stored in the ExecutionContext. If that ExecutionContext lives for a long time (such as can happen with timers), then you end up growing and growing the ExecutionContext. The solution is to set the AsyncLocal value to null, but you can't do that with a bool. When you set it to null, it actually removes it's storage slot from the ExecutionContext completely. The exception to this is if you add a change notification handler, then it stores that it's a null and will still leak. The fix is to make this an instance variable of type `AsyncLocal<object>`. You set the value to a new object instance when the lock is acquired, and you set it to null when it's released. Your Assert check validates that the Value is null.

src/Common/src/CoreWCF/Runtime/NonReentrantAsyncLock.cs
```diff
@@ -0,0 +1,94 @@
...
+            private readonly NonReentrantAsyncLock _asyncLock;
+
+            public ReusableReleaser(NonReentrantAsyncLock asyncLock)
+            {
+                _asyncLock = asyncLock;
+            }
+
+            public void Dispose()
+            {
+                _asyncLock._semaphore.Dispose();
```
mconnew, line 64: I think you meant Release() and not Dispose(). Disposing it will prevent it being used a second time.

src/Common/src/CoreWCF/Runtime/ReentrantAsyncLock.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace CoreWCF.Runtime
 {
-    internal class AsyncLock
+    internal class ReentrantAsyncLock
     {
```
mconnew, line 15: Can you replace the implementation of this class with the new implementation in dotnet/wcf as that one doesn't have the problems this one does.

mconnew:
@Danielku15, no rush. Just letting you know the AsyncLock implementation has been properly reviewed and tested so can be integrated now. If at some point it looks like things are going to be too busy for you for a long time, please let me know and I can pick this up if needed.

src/Common/src/CoreWCF/Runtime/AsyncLock.cs
```diff
@@ -2,244 +2,155 @@
...
-            if (obj.ThreadContextChanged)
-            {
-                s_heldLocks.Value = obj.PreviousValue;
-            }
+            return TakeLockAsync(Timeout.InfiniteTimeSpan);
         }
 
-        public Task<IDisposable> TakeLockAsync()
+        public Task<IAsyncDisposable> TakeLockAsync(TimeSpan timeSpan)
         {
```
mconnew, line 33: You've added a TimeSpan variant here, but unfortunately that's not supportable. The only way this might be supportable is to go one step further and implement a custom awaiter. The problem is that any timeouts will return a faulted Task containing an exception. This means we never return the IAsyncDisposable to the using block, and we can't put back the original _currentSemaphore in the correct ExecutionContext. Whenever an async method is called, any code executed after an await is running in a copy of the ExecutionContext and no changes to an AsyncLocal will be reflected in the callers ExecutionContext. So returning from an async method effectively undoes any AsyncLocal changes. This means once we've awaited currentSemaphore.WaitAsync(), any changes we make to an AsyncLocal won't be applied in the ExecutionContext of the calling code, even if we catch the exception, modify the AsyncLocal then rethrow.  

This is why SafeSemaphoreRelease.DisposeAsync isn't an async method, it does all the AsyncLocal book keeping first, then returns the ValueTask from DisposeCoreAsync which is an async method. This ensures the book keeping which modifies an AsyncLocal happens on the callers ExecutionContext. As I mentioned, a way around this might be to create a custom awaiter which modifies things before the await continuation callback is called. But I'm not sure if even that will work as I believe the callback uses it's own ExecutionContext to run the continuation. But I might be wrong on that.  

Anyway, the result is that this AsyncLocal implementation can't implement a timeout mechanism. I think this is fine as it's targeted at replacing lock(_thisLock) which doesn't have a timeout mechanism anyway (I know Monitor.Enter does, but that's not what's been targeted here).

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ServerSingletonConnectionReaderMiddleware.cs
```diff
@@ -48,7 +48,8 @@ private async Task EnsureChannelAsync(FramingConnection connection)
         {
             if (_replyChannel == null)
             {
-                using (await _lock.TakeLockAsync())
+                var releaser = await _lock.TakeLockAsync();
+                try
```
mconnew, line 52: You can make usings async like this:
```c#
await using (await _lock.TakeLockAsync())
{
    // ....
}
```
This will make the DisposeAsync call be awaited like you did manually.

src/CoreWCF.Primitives/tests/Extensibility/InstanceProviderTests.cs
```diff
@@ -95,7 +95,11 @@ public void ReleaseInstance(InstanceContext instanceContext, object instance)
 
         public async Task WaitForReleaseAsync(TimeSpan timeout)
         {
-            (await _asyncLock.TakeLockAsync(timeout))?.Dispose();
+            var relaser = (await _asyncLock.TakeLockAsync(timeout))?.DisposeAsync();
+            if (relaser != null)
```
mconnew, line 99: Ah, it makes sense why you added the timeout now. The failure mode of this test should be an exception from the WaitForReleaseAsync method. We expect the release to happen, and the timeout is to enable giving up on that. This test could easily use a semaphoreslim instead similar to the non-reentrant lock you introduced earlier.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ServerSingletonConnectionReaderMiddleware.cs
Danielku15: CoreWCF is targeting .net standard 2.0 [which defaults ot a C# language version of 7.3](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults), await usings are a feature of C# 8.0. 
Therefore I had to manually write a try-finally pattern.

Alternatively CoreWCF could activate `<LangVersion>8.0</LangVersion>` in the Directory.Build.props. IIRC various syntactic sugar features like `await using` should work.

src/Common/src/CoreWCF/Runtime/AsyncLock.cs
Danielku15: [According to the docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.waitasync?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Threading.SemaphoreSlim.WaitAsync);k(SolutionItemsProject);k(DevLang-csharp)%26rd%3Dtrue&view=net-5.0) timeouts should rather result in a task containing false and not an exception. But reading again through my changes they suffer from some different issues. 

To avoid any risks I removed the timeout overload and simply used SemaphoreSlim in `InstanceProviderTests.cs`. I tried to keep the logic close to the original code that used `AsyncLock`. It feels also cleaner because we're not linking the AsyncLock.cs class anymore into the test project.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ServerSingletonConnectionReaderMiddleware.cs
mconnew: Can you bump the lang version up to 8.0? There are some 8.0 features which don't work because there needs to be some runtime classes available, but they will fail to compile if we use them so it's safe.

Danielku15:
I added the requested changes. Overall the PR covers now: 
1. Adaption of the AsyncLock from dotnet/wcf
2. Ensure proper async-await workflow in the channel open (includes some replacements of classic locks to AsyncLock)
3. Upgrade to C# 8.0 to use `await using`.
4. Replace AsyncLock usage with `SemaphoreSlim` in unit tests requiring wait with timeout (also eliminates linked file usage)
 
Out of scope for this PR is: 
* Further adaptions of C# 8.0 features across the codebase. 
* Further replacements of classical locks to async lock beyond the places needed for the async channel open code path. 

Ping @mconnew

mconnew:
Fantastic. Thanks for having the patience to see this through, I know there was a bit of instability in the AsyncLock implementation in the WCF client for a while.

Issue #411, opened 2021-08-12
Make HttpsTransportBindingElement public


mconnew:
With the 0.2.0 release, we now require the transport binding element Scheme to match the server address listening scheme. To use a custom binding with https, you can no longer use HttpTransportBindingElement as the scheme doesn't match, so the Https version needs to be made public.

PR #426, opened 2021-08-18
Fix for issue #411

birojnayak:
Fix for issue #411

### DIFF
--- a/src/CoreWCF.Http/tests/BasicHttpTransportMessage.cs
+++ b/src/CoreWCF.Http/tests/BasicHttpTransportMessage.cs
@@ -66,7 +66,7 @@ namespace BasicHttp
         [InlineData(true)]
         public void BasicHttpsCustomBindingRequestReplyEchoString(bool useHttps)
         {
-            string testString = new string('a', 3000);
+            string testString = new string('a', 4000);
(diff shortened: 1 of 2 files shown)


mconnew:
Can you add a test which uses this in a custom binding?

birojnayak:
> Can you add a test which uses this in a custom binding? That includes a negative test to verify it fails when you use HttpsTransportBindingElement but only have an http endpoint.

@mconnew added the test(for +/- ve scenarios)

src/CoreWCF.Http/tests/BasicHttpTransportMessage.cs
```diff
@@ -61,8 +61,40 @@ public void BasicHttpRequestReplyWithTransportMessageEchoString()
             }
         }
 
-       // [Fact, Description("transport-security-with-certificate-authentication")]
-       // TODO set up in container, tested locally and this works
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void BasicHttpsCustomBindingRequestReplyEchoString(bool httpsMode)
```
mconnew, line 67: Can you rename parameter to useHttps as httpsMode is ambiguous.

PR #427, opened 2021-08-19
Fix sending of fault when MessageSecurityExecption is thrown

mconnew:


mconnew:
@birojnayak, the test failures look to all be gone with this change. It's ready to review.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ExceptionConvertingDuplexPipe.cs
```diff
@@ -0,0 +1,195 @@
...
+            public override Memory<byte> GetMemory(int sizeHint = 0) => _output.GetMemory(sizeHint);
+            public override Span<byte> GetSpan(int sizeHint = 0) => _output.GetSpan(sizeHint);
+
+            Exception ConvertSendException(SocketException socketException)
+            {
+                return ConvertTransferException(socketException, socketException, aborted: false);
+            }
+        }
+
+        static Exception ConvertTransferException(SocketException socketException, Exception originalException, bool aborted)
```
birojnayak, line 85: should be private, I guess ?

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ExceptionConvertingDuplexPipe.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            private PipeReader _input;
+
+            public ExceptionConvertingPipeReader(PipeReader input) => _input = input;
+
+            public override void AdvanceTo(SequencePosition consumed) => _input.AdvanceTo(consumed);
+            public override void AdvanceTo(SequencePosition consumed, SequencePosition examined) => _input.AdvanceTo(consumed, examined);
+            public override void CancelPendingRead() => _input.CancelPendingRead();
+            public override void Complete(Exception exception = null) => _input.Complete(exception);
+            public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
```
birojnayak, line 33: @mconnew wanted to understand why another wrapper around LoggingDuplexPipe? Can't we move the try/catch to LoggingDuplexPipe ?

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ExceptionConvertingDuplexPipe.cs
```diff
@@ -0,0 +1,195 @@
...
+                socketException.ErrorCode == UnsafeNativeMethods.WSAECONNRESET)
+            {
+                if (aborted)
+                {
+                    return new CommunicationObjectAbortedException(SR.TcpLocalConnectionAborted, originalException);
+                }
+                else
+                {
+                    CommunicationException communicationException = new CommunicationException(SR.Format(SR.TcpConnectionResetError, "unknown"), originalException);
+                    //if (TD.TcpConnectionResetErrorIsEnabled())
```
birojnayak, line 103: are the commented code intentional or else you can remove

birojnayak (changes requested):
Is there a way we can write a test case to generate the exception ? Let me know if we can't..

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ExceptionConvertingDuplexPipe.cs
mconnew: LoggingDuplexPipe is only used in Debug builds

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/ExceptionConvertingDuplexPipe.cs
mconnew: Intentionally. At some point I need to add ETW events. The code is littered with TD.XXXXX commented code in anticipation as it will be easier to light up by leaving them in.

mconnew:
We can't write a test to trigger this intentionally. The issue is whether a client disconnects before the server does. If the client does it first, then the server can get a SocketException which was making it all the way to the top and that was causing a test failure. It only seems to happen on slow hosts as I tried really hard to get it to repro on my dev machine and I just couldn't. Our tests have been flakey since you introduced the failing expiring session test for NetTcp and TransportWithMessageCredentials, but with this change, they've been run multiple times with zero failures. So we know this works because the tests have stabilized.

mconnew:
@birojnayak, are you happy with my responses and my small fix?

birojnayak (approved):

PR #434, opened 2021-08-31
Ensure same default hostname and service path is used

JohnLeyva:
Ensure same default hostname and service path is used in dotnet framework and core clients

Accepts --hostname cli argument

### DIFF
--- a/src/Samples/NetCoreClient/Program.cs
+++ b/src/Samples/NetCoreClient/Program.cs
@@ -6,16 +6,21 @@ namespace NetCoreClient
 {
     class Program
     {
-        private static readonly string s_hostname = "localhost";
-
-        static void Main()
+        /// <remarks>
+        /// use commanline argument localhost
+        /// or something similar to indicate the WCF Server hostname
+        /// </remarks>
+        static void Main(string[] args)
         {
-            Settings settings = new Settings().SetDefaults(s_hostname, "EchoService");
+            string hostname = args.Length >= 1 ? args[0] : null;
+
+            Console.Title = "WCF .Net Core Client";
+            Settings settings = ClientLogic.BuildClientSettings(hostname);
 
             static void log(string value) => Console.WriteLine(value);
-            ClientLogic.CallUsingWcf(settings, log);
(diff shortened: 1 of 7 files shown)


src/Samples/NetCoreClient/Program.cs
```diff
@@ -6,16 +6,17 @@ namespace NetCoreClient
 {
     class Program
     {
-        private static readonly string s_hostname = "localhost";
-
-        static void Main()
+        /// <param name="hostname">use commanline argument --hostname localhost
+        /// or something similar to indicate the WCF Server hostname </param>
+        static void Main(string hostname)
```
mconnew, line 11: I'm not comfortable pulling in an alpha release package for command line parsing for a single parameter. I would rather just check if args.Length == 1, then set hostname to that. So something like:
```c#
string hostname = args.Length == 1 ? args[0] : null;
```
It's a nice concept, but I had a look at their code and there's a few problems, like consuming an extra thread unnecessarily for the duration of the app. I know this is samples, but I don't want to encourage other people to use an alpha package.

src/Samples/NetCoreClient/Program.cs
```diff
@@ -6,16 +6,17 @@ namespace NetCoreClient
...
         {
-            Settings settings = new Settings().SetDefaults(s_hostname, "EchoService");
+            Console.Title = "WCF .Net Core Client";
```
mconnew, line 13: If this is run on the command line, will this permanently change the console title or will it revert once the process exits?

src/Samples/NetCoreClient/Program.cs
JohnLeyva: Makes sense, I'll remove it

src/Samples/NetCoreClient/Program.cs
JohnLeyva: When the process exits, the console title reverts to the previous value

JohnLeyva:
Please let me know if something is still pending

mconnew:
This looks good. Sorry it dropped off my radar for a while.

PR #442, opened 2021-09-11
Port mtom support from dotnet/wcf

g7ed6e:
Resolves #10 .

PR #450, opened 2021-09-19
Update SSM to 4.9.0 rc1

g7ed6e:


g7ed6e:
@mconnew is it possible to merge this i would like to rebase it in #442 ?

PR #468, opened 2021-10-07
Allow registering services without known implementation type

### LINKED ISSUE #467: Hosting Service without providing implementation class
CoreWCF seem to require currently that we provide the class implementing the contract, but this is not always possible. It should be sufficient for CoreWCF to work on contract level, and depend on the DI system to provide the implementation. 

Our use case is that our service implementation is not existing as actual class but it is a CastleProxy instance with interceptors. 

In the old-fashioned ServiceHost this worked when you directly used the constructor where you provide the service instance. 

When you are doing this today: 
```cs
IMyService apiImplementation = MyServiceProxyFactory.CreateProxy();
var webHost = WebHost.CreateDefaultBuilder()
    .ConfigureServices(services =>
    {
        services.AddSingleton(apiImplementation);
    })
    .UseNetTcp(8080)
    .Configure(app =>
    {
        {
            builder.AddService<IUAMAdminServiceContract>();
            builder.AddServiceEndpoint<IMyService, IMyService>(
                new NetTcpBinding() { .. },
                new Uri($"net.tcp://localhost:8081/MyService/")
            );
        });
    })
    .Build();
webHost.Start();
```

It fails with: 
```
   at System.RuntimeTypeHandle.VerifyInterfaceIsImplemented(RuntimeTypeHandle interfaceHandle)
   at System.RuntimeType.GetInterfaceMap(Type ifaceType)
   at CoreWCF.Description.TypeLoader`1.GetIOperationAttributesFromType[TOperation](OperationDescription opDesc, Type targetIface, Type implType)
   at CoreWCF.Description.TypeLoader`1.<>c__DisplayClass20_0.<AddBehaviors>b__0(Type currentType, KeyedByTypeCollection`1 behaviors)
   at CoreWCF.Description.TypeLoader`1.AddBehaviorsAtOneScope[IBehavior,TBehaviorCollection](Type type, TBehaviorCollection descriptionBehaviors, ServiceInheritanceCallback`1 callback)
   at CoreWCF.Description.TypeLoader`1.ApplyServiceInheritance[IBehavior,TBehaviorCollection](TBehaviorCollection descriptionBehaviors, ServiceInheritanceCallback`1 callback)
   at CoreWCF.Description.TypeLoader`1.AddBehaviors(ContractDescription contractDesc, Boolean implIsCallback, ContractReflectionInfo reflectionInfo)
   at CoreWCF.Description.TypeLoader`1.LoadContractDescriptionHelper(Type contractType, Object serviceImplementation)
   at CoreWCF.Description.TypeLoader`1.LoadContractDescription(Type contractType)
   at CoreWCF.Description.ContractDescription.GetContract[TService](Type contractType)
   at CoreWCF.ServiceHostObjectModel`1.CreateDescription(IDictionary`2& implementedContracts)
   at CoreWCF.ServiceHostBase.InitializeDescription(UriSchemeKeyedCollection baseAddresses)
   at CoreWCF.ServiceHostObjectModel`1..ctor(IServiceProvider serviceProvider, ServiceBuilder serviceBuilder, ILogger`1 logger)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at CoreWCF.Description.DispatcherBuilder.BuildDispatcher[TService](ServiceConfiguration`1 serviceConfig, IServiceProvider services)
   at CoreWCF.Configuration.ServiceConfiguration`1.GetDispatchers()
   at CoreWCF.Configuration.DispatcherBuilderImpl.BuildDispatchers(Type serviceType)
   at CoreWCF.Channels.Framing.NetMessageFramingConnectionHandler.BuildAddressTable(IServiceProvider services)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at CoreWCF.Channels.Framing.NetMessageFramingConnectionHandler.OnServiceBuilderOpened(Object sender, EventArgs e)
   at CoreWCF.Channels.CommunicationObject.OnOpened()
--- End of stack trace from previous location ---
   at CoreWCF.Configuration.WrappingIServer.StartAsync[TContext](IHttpApplication`1 application, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.WebHost.StartAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.WebHost.Start()
   at MyApp.MyService.OnStart(String[] args)
   at MyApp.MyService.Start(String[] args)
   at MyApp.Program.Main()
```

I think the problem lies here in the following code: 

https://github.com/CoreWCF/CoreWCF/blob/2d3b77ebb11619bd33ca6aa084f8502da25c53c7/src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs#L428-L441

I think we need a path here that also the "implType" is an interface and then we need to parse the operations simply from the contract definition. Maybe something like the following code flow could work already: 

```cs
var ifaceMap = default(InterfaceMapping);
bool useImplAttrs = false;
if (implType != null)
{
    if (targetIface.IsAssignableFrom(implType) && targetIface.IsInterface)
    {
        if(implType.IsInterface) 
        {
            // implType is also an interface and not a class, no need to check interface mapping
            useImplAttrs = false;
        }
        else
        {
            ifaceMap = implType.GetInterfaceMap(targetIface);
            useImplAttrs = true;
        }
    }
    else
    {
        // implType does not implement any methods from the targetIface, so there is nothing to do
        return result;
    }
}
```

Danielku15:
Fixes #467 

As described in the issue CoreWCF does not allow registering services purely using the contract interface and a provided implementation through DI (yet!). This PR adds support for exactly this. 

It seems there was only one execution path on the TypeLoader which needed to be adapted.

### DIFF
--- a/src/CoreWCF.Http/tests/InterfaceOnlyServiceTest.cs
+++ b/src/CoreWCF.Http/tests/InterfaceOnlyServiceTest.cs
@@ -1,0 +1,113 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using CoreWCF.Configuration;
+using Helpers;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BasicHttp
+{
+    public class InterfaceOnlyServiceTest
+    {
+        private readonly ITestOutputHelper _output;
+
+        public InterfaceOnlyServiceTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Fact]
+        public void BasicHttpRequestReplyEchoString()
+        {
+            string testString = new string('a', 3000);
+            IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output).Build();
+            using (host)
+            {
+                host.Start();
+                System.ServiceModel.BasicHttpBinding httpBinding = ClientHelper.GetBufferedModeBinding();
+                var factory = new System.ServiceModel.ChannelFactory<ClientContract.IEchoService>(httpBinding,
+                    new System.ServiceModel.EndpointAddress(
+                        new Uri("http://localhost:8080/BasicWcfService/basichttp.svc")));
+                ClientContract.IEchoService channel = factory.CreateChannel();
+                string result = channel.EchoString(testString);
+                Assert.Equal(testString, result);
+
+                var interceptor = host.Services.GetRequiredService<EchoServiceInterceptor>();
+                Assert.Equal(1, interceptor.NumberOfIntercepts);
+            }
+        }
+
+        private class EchoServiceInterceptor : ServiceContract.IEchoService
+        {
+            private readonly ServiceContract.IEchoService _underlyingService;
+
+            public int NumberOfIntercepts { get; private set; }
+
+            public EchoServiceInterceptor(ServiceContract.IEchoService underlyingService)
+            {
+                _underlyingService = underlyingService;
+            }
+
+            public string EchoString(string echo)
+            {
+                NumberOfIntercepts++;
+                return _underlyingService.EchoString(echo);
+            }
+
+            public Stream EchoStream(Stream echo)
+            {
+                NumberOfIntercepts++;
+                return _underlyingService.EchoStream(echo);
+            }
+
+            public async Task<string> EchoStringAsync(string echo)
+            {
+                NumberOfIntercepts++;
+                return await _underlyingService.EchoStringAsync(echo);
+            }
+
+            public async Task<Stream> EchoStreamAsync(Stream echo)
+            {
+                NumberOfIntercepts++;
+                return await _underlyingService.EchoStreamAsync(echo);
+            }
+
+            public string EchoToFail(string echo)
+            {
+                NumberOfIntercepts++;
+                return _underlyingService.EchoToFail(echo);
+            }
+        }
+
+        internal class Startup
+        {
+            public void ConfigureServices(IServiceCollection services)
+            {
+                services.AddServiceModelServices();
+                services.AddSingleton(new EchoServiceInterceptor(new Services.EchoService()));
+                // register service implementation in DI and let CoreWCF resolve it as instance
+                // in some cases implementation could be without known type (e.g. could be a castle proxy)
+                services.AddSingleton<ServiceContract.IEchoService>(sp => sp.GetRequiredService<EchoServiceInterceptor>());
+            }
+
+            public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+            {
+                app.UseServiceModel(builder =>
+                {
+                    // we fully register the service through interfaces
+                    builder.AddService<ServiceContract.IEchoService>();
+                    builder.AddServiceEndpoint<ServiceContract.IEchoService, ServiceContract.IEchoService>(
+                        new CoreWCF.BasicHttpBinding(), "/BasicWcfService/basichttp.svc");
+                });
+            }
+        }
+    }
+}
+
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
@@ -204,7 +204,7 @@ namespace CoreWCF.Description
             return contract;
         }
 
-        //This needs proper review. 
+        //This needs proper review.
         private void AddAuthorizeOperations(ContractDescription contractDesc, bool implIsCallback, ContractReflectionInfo reflectionInfo)
         {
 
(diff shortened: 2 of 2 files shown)


Danielku15:
> This needs some checks elsewhere to make sure we have a clean exception in the case that you specify an interface but it's not available in DI. In that case, we fall back to try to create it with Activator.CreateInstance. Can you add a negative test where it will fail in this condition and make sure we have a friendly error message explaining that the type needs to be a non-abstract class or the interface needs to be in DI.

I will check what I can do there. From what I've seen on my first tests is that the error is delayed until a client is calling the service and CoreWCF tries to create an instance of the service by resolving it from the DI. On client side this leads to a `FaultException` triggered through the `ErrorBehavior.ProvideFaultOfLastResort` caused by a 
`System.InvalidOperationException: The service implementation object was not initialized or is not available.
   at CoreWCF.Dispatcher.SyncMethodInvoker.InvokeAsync(Object instance, Object[] inputs)`

But as far I can see, this error might not be exposed well to the developers. I'll need to check what I can do there. But actually this kind of error is not really specific to my particular change of allowing interface only registrations. AFAIK it was already a CoreWCF feature to provide service instances through the ASP.net core DI container, this is not new. Any error in this path is similar as it would have been before when creation of a service instance failed (e.g. due to misconfiguration of the DI). 

> How does this behave with different instance context modes? Does it pick up a ServiceBehavior attribute from the interface?

The `ServiceBehaviorAttribute` is not allowed on interfaces but only on classes. Due to that CoreWCF seems to the defaults through `ServiceDescription.EnsureBehaviorAttribute` where the default value is `PerSession`. The actual instance is then created through `DependencyInjectionWithLegacyFallbackInstanceProvider` which always asks the DI system for an instance where it is created depending on the corresponding registrations.

Might be similar to https://github.com/CoreWCF/CoreWCF/issues/353 To define the `InstanceContextMode` for the interface also through DI like explained [here](https://github.com/CoreWCF/CoreWCF/issues/353#issuecomment-832844720)

mconnew:
I created an issue for improving detection of whether DI has a type as a provided service which would improve this scenario, but I'm not requiring it for this as I'm not even sure if we should do it before moving to .NET 6 as a minimum requirement. I'm open to opinions on this.

PR #473, opened 2021-10-14
Add support for scoped ctor dependency

### LINKED ISSUE #476: Scoped ctor dependencies have a singleton lifetime
When registering service within DI, the underliying scoped lifetime dependencies have a singleton lifetime. I think this is due to the probe service resolution and the fact that next resolution are not performed in a dedicated `IServiceScope`.

g7ed6e:
Resolves #476 

The below `IInstanceProvider` implementation is dead code. may i remove it ?https://github.com/CoreWCF/CoreWCF/blob/ae53e521c7bd0fe90815e7810dc4ecebfbc74af9/src/CoreWCF.Primitives/src/CoreWCF/DependencyInjectionInstanceProvider.cs#L10

This one does not use `IServiceScope`. Should it be aligned ?
https://github.com/CoreWCF/CoreWCF/blob/7ef9498f66811fa7b678e5a4c0091d72d101012a/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InstanceBehavior.cs#L249


mconnew:
We can easily add support for InstanceContextMode.Single. If you look at the code in ServiceBehaviorAttribute, we can attach a class to the singleton InstanceContext.
https://github.com/CoreWCF/CoreWCF/blob/349aad21064fbc698e6abb30b2c1a0873861fe05/src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs#L236-L247

As we aren't exposing the class ScopedServiceProviderExtension as a type and users will access it by using `Find<IServiceProvider>`, we can create another small wrapping class called something like ServiceProviderExtension which implements `IExtension<InstanceContext>, IServiceProvider`. Something like this:
```c#
        internal class ServiceProviderExtension : IExtension<InstanceContext>, IServiceProvider
        {
            private readonly IServiceProvider _serviceProvider;

            public ScopedServiceProviderExtension(IServiceProvider serviceProvider) => _serviceProvider = serviceProvider;

            public void Attach(InstanceContext owner)
            {
                // intentionally left blank
            }

            public void Detach(InstanceContext owner)
            {
                // intentionally left blank
            }

            public object GetService(Type serviceType) => _serviceProvider.GetService(serviceType);
        }
```

There's a part of me that says we shouldn't do this for InstanceContextMode.Single though. The problem is ServiceProvider keeps a reference to any instance which implements IDisposable/IAsyncDisposable. This is fine if the type is registered as a Singleton as you are returned the same instance every time. But if the type is defined as Transient, then there's a memory leak. But we also need to consider the operation parameter injection we've been discussing. Without this available for all InstanceContext modes, it may be harder to implement that functionality without making things more complicated.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DependencyInjectionWithLegacyFallbackInstanceProvider.cs
```diff
@@ -3,66 +3,127 @@
 
 using System;
 using CoreWCF.Channels;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace CoreWCF.Dispatcher
 {
     internal class DependencyInjectionWithLegacyFallbackInstanceProvider : IInstanceProvider
     {
+        class ScopedServiceProviderExtension : IExtension<InstanceContext>, IServiceProvider, IDisposable
```
mconnew, line 12: I just noticed this is missing an accessor. I think it defaults to private as it's an inner class. Can you make this explicit.

g7ed6e: done!

g7ed6e:
> We can easily add support for InstanceContextMode.Single. If you look at the code in ServiceBehaviorAttribute, we can attach a class to the singleton InstanceContext.
> 
> https://github.com/CoreWCF/CoreWCF/blob/349aad21064fbc698e6abb30b2c1a0873861fe05/src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs#L236-L247
> 
> As we aren't exposing the class ScopedServiceProviderExtension as a type and users will access it by using `Find<IServiceProvider>`, we can create another small wrapping class called something like ServiceProviderExtension which implements `IExtension<InstanceContext>, IServiceProvider`. Something like this:
> 
> ```cs
>         internal class ServiceProviderExtension : IExtension<InstanceContext>, IServiceProvider
>         {
>             private readonly IServiceProvider _serviceProvider;
> 
>             public ScopedServiceProviderExtension(IServiceProvider serviceProvider) => _serviceProvider = serviceProvider;
> 
>             public void Attach(InstanceContext owner)
>             {
>                 // intentionally left blank
>             }
> 
>             public void Detach(InstanceContext owner)
>             {
>                 // intentionally left blank
>             }
> 
>             public object GetService(Type serviceType) => _serviceProvider.GetService(serviceType);
>         }
> ```
> 
> There's a part of me that says we shouldn't do this for InstanceContextMode.Single though. The problem is ServiceProvider keeps a reference to any instance which implements IDisposable/IAsyncDisposable. This is fine if the type is registered as a Singleton as you are returned the same instance every time. But if the type is defined as Transient, then there's a memory leak. But we also need to consider the operation parameter injection we've been discussing. Without this available for all InstanceContext modes, it may be harder to implement that functionality without making things more complicated.

I think we can avoid memory leak having the source generator emit a slightly different method when we detect `InstanceContextMode.Single`. Something like this:
```csharp
public partial class SimpleService : ISimpleService
{
    public string Echo(string echo)
    {
        IServiceProvider serviceProvider = OperationContext.Current.Extensions.Find<IServiceProvider>();
        if (serviceProvider == null) throw new InvalidOperationException("Missing IServiceProvider in InstanceContext extensions"); // Need to work on error, maybe link to documentation
        using (var scope = serviceProvider.CreateScope())
        {
            Func<string, string> identity = scope.GetService<Func<string, string>>();
            return Echo(echo, identity);
        }
    }
}
```

mconnew:
That's a good idea. Reviewing your latest changes now.

src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs
```diff
@@ -249,6 +274,7 @@ private void ApplyInstancing(ServiceDescription description, ServiceHostBase ser
                                 }
 
                                 singleton.AutoClose = false;
+                                singleton.Extensions.Add(new ServiceProviderExtension(_serviceProvider));
                             }
```
mconnew, line 278: Would it be worth adding a null check here? If this is expected to always not be null (which as far as I can tell is the case), can you add an Fx.Assert so we fail during testing if this isn't the case.

mconnew:
Just needs a validation assert on the service provider then this is ready to merge. This is great that we're now expanding CoreWCF to go beyond what WCF could do.

src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs
g7ed6e: > Just needs a validation assert on the service provider then this is ready to merge. This is great that we're now expanding CoreWCF to go beyond what WCF could do.

@mconnew  I just pushed #493 There'is is still a lot to do in the generator but that's a starting point

g7ed6e:
> Just needs a validation assert on the service provider then this is ready to merge. This is great that we're now expanding CoreWCF to go beyond what WCF could do.

@mconnew I just pushed #493 There'is is still a lot to do in the generator but that's a starting point

src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs
g7ed6e: done

Issue #103, opened 2020-02-20
Need mechanism to translate between FaultException<T> types


Treit:
System.ServiceModel.FaultException<T> is .NET Standard compliant, so code from .NET Standard libraries that are hosted in CoreWCF may throw this FaultException type and assume it should result in the correct SOAP fault being returned to the caller.

CoreWCF defines it's own CoreWCF.FaultException<T> type, however, so .NET Standard code that should ideally work whether hosted in either CoreWCF or a traditional .NET Framework WCF service does not work as desired.

When porting a large existing service to CoreWCF, it is highly desirable to be able to share code between the existing service and the new one. The lack of a common FaultException<T> type makes this problematic.

We need a mechanism to somehow translate between the two FaultException<T> types, so code hosted in a CoreWCF process that throws a System.ServiceModel.FaultException<T> is able to successfully transmit the intended fault back to the caller.

jnyrup:
Anyone has a workaround, even a partial one, or pointers to where in the WCF pipeline this should be implemented?

Treit:
@jnyrup I did come up with a workaround for this, I will try to dig it up and share an example

Treit:
@jnyrup below is the attribute I ended up implementing to work-around this problem.

This uses multi-targeting in the csproj file:

```xml
  <PropertyGroup Condition=" '$(TargetFramework)' == 'net472'">
    <DefineConstants>NET472;NETFULL</DefineConstants>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(TargetFramework)' == 'net60'">
    <DefineConstants>NETCORE</DefineConstants>
  </PropertyGroup>
```

Then we used conditional compilation to pull in the right namespaces and implemented the IErrorHandler to translate to the CoreWCF.FaultException if some code somewhere in the service threw System.ServiceModel.FaultException.

```cs
using System;
#if NETFULL
using System.ServiceModel;
using System.ServiceModel.Channels;
using System.ServiceModel.Description;
using System.ServiceModel.Dispatcher;
using Message = System.ServiceModel.Channels.Message;
#else
using CoreWCF;
using CoreWCF.Channels;
using CoreWCF.Description;
using CoreWCF.Dispatcher;
using Message = CoreWCF.Channels.Message;
#endif

namespace HostedService.Common
{
    [AttributeUsage(AttributeTargets.Class)]
    public class SharedErrorHandlerAttribute : Attribute, IErrorHandler, IServiceBehavior
    {
[...]

PR #481, opened 2021-10-23
Example Dependency Injection for dotnet core and ILogger

JohnLeyva:


src/Samples/DesktopServer/EchoService.cs
```diff
@@ -1,25 +1,76 @@
...
+namespace ServerLogic
 {
-    public class EchoService : Contract.IEchoService
+    [ServiceBehavior(IncludeExceptionDetailInFaults = true)]
+    public class EchoService : IEchoService
     {
+        private FaultException CreateFault<T>(T detail, string reason, string code)
+        {
+#if NETFRAMEWORK
+            return new System.ServiceModel.FaultException<T>(detail, new FaultReason(reason), new FaultCode(code));
```
mconnew, line 19: Where is there conditional compilation here? This file is in the DesktopServer folder so will only be used with .NET Framework.

src/Samples/NetCoreServer/NetCoreServer.csproj
```diff
@@ -7,6 +7,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
+    <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
+    <PackageReference Include="System.ServiceModel.Http" Version="4.8.1" />
```
mconnew, line 11: 4.9.0 version of the packages are now available.

src/Samples/DesktopServer/EchoService.cs
JohnLeyva: This is a copy of the file in src/Samples/NetCoreServer/EchoService.cs, 
Thought in creating a Shared project to avoid duplication

src/Samples/NetCoreServer/NetCoreServer.csproj
JohnLeyva: Updated to 4.9.0

src/Samples/DesktopServer/EchoService.cs
g7ed6e: I personaly think that sample's code should allow users to quickly get how it is expected to work and be configured, without any complexity. Thus factorizing this code lead to user not being able to copy paste and increase the time required to figure out the required steps to get started.
Please have a look to the below repos to see how they are duplicating code to ease user understanding:
https://github.com/IdentityServer/IdentityServer4/tree/main/samples
https://github.com/dotnet/roslyn-sdk/tree/main/samples/CSharp/SourceGenerators/SourceGeneratorSamples

JohnLeyva: The reasoning behind using #if NETFRAMEWORK is to indicate that System.ServiceModel should be used with dot net framework otherwise CoreWCF in case a FaultException needs to be thrown.

Thus, people migrating or trying new functionality won't lose time figuring out why Faults might not work if this is not taken into account.

I did not use a Share Project because indeed thought might not be necessary

g7ed6e: Should be fixed by https://github.com/CoreWCF/CoreWCF/pull/488

JohnLeyva: NetCoreServer still fails if System.ServiceModel.FaultException<EchoFault> is thrown.
It works when using CoreWCF<EchoFault>

g7ed6e: Can you confirm you are trying to throw a SSM Fault.Exception from your CoreWCF codebase ?
If this is the case the related issue is https://github.com/CoreWCF/CoreWCF/issues/103.
However i do not understand why this really specific use case is in a sample.

JohnLeyva: Yes, NetCoreServer uses CoreWCF and #103 describes the issue.

Given that the general MS guidance is to use other technologies, it should be expected that most code should try to use System.ServiceModel types instead of CoreWCF as the most common case will be migration.

Examples should try to guide CoreWCF adopters to cover the most common scenarios, such as returning Faults or using Dependency injection. A complementary alternative is to provide documentation and/or tutorials.

Although the information can be found in issues, is common that the solution/workaround is not clear and takes important time to find the right way to apply.

JohnLeyva: btw, just noticed #488 hasn't been completed yet, so my latest test was with just with the lastest main.

mconnew:
I've been thinking about this for a while as I'm not completely happy with how the samples have evolved, and I've also had some feedback from someone who was trying to get started and echoed my same thoughts. The samples are getting a bit over engineered. They are intended to be super simple examples that you could even copy and use as the beginning of your own app to get started. It's time to create a Samples repo and move these samples over there. Each usage scenario should be in its own folder with a small README.md  file explaining what the sample does, and maybe a few remarks explaining where and what the code is that enables the sample scenario. The idea is to have as little common source code as possible, preferably none, so that each sample folder stands alone.  

@JohnLeyva, you have contributed quite a lot with the Samples. Would you be interested in helping to lead this effort?

JohnLeyva:
@mconnew, happy to help with it  :)

PR #484, opened 2021-10-27
Fix basepath being broken for subsequent middleware

### LINKED ISSUE #483: CoreWCF breaks PathBase for the rest of the app
the following middlewares are called inside _branch, before Path and PathBase are reset

image

This is with CoreWcf version 0.3.1.1, which isn't the latest version, but the last change here was in march: https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs#L38

mconnew:
Fixes #483

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
@@ -13,6 +13,7 @@ namespace CoreWCF.Channels
 {
     public partial class ServiceModelHttpMiddleware
     {
+        private const string RestorePathsDelegateItemName = nameof(ServiceModelHttpMiddleware) + "_RestorePathsDelegate";
         private readonly IApplicationBuilder _app;
         private readonly IServiceBuilder _serviceBuilder;
         private readonly IDispatcherBuilder _dispatcherBuilder;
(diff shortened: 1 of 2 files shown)


mconnew:
I just spent way too long to try and get relative paths working correctly. It's not going to be possible without changing some of the fundamental API's in CoreWCF (IServiceBuilder, and probably EndpointAddress, to start with) so isn't doable until a major version change, e.g. between 1.0 and 2.0. Here's the root of the problem. When adding a service endpoint, we need 3 pieces of information. We need the scheme (e.g. http or net.tcp), the port number, and the url. The port number is optional to specify as we can use the default port number for the scheme. We have some ambiguity here which needs to be tightened up at a later date and there's a bit of a hack to make it work today so currently the port number isn't as important, but it should be. In WCF, a service can only have one base address per scheme, so multiple listening ports for a single scheme aren't supported. This means you can't listen for http requests on port 80 and 81 in the same service instance. CoreWCF loosens that constraint and accepts requests for any port for the scheme. We need a way to specify a particular service only listens for requests on a single port, but that's a different issue, just giving some background.  

When you add a service endpoint CoreWCF looks at all aspnetcore server addresses (when you add net.tcp, we add that to the IServerAddresses feature collection) to find one which matches the binding scheme. It then combines the passed with q base address into a Uri object. So if you have a server listening address of http://localhost/ and specify "/BasePath/MyService.svc", we effectively do `new Uri(new Uri("http://localhost/", "/BasePath/MyService.svc"))`. The Uri class can have the concept of a relative path, but only if you don't specify a scheme and hostname. Executing `new Uri("RelativePath/MyService.svc")` is valid and you can retrieve the relative path. But the moment you need to specify a scheme and optionally port number, the Uri class can't represent a relative path and converts it to a base path, pretending the relative path starts at the root. This address is stored an `ServiceEndpoint.Address` which is of type `EndpointAddress` which stores a Uri of the listening endpoint. We don't have this problem in WCF because you either specify the base path when you construct the ServiceHost, or IIS tells WCF what the base path is.  

So basically the problem comes down to:
* There is no way in aspnetcore to find what the base path is until you receive a request.
* You can't encode into a Uri instance that a path is relative if you have a scheme.
* CoreWCF fundamentally expects to know what the full listening Url is at initialization.

Given that the first two problems are out of our control, the only option is a significant overhaul of some core concepts of CoreWCF, which then has consequences in other API's where we expose addresses. I might be able to come up with a solution, but it's going to be disruptive so somebody, so a servicing patch isn't where I can make that sort of change.

birojnayak (approved):

PR #485, opened 2021-10-29
Add NuGet package status

paulomorgado:


mconnew:
I think this should be a bit more prominent. I don't think you should have to scroll to get to this info. Can you place it just before the announcements section?

paulomorgado:
> I think this should be a bit more prominent. I don't think you should have to scroll to get to this info. Can you place it just before the announcements section?

Done!

mconnew:
Thank you for this. Sometimes I get too focused on the code implementation and don't take a step back to look at other areas to improve.

paulomorgado:
That's the value of open-source and community.

PR #488, opened 2021-10-30
Fix FaultException<T> namespace not correctly handled when using SSM FaultContractAttribute

### LINKED ISSUE #472: System.ServiceModel.FaultContractAttribute typeof(T[ ])  is generated as ArrayT instead of T in FaultResponse
CoreWCF Is required to make it work, but svcutil generates System.ServiceModel

    [System.ServiceModel.ServiceContractAttribute(Namespace="http://my.com/service")]
    [CoreWCF.ServiceContractAttribute(Namespace = "http://my.com/service")]

// WCF Client does not catch it as a normal FaultException but not  FaultException<EchoFault>
                <EchoFault xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
                    <Text>TextVal</Text>
                </EchoFault>

// WCF Client catches the fault using FaultException<EchoFault>
                <EchoFault xmlns="http://my.com/service" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
                    <Text>TextVal</Text>
                </EchoFault>

the missing part is the namespace xmlns="http://my.com/service"

g7ed6e:
Fixes #472

### DIFF
--- a/src/CoreWCF.Http/tests/ServiceWithCoreWCFFaultContractTest.cs
+++ b/src/CoreWCF.Http/tests/ServiceWithCoreWCFFaultContractTest.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Net.Http;
+using System.Runtime.InteropServices;
 using System.ServiceModel.Channels;
 using System.Text;
 using System.Threading.Tasks;
@@ -78,15 +79,30 @@ namespace BasicHttp
             var responseBody = await response.Content.ReadAsStringAsync();
             //_output.WriteLine(responseBody);
 
-            const string expected = "<s:Envelope xmlns:s=\"http://schemas.xmlsoap.org/soap/envelope/\">" +
+            string expected = "<s:Envelope xmlns:s=\"http://schemas.xmlsoap.org/soap/envelope/\">" +
                 "<s:Body><s:Fault>"
                 +"<faultcode>s:Client</faultcode>"
-                +"<faultstring xml:lang=\"en-US\">The creator of this fault did not specify a Reason.</faultstring>"
+                +"<faultstring"
+                + $"{GetXmlLangAttributeOrNot()}"
+                + ">The creator of this fault did not specify a Reason.</faultstring>"
                 +"<detail>"
                 +"<SSMCompatibilityFault xmlns=\"https://ssm-fault-contract-compatibility.com\" xmlns:a=\"http://schemas.datacontract.org/2004/07/Services\" xmlns:i=\"http://www.w3.org/2001/XMLSchema-instance\">"
                 +"<a:Message>An error occured</a:Message></SSMCompatibilityFault></detail></s:Fault></s:Body></s:Envelope>";
 
             Assert.Equal(expected, responseBody);
+
+            string GetXmlLangAttributeOrNot()
+            {
+                const string enUsXmlLang = " xml:lang=\"en-US\"";
+#if NETCOREAPP3_1_OR_GREATER
+                return RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ?
+                    string.Empty
+                    : enUsXmlLang;
+#else
+                return enUsXmlLang;
+#endif
+            }
+
         }
 
         public class Startup
--- a/src/CoreWCF.Http/tests/ServiceWithCoreWCFFaultContractTest.cs
+++ b/src/CoreWCF.Http/tests/ServiceWithCoreWCFFaultContractTest.cs
@@ -49,6 +49,7 @@ namespace BasicHttp
         }
 
         [Fact]
+        [UseCulture("en-US")]
         public async Task BasicScenarioServiceMessageParameterWithHttpClient()
         {
             var client = _factory.CreateClient();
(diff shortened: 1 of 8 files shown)


g7ed6e:
It seems that starting from netcoreapp3.1 on linux only, the serialization of the faultstring does not output the `xml:lang` attribute.
```
  Error Message:
   Assert.Equal() Failure
                                 ↓ (pos 124)
Expected: ···ultcode><faultstring xml:lang="en-US">The creator of this fau···
Actual:   ···ultcode><faultstring>The creator of this fault did not specif···
                                 ↑ (pos 124)
  Stack Trace:
``` 

see https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=60058&view=logs&j=ce7f1b94-9c36-52ca-aaa0-6a89018eeb60&t=dc46acd6-af92-576a-ea1c-f197aebd2dbc&l=963

g7ed6e:
@JohnLeyva it should fix #472 can you please have a look?

src/CoreWCF.Http/tests/ServiceWithCoreWCFFaultContractTest.cs
```diff
@@ -0,0 +1,125 @@
...
+                +"<SSMCompatibilityFault xmlns=\"https://ssm-fault-contract-compatibility.com\" xmlns:a=\"http://schemas.datacontract.org/2004/07/Services\" xmlns:i=\"http://www.w3.org/2001/XMLSchema-instance\">"
+                +"<a:Message>An error occured</a:Message></SSMCompatibilityFault></detail></s:Fault></s:Body></s:Envelope>";
+
+            Assert.Equal(expected, responseBody);
+
+            string GetXmlLangAttributeOrNot()
+            {
+                const string enUsXmlLang = " xml:lang=\"en-US\"";
+#if NETCOREAPP3_1_OR_GREATER
+                return RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ?
```
mconnew, line 98: Did you find the response is different on Linux than on Windows? I wouldn't expect this to be the case. It might be a bug in the runtime. Can you open an issue for this and I can look into it at a later time.

g7ed6e: I opened #490

mconnew: Just noticed your earlier comment. I tracked down the problem, and it's caused by CultureInfo.CurrentCulture.Name returning String.Empty on Linux when we initialize the XmlLang property of FaultReasonText. This test is going to be flakey on non en-US systems and will fail for some people running tests locally. Instead can I suggest you fetch CultureInfo.CurrentCulture.Name and if it's length is greater then 0, then you add the xml:lang attribute with it's value? This will accommodate the different Linux behavior as well as make this test work for everyone.

g7ed6e: I’m running fr-FR culture thus I forced the culture to be en-US with `[UseCulture]` attribute on test method to make the culture name invariant.

g7ed6e:
I just added more tests to ensure `Fault`'s namespace is also pulled from `ServiceContract` attribute

g7ed6e:
This bug prevent user to share Contract using FaultException<T> namespace between client and server side code using a dedicated Contract assembly. It would be great to get it in next release.

ping @JohnLeyva @mconnew

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomAttributeProvider.cs
```diff
@@ -295,6 +306,38 @@ private static XmlSerializerFormatAttribute ConvertFromServiceModelXmlSerializer
...
 
+        public static FaultContractAttribute ConvertFromServiceModelFaultContractAttribute(object attr)
+        {
+            Fx.Assert(attr.GetType().FullName.Equals(ServiceReflector.SMFaultContractAttributeFullName), "Expected attribute of type S.SM.FaultContractAttribute");
+            bool hasProtectionLevel = GetProperty<bool>(attr, "HasProtectionLevel");
+            if (hasProtectionLevel)
+            {
+                // ProtectionLevel isn't supported yet so if it was set on the S.SM.SCA, then we can't do the mapping so throw
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new PlatformNotSupportedException("System.ServiceModel.OperationContractAttribute.ProtectionLevel"));
+            }
```
mconnew, line 317: I've seen contracts set ProtectionLevel but set it to None. Some people like to explicitly set everything. In that case, it would be safe to do the mapping. Could you modify the logic here to only throw if it was set to anything other than None?

g7ed6e: The property HasProtectionLevel does not exist in CoreWCF attributes it only exist in SSM ones.
I reproduced the logic that already exist to convert other SSM attributes. please see:
- https://github.com/CoreWCF/CoreWCF/blob/8a15b417fe2f6fedafd88d1c210fb64dad99ae31/src/CoreWCF.Primitives/src/CoreWCF/Description/CustomAttributeProvider.cs#L165
- https://github.com/CoreWCF/CoreWCF/blob/8a15b417fe2f6fedafd88d1c210fb64dad99ae31/src/CoreWCF.Primitives/src/CoreWCF/Description/CustomAttributeProvider.cs#L216
- https://github.com/CoreWCF/CoreWCF/blob/8a15b417fe2f6fedafd88d1c210fb64dad99ae31/src/CoreWCF.Primitives/src/CoreWCF/Description/CustomAttributeProvider.cs#L305

May I open another issue to support explicit `HasProtectionLevel.None` conversion for all SSM attributes in a separate pull request ? it will also cover adding appropriate unit tests.

mconnew: Consistency is more important in this case I believe so I agree with your suggestion. Please open an issue on this and I'll merge this PR.

PR #493, opened 2021-11-04
Add SourceGenerator to inject OperationContract parameters from DI

### LINKED ISSUE #480: Introduce dependency injection of services from a custom parameter attribute in Service implementation
Introduce a parameter attribute allowing to inject services from DI directly in Service implementation similar to  ASP.NET Core `[FromServices]` attribute.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.fromservicesattribute?view=aspnetcore-5.0

```c#
[ServiceContract]
public interface IService
{
    [OperationContract]
    Task<Customer> GetCustomerAsync(int customerId, [Injected]DbContext dbContext);
}
```

This would mean the customer ID comes in as part of the service call, and CoreWCF will fetch from the InstanceContext scoped ServiceProvider the dbContext and add it as a parameter. What are your thoughts?

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/issues/473#issuecomment-949130495_

g7ed6e:
Resolves #480


src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
```diff
@@ -0,0 +1,235 @@
...
+                 : "            ";
+                var dependenciesParameters = Enumerable.Range(0, dependencies.Length).Select(x => $"d{x}");
+                builder.Append($"{prefix}return {operationContract.Name}({string.Join(", ", operationContract.Parameters.Select(x => x.Name).Union(dependenciesParameters))});");
+            }
+        }
+
+        private static IEnumerable<IMethodSymbol> FindOperationContractImplementationCandidate(GeneratorExecutionContext context, INamedTypeSymbol service, IMethodSymbol operationContract)
+        {
+            var CoreWCFInjectedAttribute = context.Compilation.GetTypeByMetadataName("CoreWCF.InjectedAttribute");
+            // Find a candidate implementation with same name as contract
```
mconnew, line 164: I liked the idea of using the ASP.NET Core FromServices attribute. But checking on that class, it involves pulling in the Mvc.Core assembly. Do you have any thoughts about the pros/cons? Should we have a similarly named attribute in a different namespace? Or should it be different name to prevent confusion/namespace ambiguity?

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
```diff
@@ -0,0 +1,235 @@
...
+                string serviceProviderName = isInstanceContextModeSingle ?
+                    "scope"
+                    : "serviceProvider";
+                string prefix =  isInstanceContextModeSingle ?
+                    "                "
+                    : "            ";
+                for (int i = 0; i < dependencies.Length; i++)
+                {
+                    builder.AppendLine($@"{prefix}var d{i} = ({dependencies[i].Type.ContainingNamespace}.{dependencies[i].Type.Name}){serviceProviderName}.GetService(typeof({dependencies[i].Type.ContainingNamespace}.{dependencies[i].Type.Name}));");
+                }
```
mconnew, line 148: Is there a reason to use `GetService(typeof(typename))` instead of `GetService<typename>();`?

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
```diff
@@ -0,0 +1,235 @@
...
+                => operationContract.ReturnsVoid ?
+                    "void"
+                    : $"{operationContract.ReturnType.ContainingNamespace}.{operationContract.ReturnType.Name}";
+
+            string GetParameters()
+                => string.Join(", ", operationContract.Parameters
+                    .Select(p => $"{p.Type.ContainingNamespace}.{p.Type.Name} {p.Name}"));
+            
+            void AddDependenciesResolution()
+            {
```
mconnew, line 138: I took a look at the aspnetcore implementation around this and discovered they support a few other things which aren't available from the IServiceProvider but that we might want to add support for. The code is [here](https://github.com/dotnet/aspnetcore/blob/f58f9dcf1c10345a4fa44bbeea4777efd3b0d407/src/Mvc/Mvc.ApiExplorer/src/EndpointMetadataApiDescriptionProvider.cs#L242) and they have additional support for `HttpContext`, `HttpRequest`, `HttpResponse`, `ClaimsPrincipal`, and `CancellationToken`. While I think it's premature to do all of these (and would be feature creep), I think we can easily support HttpContext, HttpRequest and HttpResponse. The HttpContext is available as an incoming message property on the OperationContext with the key name (I think) "Microsoft.AspNetCore.Http.HttpContext". This then has the Request and Response properties to get the other two. I think it's worth adding those three. What do you think? If you want to do that as a followup item, that's fine too. I'm generally adverse to increasing scope as it makes things take longer to complete.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
```diff
@@ -0,0 +1,235 @@
...
+                                isErrorDiagnosticIssued = true;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        private static bool GetInstanceContextMode(GeneratorExecutionContext context, INamedTypeSymbol service)
+        {
```
mconnew, line 56: This doesn't get the InstanceContextMode, it just returns if it's a Single. I'm a little concerned that this is fragile as there are other ways to se the instance context mode. You can add a configure service delegate to your app which gets the ServiceBehaviorAttribute from the Description and modifies it. There is an internal property on InstanceContext called IsSingleton that we could make public. Currently it won't return the correct answer as it bases the answer on whether the IInstanceContextProvider is of type SingletonInstanceContextProvider and that's not the case any more. We could easily make that just an auto property which is set by ServiceBehavior.ApplyInstancing. Then you would just check InstanceContext.IsSingleton to decide whether to use a scope. If IsSingleton has a public setter too, then that would allow custom IInstanceContextProvider types to also set it and control this behavior.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
g7ed6e: I was trying to avoid output `using` directives in generated code. I’m now using the `GetService<T>()` extension method as I had to add a `using Microsoft.Extensions.DependencyInjection;` to call `CreateScope()`

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
g7ed6e: Indeed let’s build a skateboard before a car. I will create an issue to trace adding the support of ˋHttpContext` and related services.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
g7ed6e: I decided to introduce the `InjectedAttribute` for the the reasons you mentioned. First to avoid pulling the `Microsoft.AspNetCore.Mvc` package then to prevent confusion. Let me know if you think we should align the naming of the attribute with the one from ASP.NET Core. A good reason would be foreseeing a future `FrameworkDependency` to Microsoft.AspNetCore.App allowing to add compat code between CoreWCF and ASP.NET Core versions of `[FromServices]`. I do not know what your plan is taking a dependency on this for CoreWCF 2.0.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
g7ed6e: Thus i understand `InstanceContextMode` detection can only occur at runtime. I will adapt the generated code and expose the `IsSingleton` property as you suggested.

g7ed6e: > There is an internal property on InstanceContext called IsSingleton that we could make public. Currently it won't return the correct answer as it bases the answer on whether the IInstanceContextProvider is of type SingletonInstanceContextProvider and that's not the case any more. 

@mconnew I checked the code and i don't get why the `IsSingleton` is no longer returning the correct answer. So far in #473 we dit not introduce new `IInstanceContextProvider` implementation,
For `PerCall`/`PerSession` we adpated the `IInstanceProvider` implementation to pull `Service `instances from DI and expose a scoped `IServiceProvider` through `InstanceContext.Extensions`.
For `Single` we adapted the `ServiceBehaviorAttribute` to attach the `IServiceProvider` to the current `InstanceContext.Extensions`.
In both cases we did not alter `InstanceContext` instance lifetime,

g7ed6e: I exposed `IsSingleton` in #494

mconnew: I merged #494 so you can rebase and use that.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.cs
```diff
@@ -0,0 +1,308 @@
...
+            return false;
+        }
+
+        private static void GenerateOperationContractImplementation(GeneratorExecutionContext context, INamedTypeSymbol service, INamedTypeSymbol contract, IMethodSymbol operationContractImplementation, IMethodSymbol operationContract)
+        {
+            var SSMServiceBehaviorSymbol = context.Compilation.GetTypeByMetadataName("System.ServiceModel.ServiceBehaviorAttribute");
+            var CoreWCFServiceBehaviorSymbol = context.Compilation.GetTypeByMetadataName("CoreWCF.ServiceBehaviorAttribute");
+            var TaskSymbol = context.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task");
+            var GenericTaskSymbol = context.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task<object>");
+
```
mconnew, line 89: This doesn't look to be used anywhere. But should it be `Task<>` for an open generic? It looks like you worked around this by looking at the base type of the return type to see if it's Task. Normally this would be fragile (as Task isn't sealed), but we're constrained to what CoreWCF supports and it only supports Task and Task\<TResult\> task based return types.

g7ed6e: I finally found the correct syntax which actually is 
```csharp
context.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1")
```

g7ed6e: and use this symbol to compare with
```
ReturnType.ConstructedFrom
```

g7ed6e:
@mconnew I think that now an acceptable part of use cases are supported (please see the unit tests).

I packed the generator so it is distributed with `CoreWCF.Primitives` nuget package, I tested it locally and it works great.. in VS2022.
Regarding VS2019, as discussed previously I confirm .NET6 `IIncrementalGenerator` is not supported. Please see https://github.com/dotnet/sdk/issues/20355. However, the effort to add VS2019 support through an `ISourceGenerator` is low. The required steps are:
- Add one class that implementing .NET5 `ISourceGenerator` interface and map the current logic to the `ISourceGeneratir.Execute` method.
- Duplicate the `CoreWCF.BuildTools` csproj so we can output 2 separate assemblies (one for each roslyn version) and place them in version specific roslyn folders within the `CoreWCF.Primitives` nuget package.

Please see how STJ is handling it here https://github.com/dotnet/runtime/tree/main/src/libraries/System.Text.Json/gen. 

Regarding unit tests, as i first wrote the code agains the .NET5 API, I used the recommended _solution A_ as described here  https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#unit-testing-of-generators but when i switched to the .NET6 API i noticed that the current `Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit` does not support the `IIncrementalGenerator` yet. Thus I had to duplicate the below two files to switch a generic constraint from `ISourceGenerator` to `IIncrementalGenerator`.
https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit/SourceGeneratorVerifier%601.cs
https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing/CSharpSourceGeneratorTest%602.cs

mconnew:
Thanks, I meant to come back and say as it sounds like it won't be a lot of ongoing work to support 2019, that it would be good to support it. I've had no internet for the last 10 days due to a car crashing into the junction box in my street, so I have a bit of catching up to do. I hope to get this reviewed tomorrow or Wednesday and get this merged this week.

g7ed6e:
Support for vs2019 and vs2022 is already implemented as the effort to support both was quite low and no code duplication was required. I reproduced the exact same logic as System.Text.Json gene rator to handle both support. Unit tests are also ran with the two versions of roslyn.

src/CoreWCF.BuildTools/src/DiagnosticDescriptors.cs
```diff
@@ -0,0 +1,37 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.CodeAnalysis;
+
+namespace CoreWCF.BuildTools
+{
+    internal static class DiagnosticDescriptors
+    {
+        private static readonly DiagnosticDescriptor parentClassShouldBePartialError = new DiagnosticDescriptor(id: "COREWCF_0100",
+                                                                              title: "Parent class should be partial",
```
mconnew, line 11: Naming convention of private static fields is for them to start with s_

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -0,0 +1,216 @@
...
+                    .Select(p => p.RefKind switch
+                    {
+                        RefKind.Ref => $"ref {p.Type} {p.Name}",
+                        RefKind.Out => $"out {p.Type} {p.Name}",
+                        _ => $"{p.Type} {p.Name}",
+                    }));
+
+                var indentor = new Indentor();
+                var builder = new StringBuilder();
+
```
mconnew, line 116: Is this method called concurrently for multiple instances? If not, then it might be worth caching this StringBuilder. So we're not reallocating every time. It has a Clear() method to reset it.

g7ed6e: Strictly in this PR code there is no concurrent access. However I think it could be called concurrently, it depends of the roslyn infrastructure integration within IDE I guess. I don't know if a mutex is set.
To be pragmatic I checked the Emitter implementation of STJ generator and nothing is done there to cache the StringBuilder. Please see there https://github.com/dotnet/runtime/blob/0f671237e9b148a371ffa41062a11a2e0f7b3e7d/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs#L149

src/CoreWCF.BuildTools/src/DiagnosticDescriptors.cs
g7ed6e: Fixed with this commit https://github.com/CoreWCF/CoreWCF/pull/493/commits/62f7b20cd46341b794c9b7f3fdf92668a379a1c8

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
mconnew: Okay, seems reasonable to leave it as there's ambiguity whether that's safe or not.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -0,0 +1,216 @@
...
+            public Emitter(in OperationParameterInjectionSourceGenerationContext sourceGenerationContext, SourceGenerationSpec generationSpec)
+            {
+                _sourceGenerationContext = sourceGenerationContext;
+                _generationSpec = generationSpec;
+            }
+
+            public void Emit()
+            {
+                foreach (var operationContractSpec in _generationSpec.OperationContractSpecs)
+                {
```
mconnew, line 65: You've enabled Nullable C# feature in the csproj (which is great BTW), but there are still a small number of nullability issues in the code. For example, there are 5 instances of "Warning CS8602 Dereference of a possibly null reference." I know there are a lot of warnings throughout the codebase that need to be addressed, but I plan to get it to a clean build where we can turn on treat warnings as errors.

mconnew:
I'm seeing a warning for each of the new rules which all look similar to this:
```
Warning RS2008 Enable analyzer release tracking for the analyzer project containing rule 'COREWCF_0100'
```
The RS2008 rule link takes you to [this](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) page. Can you add the release tracking metadata files? Once done, I'll update the release instructions to include moving the unreleased rules sections into the release instructions.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
g7ed6e: Done here https://github.com/CoreWCF/CoreWCF/pull/493/commits/c150ef935aba68d117b05cdf2daefd08e5bb1f48.

g7ed6e:
> I'm seeing a warning for each of the new rules which all look similar to this:
> 
> ```
> Warning RS2008 Enable analyzer release tracking for the analyzer project containing rule 'COREWCF_0100'
> ```
> 
> The RS2008 rule link takes you to [this](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) page. Can you add the release tracking metadata files? Once done, I'll update the release instructions to include moving the unreleased rules sections into the release instructions.

g7ed6e:
> I'm seeing a warning for each of the new rules which all look similar to this:
> 
> ```
> Warning RS2008 Enable analyzer release tracking for the analyzer project containing rule 'COREWCF_0100'
> ```
> 
> The RS2008 rule link takes you to [this](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) page. Can you add the release tracking metadata files? Once done, I'll update the release instructions to include moving the unreleased rules sections into the release instructions.

done here https://github.com/CoreWCF/CoreWCF/pull/493/commits/cd9cb227e63e71496701b14e4fc26d1518b59bc4

mconnew:
I'm going to give this a try locally on my machine and presuming no issue come up, I'll merge it.

mconnew:
There's a small problem on the error report for COREWCF_0100. It doesn't populate the line number and filename of the class which needs to be made partial.
image
It says the file is CSC and line number is 1.

g7ed6e:
Indeed that's because the diagnostic is issued at build time by the source generator itself and not by a separate roslyn analyzer I think.

Issue #477, opened 2021-10-19
Addressing Version AddressingNone is not supported


JohnLeyva:
Addressing Version AddressingNone is explicitly not supported at the moment  by CoreWCF.

``` c# 
        public static CustomBinding CreateCustomBinding()
        {
            CustomBinding result = new();
            TextMessageEncodingBindingElement textBindingElement = new()
            {
            //System.ArgumentException : Addressing Version 'AddressingNone (http://schemas.microsoft.com/ws/2005/05/addressing/none)' is not supported. (Parameter 'addressingVersion')
                MessageVersion = MessageVersion.CreateVersion(EnvelopeVersion.Soap12, AddressingVersion.None)
            };
            result.Elements.Add(textBindingElement);
            HttpTransportBindingElement httpBindingElement = new()
            {
                //AllowCookies = true,
                MaxBufferSize = int.MaxValue,
                MaxReceivedMessageSize = int.MaxValue
            };
            result.Elements.Add(httpBindingElement);
            return result;
        }
```

PR #495, opened 2021-11-10
Add Support to Addressing Version AddressingNone 

JohnLeyva:
Add Support to 

Addressing Version AddressingNone is not supported #477

Please just look at the latest commit, if that seems as a acceptable approach I'll prune all the changes that are not related and also create the corresponding Test


mconnew:
There's a bunch of combinations missing. I hadn't realized we had so much missing. This is a hang over from pulling over code from the WCF Core client where initially we had a very limited set of options. We need tests for each of these for both NetTcp and BasicHttp as there might be some transport level code missing for some of them (e.g. HTTP places the Action header into HTTP headers depending on the addressing version). I'm going to do a quick cleanup with tests for all the missing variants so don't worry about this PR. Would you be able to review my PR when it's done?

mconnew:
@JohnLeyva, would you have time to review #500?

Issue #497, opened 2021-11-15
Make the ServiceSecurityContext.GetIdentities method public


JeffLeBert:
In WCF, this method is public.

For the product I work on, we use this method in our own AuthorizationManager. For now, I've copied the code from the private method into our own project as all the information it uses is public. The code duplication is less than optimal.

This looks like a pretty simple thing to fix, write unit tests for and submit a pull request, so I will give that a shot.

PR #498, opened 2021-11-15
Make the ServiceSecurityContext.GetIdentities method public

JeffLeBert:
- Made GetIdentities method public.
- Added unit tests to cover newly exposed functionality.

#497


mconnew:
Thank you. We have a few new changes available now so I'll try and get a new public release out next week when I have some down time.

Issue #496, opened 2021-11-15
Make the DispatchOperation.CallContextInitializers property public


JeffLeBert:
In WCF, this property is public. This will also force the `ICallContextInitializer` interface to be made public.

For the product I work on, we use the `ICallContextInitializer` `BeforeInvoke` and `AfterInvoke` methods to configure call specific information for each call.

I've already tested this in our product and it works. I will try to write unit tests for where this is used in the `DispatchOperationRuntime` and submit a pull request.

JeffLeBert:
Advice needed... I'm trying to write a unit test to verify that the `BeforeInvoke` and `AfterInvoke` interface methods are called. I was unable to find any public method I could call that would hit these methods without going through a significant amount to other code first. This would make the test very difficult to write and very fragile.

When I come up against this, I commonly make the method I'm trying to test internal and make sure I have an `InternalsVisibleTo(...)` attribute pointing to my unit test project. This allows me to call the method directly from the unit test without making it public.

I don't see anything like this in the current suite of unit tests so I'm pretty sure I shouldn't add it. Any help would be greatly appreciated.

JeffLeBert:
I decided to just submit a pull request without unit tests.

PR #499, opened 2021-11-16
Make the DispatchOperation.CallContextInitializers property public

JeffLeBert:
- Made DispatchOperation.CallContextInitializers public.
- Made ICallContextInitializer public as it is required by above.

#496


JeffLeBert:
I was not able to create any unit tests for this simple change. The methods I would have tested are private and deep in a call stack that is internal or private. Any unit test I would make from this would by difficult and fragile. I would be happy to do this if you can tell me how...

I have tested this code inside my companies application and it worked as expected.

mconnew:
The way you would test this is to create an instance in your test which records how many times each of the methods are called and ensure it records the correct count. One way to do this is to have a static property on the startup class which holds a reference to the instance. Create a service behavior which configures the service to use this instance and add an instance of the behavior to DI in the ConfigureServices method. Then in the actual test, you make a service call and validate that the call context initializer had each of it's methods called once. Make a second call and ensure the counters go up by 1 again. 

An additional test is to return an object in the before method, and make sure that it's passed in as the correlation object to the after call and record that it was (eg boolean property on the instance). Then in the unit test check that the boolean has been set indicating that the correlation object was correctly passed. 

Finally, do some basic null checks on the parameters to make sure the InstanceContext and Message objects are passed in.

We already have some extensibility tests which do a similar thing so if this isn't making sense to you, let me know and I can point you to them as an example.

JeffLeBert:
Thanks! I didn't think an "integration" test was the best path to test, but we probably would need one even if we had a "unit" test just testing the specific sections of code. I will try to find a good example, if you don't give me one.

I will try to get back to this as soon as I can.

mconnew:
Take a look at src\CoreWCF.Primitives\tests\Extensibility\MessageInspectorTests.cs as an example of how to do this. It's a very similar scenario of a method called on the request and on the reply. This is actually closer to a unit test as it uses a bunch of helpers which don't require a real transport. I created a WCF client binding which dispatches to a CoreWCF dispatch pipeline with no actual real transport in between. A tel for CallContextInitializers should go in the Extensibility folder that this test is in. Feel free to ask if you have any questions.

JeffLeBert:
Perfect! I will look at this when I get a chance. Other projects have me pretty busy right now so I don't know when I will get back to this.

cyanite:
Here's hoping you get around to it... so I can remove my horrible reflection hack to add these inspectors ;)

JeffLeBert:
I don't think in "github" so I think I've gotten the three files in the right state for this pull request to be completed. The CoreWCF version bumped from 0.3 to 0.4 between my changes so I have no idea what could go wrong. :-)

The third file should be the unit tests associated with this change. That should be the last bit needed for this pull request.

Thanks @mconnew for your help. I was able to find a good set of unit tests to help me write ones specific to my changes.

src/CoreWCF.Primitives/tests/Extensibility/CallContextInitializersTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        public void AfterInvoke(object correlationState)
+        {
+            AfterInvokeCount++;
+        }
+
+        public object BeforeInvoke(InstanceContext instanceContext, IClientChannel channel, Message message)
+        {
+            BeforeInvokeCount++;
+            return new object();
+        }
```
mconnew, line 111: Can you add a test to ensure that the returned correlation object is passed into AfterInvoke?

JeffLeBert:
It looks like the build failed in CoreWCF.NetTcp\tests\CoreWCF.NetTcp.Tests.csproj. Since I only added a unit test in CoreWCF.Primitives.Tests.csproj, I'm guessing I didn't cause that. Not sure if I need to do anything or not...

Thanks for pointing out the missing test case. Happy to help and make things just a tiny bit better. :-)

mconnew:
The failure is caused by a test which hits some race condition that I have been unable to track down so far. I've kicked off the tests again and it should be good now. After it passes, I'll merge.

PR #500, opened 2021-11-17
Add support for all values of MessageVersion

mconnew:
Replaces #495 
This adds support for all MessageVersion variations that exist in WCF, with tests to validate them (Excluding None which is for special purposes which CoreWCF doesn't support yet).

### DIFF
--- a/src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
+++ b/src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
@@ -127,6 +127,7 @@ namespace Helpers
                     });
                 })
             .UseStartup<TStartup>();
+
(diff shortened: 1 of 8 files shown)


JohnLeyva:
Tested it and review the payload. It works 💯

JohnLeyva:
Build is Failing, 

Looks like the following fixes it
https://github.com/CoreWCF/CoreWCF/pull/495/commits/69648b08a9a8dbd7a5d20373eeb93b0115fb3c80

This is currently part of
https://github.com/CoreWCF/CoreWCF/pull/481

mconnew:
The failure is due to a test for a failure case which there's a race condition for which I haven't been able to repro locally. It causes the service to kill the connection instead of sending a fault message and then closing the connection. But it only reproes on Linux, and only one a super slow machine (build machines with high CPU contention). Those code changes aren't needed to get this to pass, I'm rerunning the test and it should pass.

Issue #489, opened 2021-10-30
Display SonarCloud badge in readme 


g7ed6e:
The SonarCloud analysis is not available.

https://sonarcloud.io/project/overview?id=CoreWCF_CoreWCF

edit: it would be great to get code coverage metrics to measure test coverage progress and spot potential weakness

PR #502, opened 2021-11-18
Fix SonarCloud integration & Code coverage reports

g7ed6e:


g7ed6e:
Resolves #489 

@mconnew :
- CodeCoverage report is fixed and published in a tab of the build in azure devops.
- SonarCloud service endpoint is only available to build ran from a CoreWCF repo branch. Thus I cannot trigger it from PR. But I think it should now work.

Issue #504, opened 2021-11-28
GetKnownTypes for dynamic known type list


absinghal:
WCF in .Net framework had an overload for ServiceKnownType as follows:

```
[ServiceKnownType("GetKnownTypes", typeof(DataProviderTypes))]
public interface IService
{
}

public class DataProviderTypes
{
    public static IEnumerable<Type> GetKnownTypes(ICustomAttributeProvider provider)
    {
        List<Type> lst = new List<Type>();
        '
        '
        return lst;
    }
}
```

This allowed to dynamically generate list of known types.

Is there any way to achieve this ? If not, when is this planned ? 
Coz this is looking like a showstopper.

Thanks

absinghal:
In the source at [this](https://github.com/CoreWCF/CoreWCF/blob/5408657365fe03de44dc236479bfe0b215871aad/src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs#L382) location, following remark is given and relevant code is commented:

"// The named method must take a parameter of ICustomAttributeProvider which isn't available so this can only specify known types by Type"

ICustomAttributeProvider is [available](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.icustomattributeprovider?view=net-6.0) in .Net Standard 2.0, so why cant we just uncomment this code and enable this functionality.

@mconnew will you accept a pull request as per above ?

mconnew:
That comment is a holdover from very early .NET Standard days when that interface wasn't available. It wasn't the reason why that overload wasn't supported though. It was because this code was originally ported for UWP apps running on .NET Native which didn't allow reflection and dynamically calling of methods. This can definitely be re-enable. I'll take a look later to see how much work this will be.

mconnew:
I took a look and here's the steps that need to be done to enable this.
1. Uncomment the disabled properties in ServiceKnownTypeAttribute and fix (just turn into getter only auto properties)
2. Modify CustomAttributeProvider.GetCustomAttributes to also know about ServiceKnownTypeAttribute. This is to provide compat with System.ServiceModel namespaced version of attribute. This will require a new method ConvertFromServiceModelServiceKnownTypeAttribute to be created. It should be easy enough to look at the existing methods to copy the pattern.
3. Uncomment code at top of TypeLoader\<TService\>.GetKnownTypes which handles this case. I believe you can change the second parameter to ICustomAttributeProvider as the abilities of CustomAttributeProvider aren't needed in this method.
[...]

PR #516, opened 2021-12-11
 GetKnownTypes for dynamic known type list #504 

absinghal:


src/CoreWCF.Primitives/src/CoreWCF/Description/CustomAttributeProvider.cs
```diff
@@ -372,7 +318,17 @@ private static OperationContractAttribute ConvertFromServiceModelOperationContra
...
+        {
+            Fx.Assert(attr.GetType().FullName.Equals(ServiceReflector.SMServiceKnownTypeAttributeFullName), "Expected attribute of type S.SM.ServiceKnownType");
+            var skta = new ServiceKnownTypeAttribute();
+
+            skta.MethodName= GetProperty<string>(attr, nameof(ServiceKnownTypeAttribute.MethodName));
+            skta.Type = GetProperty<Type>(attr, nameof(ServiceKnownTypeAttribute.Type));
+            skta.DeclaringType = GetProperty<Type>(attr, nameof(ServiceKnownTypeAttribute.DeclaringType));
 
+            return skta;
+        }
```
mconnew, line 331: Missing blank line between methods.

absinghal: Done

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            var factory = ServiceKnownTypeTests.CreateFactory<IPingService4, PingService4>();
+            var channel = ServiceKnownTypeTests.CreateChannel(factory);
+            BaseMsg msg = channel.Ping("hello");
+            Assert.Equal("hello", msg.msg);
+            ServiceKnownTypeTests.CloseChannel(factory, channel);
+        }
+    }
+
+    internal class PingServiceEx : IPingServiceEx
+    {
```
mconnew, line 96: There's a lot of redundancy in these tests. If you declare a base `PingServiceBase` class like this:
```c#
internal class PingServiceBase
{
    public BaseMsg Ping(string msg) => new DerivedMsg { msg = msg };
}
```
Then each of the implementation classes like `PingServiceEx` just need to derive from this class and declare the service interface they are implementing. For example, `PingServiceEx` now becomes this simple one liner:
```c#
internal class PingServiceEx : PingServiceBase, IPingServiceEx { }
```

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            Assert.Throws<System.Xml.XmlException>(() =>
+            {
+                BaseMsg msg = channel.Ping("hello");
+            });
+            ServiceKnownTypeTests.CloseChannel(factory, channel);
+        }
+
+        [Fact]
+        public static void ServiceKnownTypeTest1()
+        {
```
mconnew, line 56: These test names are a bit obscure. Can you rename them to have a bit of description. E.g. this one could be `ServiceKnownTypeInlineType` as the type is inline to the attribute. Maybe add a comment at the start with the `ServiceKnownType` usage. So something like this:
```c#
        public static void ServiceKnownTypeInlineType()
        {
            // Testing [ServiceKnownType(typeof(DerivedMsg))]
            /* Rest of test */
        }
```

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+        public BaseMsg Ping(string msg)
+        {
+            return new DerivedMsg { msg = msg };
+        }
+
+    }
+
+    [ServiceContract]
+    [System.ServiceModel.ServiceContract]
+    public interface IPingServiceEx
```
mconnew, line 106: The set of attributes is a bit overwhelming to be able to quickly look over at a glance. We already have tests to validate `System.ServiceModel.ServiceContract` vs `CoreWCF.ServiceContract` (and the same for `OperationContract`) so we can assume you can just use `System.ServiceModel.ServiceContract` as the compatibility just works and leave out the CoreWCF namespaced one (same for OperationContract).

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+    [System.ServiceModel.ServiceContract]
+    [ServiceKnownType(typeof(DerivedMsg))]
+    [System.ServiceModel.ServiceKnownType(typeof(DerivedMsg))]
+    public interface IPingService1
+    {
+        [OperationContract]
+        [System.ServiceModel.OperationContract]
+        BaseMsg Ping(string msg);
+
+    }
```
mconnew, line 133: There are extra blank lines before the close of the interface and the implementing classes (which should be refactored to one liners anyway).

mconnew:
You missed a test for the variation `ServiceKnownTypeAttribute(string methodName)`. This is a bit trickier to test as it's used in a really uncommon scenario. You don't actually need a contract interface to use WCF/CoreWCF. You can apply the `ServiceContract` and `ServiceOperation` attributes directly to the implementation class. When you do this, you can create a static method to provide the known types and only supply the method name to `ServiceKnownTypeAttribute`. This requires some changes to DispatcherHelper to create an overload/alternative method to wire everything up as the service type needs to be used in place of the contract type when building the dispatcher, but the contract type must still be used for the ChannelFactory. You also need to massage the action's on the operation to get things to match.  
 
If you want to add a test for this scenario you can, but as this is a really obscure feature and we have no other tests testing this way of using CoreWCF, it's okay to exclude that scenario. I did do a quick hack prototype of this scenario based on your existing tests and I confirmed it works.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+    public class DerivedMsg2 : BaseMsg
+    {
+    }
+
+    public class DataProviderTypes
+    {
+        public static IEnumerable<Type> GetKnownTypes(ICustomAttributeProvider provider)
+        {
+            List<Type> lst = new List<Type>();
+            lst.Add(typeof(DerivedMsg));
```
mconnew, line 218: Array's implement IEnumerable so you could do `return new Type[] { typeof(DerivedMsg) };`. Do feel obligated to make a change though, this is more of an FYI.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    [DataContract]
+    public class DerivedMsg : BaseMsg
+    {
+    }
+
+    [DataContract]
+    public class DerivedMsg2 : BaseMsg
+    {
```
mconnew, line 210: DerivedMsg2 is never used

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: Good suggestion. Implemented.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: Done

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: Done

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: Should look Ok after refactoring

absinghal:
> You missed a test for the variation `ServiceKnownTypeAttribute(string methodName)`. This is a bit trickier to test as it's used in a really uncommon scenario. You don't actually need a contract interface to use WCF/CoreWCF. You can apply the `ServiceContract` and `ServiceOperation` attributes directly to the implementation class. When you do this, you can create a static method to provide the known types and only supply the method name to `ServiceKnownTypeAttribute`. This requires some changes to DispatcherHelper to create an overload/alternative method to wire everything up as the service type needs to be used in place of the contract type when building the dispatcher, but the contract type must still be used for the ChannelFactory. You also need to massage the action's on the operation to get things to match.
> 
> If you want to add a test for this scenario you can, but as this is a really obscure feature and we have no other tests testing this way of using CoreWCF, it's okay to exclude that scenario. I did do a quick hack prototype of this scenario based on your existing tests and I confirmed it works.

Yes, left out intentionally for above reasons. Good that you could confirm its working. Yes we can exclude for now.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: Leftover from experimentation. Removed now.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
absinghal: I tend to prefer List<T>, specially when the number of elements is too small to worry about any performance hit.

src/CoreWCF.Primitives/tests/ServiceKnownTypeTests.cs
```diff
@@ -0,0 +1,187 @@
...
+
+    internal class PingService_Inline : PingServiceBase, IPingService_Inline { }
+   
+    [System.ServiceModel.ServiceContract]
+    [ServiceKnownType(typeof(DerivedMsg))]
+    [System.ServiceModel.ServiceKnownType(typeof(DerivedMsg))]
+    public interface IPingService_Inline
+    {
+       
+        [System.ServiceModel.OperationContract]
```
mconnew, line 114: There's still some issues with extra blank lines in the interfaces and in PingServiceBase. Once those are fixed, this looks good to merge.

absinghal: Done as per above

PR #523, opened 2021-12-18
Add support to inject HttpContext related services from operation parameters

### LINKED ISSUE #507: Add support to inject ASP.NET Core implementations in OperationContractParameterInjectionGenerator
I took a look at the aspnetcore implementation around this and discovered they support a few other things which aren't available from the IServiceProvider but that we might want to add support for. The code is [here](https://github.com/dotnet/aspnetcore/blob/f58f9dcf1c10345a4fa44bbeea4777efd3b0d407/src/Mvc/Mvc.ApiExplorer/src/EndpointMetadataApiDescriptionProvider.cs#L242) and they have additional support for `HttpContext`, `HttpRequest`, `HttpResponse`, `ClaimsPrincipal`, and `CancellationToken`. While I think it's premature to do all of these (and would be feature creep), I think we can easily support HttpContext, HttpRequest and HttpResponse. The HttpContext is available as an incoming message property on the OperationContext with the key name (I think) "Microsoft.AspNetCore.Http.HttpContext". This then has the Request and Response properties to get the other two. I think it's worth adding those three. What do you think? If you want to do that as a followup item, that's fine too. I'm generally adverse to increasing scope as it makes things take longer to complete.

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/493#discussion_r743156467_

followup item of #493

g7ed6e:
Resolves #507


src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -141,6 +141,15 @@ namespace {operationContractSpec.ServiceContractImplementation.ContainingNamespa
                 indentor.Increment();
                 builder.AppendLine($@"{indentor}var serviceProvider = CoreWCF.OperationContext.Current.InstanceContext.Extensions.Find<IServiceProvider>();");
                 builder.AppendLine($@"{indentor}if (serviceProvider == null) throw new InvalidOperationException(""Missing IServiceProvider in InstanceContext extensions"");");
+
+                if (dependencies.Any(x => SymbolEqualityComparer.Default.Equals(x.Type, operationContractSpec.HttpContextSymbol)
+                    || SymbolEqualityComparer.Default.Equals(x.Type, operationContractSpec.HttpRequestSymbol)
+                    || SymbolEqualityComparer.Default.Equals(x.Type, operationContractSpec.HttpResponseSymbol)))
+                {
+                    builder.AppendLine($@"{indentor}var httpContext = CoreWCF.OperationContext.Current.RequestContext.RequestMessage.Properties[""Microsoft.AspNetCore.Http.HttpContext""] as Microsoft.AspNetCore.Http.HttpContext;");
+                    builder.AppendLine($@"{indentor}if (httpContext == null) throw new InvalidOperationException(""Missing HttpContext in RequestMessage properties"");");
```
mconnew, line 150: The this[string] item accessor for MessageProperties throws ArgumentException if the item isn't found so your exception won't get thrown. Maybe use ContainsKey to throw. Or you could do the check and fetch with TryGetValue. The returned value may still be null if there's a key with that name but the value is null.

g7ed6e: I fixed the generated code in latest commit

mconnew:
Looks good, thanks.

PR #524, opened 2021-12-18
Improve error message when adding ServiceEndpoint to service builder for a service type which hasn't been added

### LINKED ISSUE #437: Improve error message when adding ServiceEndpoint to service builder for a service type which hasn't been added
Hello,
while configuring a NetTcp binding using CoreWCF I get this really not useful exception. I tracked it and it should be thrown by aspnetcore. Since this is pretty much equal to the example you provide, I'm not sure which is the problem, so if you have any idea I would be happy to listen :D. Below I uploaded the startup, the webhost, the exception and even the stacktrace (TConnector is a generic, but in this case it's a class that implements IBqConnector with and empty constructor)
Thank you,
Manuele

image

image

image

image

g7ed6e:
Resolves #437

### DIFF
--- a/src/CoreWCF.Primitives/src/Resources/Strings.resx
+++ b/src/CoreWCF.Primitives/src/Resources/Strings.resx
@@ -8217,4 +8217,7 @@
   <data name="SynchronousUserNamePasswordValidationIsDeprecated" xml:space="preserve">
     <value>Implementers should implement ValidateAsync instead.</value>
   </data>
+  <data name="UnknownServiceConfiguration" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


src/CoreWCF.Primitives/src/Resources/Strings.resx
```diff
@@ -8217,4 +8217,7 @@
   <data name="SynchronousUserNamePasswordValidationIsDeprecated" xml:space="preserve">
     <value>Implementers should implement ValidateAsync instead.</value>
   </data>
+  <data name="UnknownServiceConfiguration" xml:space="preserve">
+    <value>Could not find service '{0}' in configured services. Service must be registered using 'builder.AddService&lt;TService&gt;(...)' before registering a service endpoint using 'builder.AddServiceEndpoint&lt;TService, TContract&gt;(...)'</value>
+  </data>
```
mconnew, line 8222: Can you change the error message to refer to the IServiceBuilder interface instead of a presumed variable name?

g7ed6e: done!

Issue #526, opened 2022-01-04
Exception when exposing multiple contracts with same address (listen URI)


cyanite:
This was possible in old WCF, and it seems to be a bug in `DispatcherBuilder.cs`, in the `InitializeServiceHost` method. We have:

    // Ensure ListenUri and group endpoints per ListenUri
    for (int i = 0; i < description.Endpoints.Count; i++)
    {
        ServiceEndpoint endpoint = description.Endpoints[i];

        ListenUriInfo listenUriInfo = GetListenUriInfoForEndpoint(serviceHost, endpoint);
        if (!stuffPerListenUriInfo.ContainsKey(listenUriInfo))
        {
            stuffPerListenUriInfo.Add(listenUriInfo, new StuffPerListenUriInfo());
        }
        stuffPerListenUriInfo[listenUriInfo].Endpoints.Add(endpoint);
        stuffPerListenUriInfo[listenUriInfo].Parameters.Add(services);
    }

The problem is that `stuffPerListenUriInfo[listenUriInfo].Parameters.Add(services);` is done for each endpoint, but the collection doesn't accept duplicates. Here `services` is the current `IServiceProvider` which is obviously the same for the entire service host.

I suspect this should only be added once per listen URI.

This makes it impossible to listen to multiple contracts on the same URI.

cyanite:
I have confirmed that my PR fixes the problem. My WCF code now runs normally :)

cyanite:
Pull request updated with a test.

PR #527, opened 2022-01-04
Fix multiple endpoints with same address bug (#526)

cyanite:


cyanite:
I can't reproduce the failed test in Rider. I also can't imagine how my change, which removes a crash, would affect this.

mconnew:
Could you add a test which fails without this fix to make sure it doesn't regress? There's some outstanding issues that are likely to result in some refactoring of the code around where you made your change and I wouldn't want to regress this in the future.

cyanite:
All right, I added a regression test. It works fine, but I am not necessarily too happy with it. I am not sure how you would prefer it. This is what I did:

1. I added it in CoreWCF.Primitives/tests since it's unrelated to any particular binding.
2. I can't test the DispatcherBuilder directly, because it's internal, so it's essentially an integration test.
3. Since I need some binding, I added a dependency on the NetTcp project. It has nothing to do with NetTcp, though, but I do need some binding, since the error only happens when the host is started, not just built.
4. As mentioned, I have to start the host. I am not too happy about that, as it's slow and uses a port on the machine etc.
5. I put it in a new file called EndpointTests, since I didn't find any suitable place.

If you prefer it done differently, I think I'm gonna need some pointers :)

src/CoreWCF.Primitives/tests/EndpointTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                });
+            }
+        }
+
+        [Fact]
+        public void MultipleEndpointsWithSameListenAddressShouldWork()
+        {
+            var builder = WebHost.CreateDefaultBuilder<Startup>(null);
+            builder.UseNetTcp(8080);
+            var host = builder.Build();
```
mconnew, line 62: You don't need to do this. If you look at other tests, we pass 0 as the port to let Kestrel pick one.

mconnew:
Once you've fixed the test to pass port 0 I'll merge this fix.

src/CoreWCF.Primitives/tests/EndpointTests.cs
cyanite: Ok. Initially I called the method without arguments, but this picks port 808 or something, which is privileged on Linux, making all the Linux tests fail. I'll take a look at the code, but if 0 picks 808, I don't think it will work.

cyanite:
Thanks. It seems it works :)

Issue #530, opened 2022-01-07
NetTcp binding: NullReferenceException on the server when using wrong end point address on the client


cyanite:
I accidentally used "IFoo" instead of "Foo" in the service URL on the client (something like "net-tcp://localhost/Some.Namespace.IFoo"), and it didn't work, since the endpoint was exposed with "Foo". After debugging, I see that it throws a NullReferenceException on the server. This happens in `NetMessageFramingConnectionHandler.cs`, in the method `BuildHandshake`. It calls `GetServiceHandshakeDelegate` and then immediately calls the returned delegate. Unfortunately,  `GetServiceHandshakeDelegate` simply returns null if no suitable delegate is found, apparently when you use a wrong endpoint.

I feel the framework should never throw a NullReferenceException in any situation. I'd be happy to help with a PR, if possible.

mconnew:
You're right, that is a bug. There's a special endpoint not found sequence the service needs to send back to the client. It's not a SOAP message, it's part of the NetMF Framing protocol. The approach I would take is to have a default handler which does this that GetServiceHandshakeDelegate would return instead of null. If you need help finding the code which sends back that byte sequence, let me know. It should be relatively quick for me to dig out.

cyanite:
I created a pull request. No tests yet, but a take on the implementation. It works for me in local testing; I now get correct faults on the client.

mconnew:
The fix was merged

PR #531, opened 2022-01-08
Fix incorrect endpoint not found handling in NetTcp framing (#530)

cyanite:


cyanite:
I didn't implement it exactly as suggested since we have to not invoke the next middleware delegate in case of a missing endpoint. Other details:

- I tried to somewhat adapt naming and style from other code. Let me know if it should be different.
- I made an internal method private, since it wasn't otherwise used.
- Similar to errors early on (initial framing handling), we don't have a dispatcher from which to read timeout values from, so I put it at infinite so far. This is also done in the mentioned initial cases, with a TODO comment to implement timeout.
- I used MaxDrain like in some other places, which seems appropriate to me.

cyanite:
It occurs to me that neither `PerformServiceHandshake` or the existing `GetServiceHandshakeDelegate` are very good names? No communication is done at this step. It's more routing or pre-dispatching. Should we rename them?

mconnew:
With middleware, it's up to the middleware implementation to decide whether to pass on to the next middleware. Any middleware implementation can short circuit things and decide it's the final piece. Basically any middleware can terminate the chain on a per request basis. This is exactly how ASP.NET Core middleware works. When you use app.Map, it basically stacks middleware implementations which check if the incoming request matches the path, and if it does, handles the request without going to the next middleware, but if it doesn't match, it passes the request to the next middleware.  

Having said that, what you've done is the minimal change and this part of the code already had broken away from the explicit middleware class model so I'm okay keeping it the way you've done it.  

For the timeout, use `ServiceDefaults.SendTimeout` as that's what's used in WCF. The drain size you used is the same as on WCF so that's good.  

I think PerformServiceHandshake is a good name as the serviceHandshake delegate does do a handshake. It negotiates security and any other upgrades. I believe the work it's doing is actually called the connection handshake in the NetMF spec.  

Once you've made the timeout change, I'll merge this. I'm also no seeing any TODO's, did you miss pushing a change?

cyanite:
> I'm also not seeing any TODO's, did you miss pushing a change?

Ah no, sorry, I didn't put it in after all. What I was referring to is another infinite timeout earlier in the framing code, which does have a TODO about using a proper timeout. I meant to copy that, but didn't. And now it's using the one you suggested, so moot. It seems to work.

PR #542, opened 2022-01-29
Making it a bit clearer on how to get started

samsp-msft:


README.md
```diff
@@ -21,7 +23,19 @@ To keep up to date on what's going on with CoreWCF, you can subscribe to the [an
 
 ### How do I get started?
 
-There are pre-release packages available from a NuGet feed hosted in Azure DevOps. You can download the packages by adding the following package source to your list of feeds.
+* Install the Nuget packages listed above, either via the Package Manager Console or the UI.
```
mconnew, line 26: You can also add packages using the dotnet cli. Would it be worth linking to some .NET documentation about adding packages the various ways?

samsp-msft: I would suggest that the vast majority of users for this are probably in VS, and those who aren't will know how to use the dotnet cli to do the same thing.

PR #543, opened 2022-01-29
Pared-down sample for .NET 6 with a single endpoint

samsp-msft:
Note this sample uses nuget package refs rather than relative project refs. This makes it easier for project consumers who probably are after something that doesn't require them to build the entire repo.

### DIFF
--- a/src/Samples/dotnet6server/EchoService.cs
+++ b/src/Samples/dotnet6server/EchoService.cs
@@ -1,6 +1,5 @@
-using System.Threading.Tasks;
-using MyContracts;
 using CoreWCF;
+using MyContracts;
 
 namespace dotnet6Server
 {
@@ -21,10 +20,5 @@ namespace dotnet6Server
         public string FailEcho(string text)
             => throw new FaultException<EchoFault>(new EchoFault() { Text = "WCF Fault OK" }, new FaultReason("FailReason"));
 
-        [AuthorizeRole("CoreWCFGroupAdmin")]
-        public string EchoForPermission(string echo)
-        {
-            return echo;
-        }
     }
 }
--- a/src/Samples/dotnet6server/EchoService.cs
+++ b/src/Samples/dotnet6server/EchoService.cs
@@ -1,0 +1,30 @@
+using System.Threading.Tasks;
+using MyContracts;
+using CoreWCF;
+
+namespace dotnet6Server
+{
+    public class EchoService : IEchoService
+    {
+        public string Echo(string text)
+        {
+            System.Console.WriteLine($"Received {text} from client!");
+            return text;
+        }
+
+        public string ComplexEcho(EchoMessage text)
+        {
+            System.Console.WriteLine($"Received {text.Text} from client!");
+            return text.Text;
+        }
+
+        public string FailEcho(string text)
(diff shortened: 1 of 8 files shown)


samsp-msft:
This binds to the ports that Kestrel is hooked up to - so I'm not sure what happens if you use other bindings that may depend on https for example?

src/Samples/dotnet6server/dotnet6Server.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CoreWCF.ConfigurationManager" Version="0.4.0" />
+    <PackageReference Include="CoreWCF.Http" Version="0.4.0" />
```
mconnew, line 12: This sample isn't using Configuration or NetTcp so you don't need to reference either of them.

src/Samples/dotnet6server/EchoService.cs
```diff
@@ -0,0 +1,30 @@
...
+            System.Console.WriteLine($"Received {text.Text} from client!");
+            return text.Text;
+        }
+
+        public string FailEcho(string text)
+            => throw new FaultException<EchoFault>(new EchoFault() { Text = "WCF Fault OK" }, new FaultReason("FailReason"));
+
+        [AuthorizeRole("CoreWCFGroupAdmin")]
+        public string EchoForPermission(string echo)
+        {
```
mconnew, line 26: The EchoForPermission method with the AuthorizeRoleAttribute is demonstrating using Windows group membership to restrict the ability to call certain operations. This method is uncallable as you don't have anything configured to populate the Windows auth claimset. Can you remove the EchoForPermission method from the interface and implementation as it just confuses things.

src/Samples/dotnet6server/Properties/launchSettings.json
```diff
@@ -0,0 +1,28 @@
...
+    "iisExpress": {
+      "applicationUrl": "http://localhost:8113",
+      "sslPort": 44345
+    }
+  },
+  "profiles": {
+    "dotnet6Server": {
+      "commandName": "Project",
+      "dotnetRunMessages": true,
+      "launchBrowser": true,
```
mconnew, line 14: I don't think `"launchBrowser": true` is applicable to CoreWCF. I think it will just be a nuisance having a browser window open which can't show anything.

src/Samples/dotnet6server/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+});
+
+builder.Services.AddServiceModelServices();
+var app = builder.Build();
+
+app.UseServiceModel(builder =>
+{
+    builder.AddService<EchoService>()
+    .AddServiceEndpoint<EchoService, IEchoService>(new BasicHttpBinding(), "/EchoService/basichttp")
+});
```
mconnew, line 19: Does this compile without the ; at the end of the line?

Issue #270, opened 2020-12-14
Add support for ServiceMetadataBehavior features


mconnew:
This includes support for:

- ?SingleWsdl and ?Wsdl pages to retrieve the wsdl
- html helper page which includes a link to the wsdl as well as client usage examples. This is used as a service health page
- mex endpoint supporting the IMetadataExchange contract on multiple bindings

mconnew:
Some notes about implementation.  
The html helper page needs to be identical to what WCF produces as some people have built tooling around the output from that page.  
This has a dependency on types in the System.Web.Services.Description. I have opened an issue dotnet/wcf#4464 in the wcf repo to track creating a package holding these types.

Danielku15:
@mconnew Anything the community might be able to help you on this topic?

mconnew:
At this stage, no. Once I have something working, throwing every contract and service at it and verifying the outcome against what's expected will be the most help. Contributing a test library of contracts, bindings, and expected WSDL's is really useful. Unfortunately this is a big chunk of work which is an all or nothing. Nothing works without every piece, and very little is testable in isolation.

mconnew:
For anyone watching this issue, I'm about to commit a large amount of code to the feature/wsdl branch to enable contribution of tests. There's a lot of cleanup I need to do in the code, and need to modify the api a little to allow providing a url to expose the metadata page on. But that shouldn't affect the ability to provide tests as I've abstracted all of that away in the 2 initial tests I've created.  
The metadata feature had to go into the Primitives package as there was a need for types in Primitives to implement `IWsdlExportExtension`. That interface then pulled in `WsdlExporter` and `WsdlContractConversionContext` which then pulled in pretty much everything. So it didn't make sense to create a separate package.  

If you want to contribute test scenarios, it should be relatively simple. Take a look at `BasicHttpSimpleServiceTest.cs` and `NetTcpSimpleServiceTest.cs`. The `BasicHttpRequestReplyEchoString` test body is just this:
```c#
        [Fact]
        public async Task BasicHttpRequestReplyEchoString()
        {
            await TestHelper.RunSingleWsdlTestAsync<SimpleEchoService, IEchoService>(new BasicHttpBinding(), _output);
        }
```
[...]

PR #550, opened 2022-02-11
WSDL support

mconnew:


src/CoreWCF.Primitives/src/CoreWCF.Primitives.csproj
```diff
@@ -22,6 +22,7 @@
     <PackageReference Include="System.Security.Principal.Windows" Version="5.0.0" />
     <PackageReference Include="System.Reflection.DispatchProxy" Version="4.7.1" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
+    <PackageReference Include="System.Web.Services.Description" Version="4.9.0" />
```
birojnayak, line 25: does this have any issues across platforms ?

src/CoreWCF.Primitives/src/CoreWCF/Channels/SecurityBindingElement.cs
```diff
@@ -779,7 +783,7 @@ public static SecurityBindingElement CreateSecureConversationBindingElement(Secu
         {
             if (bootstrapSecurity == null)
             {
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("bootstrapBinding");
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(bootstrapSecurity));
```
birojnayak, line 786: thank you !

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceConfiguration.cs
```diff
@@ -27,7 +29,18 @@ public List<IServiceDispatcher> GetDispatchers()
         {
             if (_dispatchers == null)
             {
-                _dispatchers = DispatcherBuilder.BuildDispatcher<TService>(this, _services);
+                // The services can't be Open'd until after ASP.NET Core is running and listening for requests. The Open call
+                // is started and allowed to run in the background. If a request comes in before the dispatchers have finished
```
birojnayak, line 33: did you run into this issue in your testing ?

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceOptions.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using CoreWCF.Description;
+
+namespace CoreWCF.Configuration
+{
+    public class ServiceOptions
```
birojnayak, line 10: need to understand about this class ? will sync

birojnayak (changes requested):
looks good.. some minor comments... need clarifications around serviceoption

src/CoreWCF.Primitives/src/CoreWCF/Channels/SecurityBindingElement.cs
```diff
@@ -838,5 +842,263 @@ public static TransportSecurityBindingElement CreateSspiNegotiationOverTransport
...
+                        if (be.GetType().FullName.Equals("CoreWCF.Channels.TcpTransportBindingElement"))
+                        {
+                            transportTokenAssertionProvider = new SslStreamSecurityBindingElement();
+                            break;
+                        }
+                    }
+                }
+            }
+
+            XmlElement transportTokenAssertion = transportTokenAssertionProvider.GetTransportTokenAssertion();
```
jnyrup, line 973: Is it safe to assume that a `transportTokenAssertionProvider` is always found above?
Otherwise this line may throw an NRE.

src/CoreWCF.Primitives/src/CoreWCF/Description/WsdlHelper.cs
```diff
@@ -0,0 +1,328 @@
...
+            // policy
+            IEnumerable<XmlElement> bindingPolicies = FindAllElements(wsdl.Extensions, MetadataStrings.WSPolicy.Elements.Policy);
+            string policyIdStringPrefixFormat = "{0}_";
+            foreach (XmlElement policyElement in bindingPolicies)
+            {
+                XmlNode policyId = policyElement.Attributes.GetNamedItem(MetadataStrings.Wsu.Attributes.Id, MetadataStrings.Wsu.NamespaceUri);
+                string policyIdString = policyId.Value;
+                string policyIdStringWithOldBindingName = string.Format(CultureInfo.InvariantCulture, policyIdStringPrefixFormat, oldBindingName);
+                string policyIdStringWithNewBindingName = string.Format(CultureInfo.InvariantCulture, policyIdStringPrefixFormat, newBindingName);
+                if (policyId != null && policyIdString != null && policyIdString.StartsWith(policyIdStringWithOldBindingName, StringComparison.Ordinal))
```
jnyrup, line 161: `policyId` cannot be null here due to `string policyIdString = policyId.Value;` on line 158

src/CoreWCF.Primitives/src/CoreWCF/Description/WsdlHelper.cs
```diff
@@ -0,0 +1,328 @@
...
+        {
+            IEnumerable<XmlElement> bindingPolicyReferences = FindAllElements(extensions, MetadataStrings.WSPolicy.Elements.PolicyReference);
+            string policyReferencePrefixFormat = "#{0}_";
+            foreach (XmlElement policyReferenceElement in bindingPolicyReferences)
+            {
+                XmlNode policyReference = policyReferenceElement.Attributes.GetNamedItem(MetadataStrings.WSPolicy.Attributes.URI);
+                string policyReferenceValue = policyReference.Value;
+                string policyReferenceValueWithOldBindingName = string.Format(CultureInfo.InvariantCulture, policyReferencePrefixFormat, oldBindingName);
+                string policyReferenceValueWithNewBindingName = string.Format(CultureInfo.InvariantCulture, policyReferencePrefixFormat, newBindingName);
+                if (policyReference != null && policyReferenceValue != null && policyReferenceValue.StartsWith(policyReferenceValueWithOldBindingName, StringComparison.Ordinal))
```
jnyrup, line 199: `policyReference` cannot be null here due to `string policyReferenceValue = policyReference.Value;` on line 196

src/CoreWCF.Primitives/src/CoreWCF/Xml/Serialization/SoapSchemaExporter.cs
```diff
@@ -0,0 +1,478 @@
...
+                AddSchemaImport(ns, referencingNs);
+        }
+
+        void AddSchemaImport(string ns, string referencingNs)
+        {
+            if (referencingNs == null || ns == null) return;
+            if (ns == referencingNs) return;
+            XmlSchema schema = schemas[referencingNs];
+            if (schema == null) throw new InvalidOperationException(SR.Format(SR.XmlMissingSchema, referencingNs));
+            if (ns != null && ns.Length > 0 && FindImport(schema, ns) == null)
```
jnyrup: `ns` is already null-checked on line 136

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
```diff
@@ -325,6 +339,168 @@ public ExtendedProtectionPolicy ExtendedProtectionPolicy
...
+                {
+                    assertionNames.Add(TransportPolicyConstants.DigestHttpAuthenticationName);
+                }
+
+                if (effectiveAuthenticationSchemes.IsSet(AuthenticationSchemes.Basic))
+                {
+                    assertionNames.Add(TransportPolicyConstants.BasicHttpAuthenticationName);
+                }
+
+                if (assertionNames != null && assertionNames.Count > 0)
```
jnyrup, line 397: I guess this might be an extra null cautious style, but `assertionNames` is never null here

g7ed6e:
@mconnew I started to review this PR but most of my remarks are more related to C# syntax modernisation than the actual WSDL feature. Maybe the code modernisation could be handled later with a more systematic approach to ensure syntax consistency in the whole project so feel free to ignore remarks. If you are open to accept PR that modernize the codebase please let me know.

roend83:
@mconnew Is there anything I can help with to get this merged and released? This is the last thing holding me back from migrating my services over.

mconnew:
@g7ed6e, that would be great. Timing wise I think it would be good to do after we've released the 1.0 version. @roend83, just more eyes on the PR to review it. I'm working on reviewing the WebHttpBinding PR and the WS-Federation PR to merge them first, then I'll merge this. Hopefully in the next couple of weeks that will be done.

cpx86:
Out of curiosity, are there any plans yet for solving the issue mentioned [here](https://github.com/CoreWCF/CoreWCF/issues/270#issuecomment-1041542485) regarding EnumMembers not working under .NET Core/.NET 5+? For myself at least, my main interest in CoreWCF is to be bale to upgrade WCF applications to the latest .NET without breaking backwards-compatibility; having WSDL support is a fairly important part of this.

mconnew:
@cpx86, it's on my radar. I'm currently spending most of my time reviewing a couple of big PR's from other contributors. After I've completed that, I will look into this problem. There are many other things broken about schema export in .NET that I've had to work around in CoreWCF so I'll probably need to solve this in a similar way. CoreWCF works on all supported versions of .NET (including .NET Framework) so even if I fixed it in .NET 6 as a servicing fix, I still need to find a workaround in CoreWCF.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpsTransportBindingElement.cs
```diff
@@ -73,5 +78,97 @@ public override T GetProperty<T>(BindingContext context)
...
+                SecurityBindingElement.ExportTransportSecurityBindingElement(dummyTransportBindingElement, this, exporter, context);
+            }
+        }
+
+        private XmlElement CreateWsspAssertion(Security.SecurityPolicyVersion policyVersion, string name)
+        {
+            string policyNamespace = string.Empty;
+            if (policyVersion == Security.SecurityPolicyVersion.WSSecurityPolicy11)
+            {
+                policyNamespace = @"http://schemas.xmlsoap.org/ws/2005/07/securitypolicy"; //WSSecurityPolicy11.WsspNamespaceUri
```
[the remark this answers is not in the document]
mconnew: It's internal

src/CoreWCF.Http/src/CoreWCF/Channels/HttpsTransportBindingElement.cs
```diff
@@ -73,5 +78,97 @@ public override T GetProperty<T>(BindingContext context)
...
+            }
+            else if(policyVersion == Security.SecurityPolicyVersion.WSSecurityPolicy12)
+            {
+                policyNamespace = @"http://docs.oasis-open.org/ws-sx/ws-securitypolicy/200702"; //WSSecurityPolicy12.WsspNamespaceUri
```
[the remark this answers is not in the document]
mconnew: Same as above :smile:

src/CoreWCF.Http/src/CoreWCF/Channels/HttpsTransportBindingElement.cs
```diff
@@ -73,5 +78,97 @@ public override T GetProperty<T>(BindingContext context)
...
+            }
+
+            var doc = new XmlDocument();
+            XmlElement result = doc.CreateElement("sp", // WSSecurityPolicy.WsspPrefix
```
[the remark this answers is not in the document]
mconnew: WSSecurityPolicy is also internal.

src/CoreWCF.Http/src/CoreWCF/DiagnosticUtility.cs
```diff
@@ -1,193 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
-using System;
-using System.Diagnostics;
-using System.Runtime.CompilerServices;
-using CoreWCF.Runtime;
```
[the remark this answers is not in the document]
mconnew: Into common. So it will get pulled in to all the projects and appear under shared\CoreWCF in VS.

src/CoreWCF.Http/src/CoreWCF/Description/MetadataStrings.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
```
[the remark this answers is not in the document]
mconnew: My current plan is to merge them each into their own feature/FEATURENAME branch then do a massive merge session working out what order is best.

src/CoreWCF.Metadata/tests/Helpers/WebHostBuilderTestHelperExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
```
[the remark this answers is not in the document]
mconnew: Oops, I'll remove this. This is left over from some test helper refeactoring.

src/CoreWCF.Metadata/tests/CoreWCF.Metadata.Tests.csproj
```diff
@@ -0,0 +1,59 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
mconnew: Tests only. I did originally plan on creating a package with this name which depended on Primitives, but I ended up with Primitives needing to depend on Metadata types which meant it would always get pulled in anyway.  I tried moving just the needed types into Primitives but that ended up pulling across 90% of the code, so I folded the implementation into Primitives. I kept the tests as a separate project as it depends on Http, NetTcp, and anything else which we build which provides WSDL support.

src/CoreWCF.Metadata/tests/Helpers/WsdlHelper.cs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                var response = await client.GetAsync(singleWsdlPath);
+                Assert.True(response.IsSuccessStatusCode, $"Response status for url {singleWsdlPath} is {(int)response.StatusCode} {response.StatusCode} {response.ReasonPhrase}");
+                generatedWsdlTxt = await response.Content.ReadAsStringAsync();
+            }
+
+            var xmlFileName = Path.Combine("Wsdls", Path.GetFileNameWithoutExtension(sourceFilePath) + "." + callerMethodName + ".xml");
+            if (!File.Exists(xmlFileName))
+            {
+                // If sourceFilename.methodname.xml doesn't exist, then look for sourceFilename.xml. This enables use of a single expected wsdl file
```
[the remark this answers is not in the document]
mconnew: You've got it, this is to enable multiple test cases generating the same wsdl. An example might be one test uses asp.net core endpoints, another case doesn't. Should still produce the same WSDL in both cases. Or you could have one test use localhost and another test use the hostname.

src/CoreWCF.NetTcp/src/CoreWCF/DiagnosticUtility.cs
```diff
@@ -1,195 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
-using System;
-using System.Diagnostics;
-using System.Runtime.CompilerServices;
-using CoreWCF.Runtime;
-
```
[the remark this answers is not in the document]
mconnew: Common shared code, see earlier comment.

src/CoreWCF.Primitives/src/CoreWCF.Primitives.csproj
mconnew: None. This works on all platforms as it targets netstandard2.0 and type forwards on .NET Framework to system.web.services.dll in the GAC.

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceConfiguration.cs
mconnew: Yes. It's a hard race condition to hit, but when setting the right break points while debugging, you can hit it reliably.

resources.targets
```diff
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Target Name="GenerateResourcesSource" 
-      Inputs="$(StringResourcesPath)"
-      Outputs="$(IntermediateResOutputFileFullPath)"
+      Inputs="$(StringResourcesPath);$(CommonStringResourcesPath)"
+      Outputs="$(IntermediateResOutputFileFullPath),$(IntermediateCommonResOutputFileFullPath)"
```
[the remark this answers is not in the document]
mconnew: Nice catch. This explains an odd failure where sometimes modifying the common resx file failed to regenerate the source file without a clean and build.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SecurityBindingElement.cs
```diff
@@ -838,5 +842,263 @@ public static TransportSecurityBindingElement CreateSspiNegotiationOverTransport
...
+
+            if (transportTokenAssertion == null)
+            {
+                if (transportTokenAssertion == null)
```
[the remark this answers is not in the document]
mconnew: Thanks. This was left over from where some other code ran to try and generate the transportTokenAssertion, and if it was still null, then it would throw. I refactored to have the caller always generate it and didn't clean this up correctly.

src/CoreWCF.Primitives/src/CoreWCF/Channels/SecurityBindingElement.cs
mconnew: In theory it should, but there could be a bug, or a third party transport which could mean it isn't. I'll use null propagation to fix this: `transportTokenAssertionProvider?.GetTransportTokenAssertion()`, that way the following null check for transportTokenAssertion will fire if this is null.

src/CoreWCF.Primitives/src/CoreWCF/Description/WsdlHelper.cs
mconnew: policyId could be null in which case line 158 would throw NRE and this code wouldn't be reached. All binding polices created by CoreWCF will have an Id, but this is extensible and someone else's code might not put the Id there. The code needs to be more defensive. I think using null coalescing on line 158 is sufficient. This way line 158 can't throw a NRE and this line becomes reachable even if policyId is null, in which case the null check here becomes useful.

src/CoreWCF.Primitives/src/CoreWCF/Description/WsdlHelper.cs
mconnew: Same pattern as policyId. I'll do the same fix, use null coalescence when setting policyReferenceValue.

src/CoreWCF.Primitives/src/CoreWCF/Xml/Serialization/SoapSchemaExporter.cs
mconnew: Good catch

src/CoreWCF.Primitives/src/CoreWCF/Channels/MetadataMiddleware.cs
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                    if (helpPageUri != null && !helpPageUri.Equals(httpsMetadataUri))
+                    {
+                        // Only map the help page uri if it's not null, and different from https metadata uri.
+                        MapMetadata(branchApp, helpPageUri.AbsolutePath, metadataExtension.ConfigureWith(helpPageUri, true));
+                    }
+                }
+            }
+
+            branchApp.Use(_ => {
```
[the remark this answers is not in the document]
mconnew: I've reworked this, but we need potentially 4 mappings. HTTP and HTTPS for the help page and the wsdl. There's code to avoid duplicate mapping when the help page and wsdl are at the same url, but we do still need one for each of HTTP and HTTPS is they are both enabled.

mconnew:
@cpx86, Enum's now work. I detect that case and fix up the missing value in the EnumDataContract using reflection.

cpx86:
@mconnew Great to hear! I will try it out :)

cpx86:
@mconnew I found a limitation in the implementation. It only works if the Enum is directly referenced by the ServiceContract. If the Enum is wrapped by another DataContract then it won't be detected.

For example, this contract will fail (adapted from the unit test contract, just wrapped the Enum in another type):
```csharp
[ServiceContract(Namespace = Constants.NS, Name = "EnumService")]
public interface IEnumService
{
    [OperationContract] void AcceptWrapped(TestWrappedEnum accept);
    [OperationContract] TestWrappedEnum RequestWrapped();
}
[DataContract]
public enum TestEnum
{
    [EnumMember] One = 1,
    [EnumMember] Two = 2,
    [EnumMember] Three = 3,
    [EnumMember] Five = 5,
}

[DataContract]
public class TestWrappedEnum
{
    [DataMember] public TestEnum Enum { get; set; }
}```

Interestingly the above starts working if you add back the previous operations where the Enum is directly referenced. So I guess the issue is that the enum fix needs to be somehow done recursively throughout the schema, whereas now it's only applied to the input/output contracts of the service?

**Comment by mconnew** [comment]
I'll take a look and make sure it's fixed. Thanks for the sample contract, it really helps a lot.

**Comment by mconnew** [comment]
@cpx86, the issue was when wrapping the Enum type in another type, CoreWCF doesn't directly add it to the DataContractSet, it's done recursively by the runtime. The fix was to go through all DataContract's in the DataContractSet before exporting to fixup the EnumDataContract instances instead of fixing them at the time of adding.

**Comment by cpx86** [comment]
@mconnew Nice, that seems to have resolved the issue completely.

...And I found yet another issue. 😅  If you have a contract which includes two types that would be represented as the same WSDL type, the exporter fails. For example `IEnumerable<string>` and `List<string>` both map to `ArrayOfstring`. The exception and partial stack trace is:

> System.InvalidOperationException: DataContract for type 'System.Collections.Generic.IEnumerable`1[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' cannot be added to DataContractSet since type 'System.Collections.Generic.List`1[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' with the same data contract name 'ArrayOfstring' in namespace 'http://schemas.microsoft.com/2003/10/Serialization/Arrays' is already present and the contracts are not equivalent.
>    at System.Runtime.Serialization.DataContractSet.InternalAdd(XmlQualifiedName name, DataContract dataContract)
>    at System.Runtime.Serialization.DataContractSet.Add(DataContract dataContract)
>    --- End of inner ExceptionDetail stack trace ---
>    at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1&amp; arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
>    at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
>    at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
>    at CoreWCF.Runtime.Serialization.DataContractSetEx.Add(Type type) in C:\dev\gh\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\Runtime\Serialization\DataContractSetEx.cs:line 41

Smallest sample code to reproduce:
```csharp
[OperationContract()]
List<string> EchoStringList(List<string> echo);

[OperationContract()]
IEnumerable<string> EchoStringEnumerable(IEnumerable<string> echo);
```

mconnew:
@cpx86, thank you. I'll get this fixed too as soon as I work out the problem. Your testing has been really helpful.

mconnew:
@cpx86, I've opened an issue in the runtime repo here: dotnet/runtime#67371. As a workaround I'll implement the missing Equals methods in CoreWCF and "prescreen" the DataContract instances before calling Export.

mconnew:
This PR was merged to main via an integration branch.

PR #568, opened 2022-03-10
Fix WebSocketSettings not being applied to https

### LINKED ISSUE #567: NetHttpBinding.WebSocketSettings must be set on HTTPS transport binding element too
When using the NetHttpBinding with HTTPS transport, configuring any WebSocketSettings is not passed to the actual transport binding element. This happens because `HttpBindingBase` is only exposing the HTTP web socket settings - in .NET 4.8 that instance is forwarded to the HTTPS transport binding element too.

mconnew:
Fixes #567


gvkries:
Looks good to me. Thank you!

Issue #428, opened 2021-08-23
Set up multiple services with IServiceBuilder


tapizquent:
Is it possible to add multiple services with the IServiceBuilder?

That is, implement something like the following

`builder.AddService<Service1>();`
`builder.AddServiceEndpoint<Service1, IService1>(new BasicHttpBinding(), "/service1");`
`builder.AddService<Service2>();`
`builder.AddServiceEndpoint<Service2, IService2>(new BasicHttpBinding(), "/service2");`

mconnew:
I designed it to do so, but haven't actually tested it. The implementation does a foreach over the service type to initialize it. There's a few rough edges around doing that, such as dependency injection of IServiceBehavior implementations has no way to configure a behavior to only be for a particular service. I have some ideas how to do that, but as nobody has asked, I haven't implemented them yet.  

So give it a try and let me know if you have any problems with doing so. If it doesn't work, I'll do my best to get it working as soon as possible.

mconnew:
Keeping open as we need to add tests for this scenario and fix any problems that arise. Marking as help wanted for the test writing aspect. You wouldn't be required to fix any problems discovered if not comfortable with that.

BradBarnich:
this is what we are doing, and it seems to work well?
```
sb.AddService<ApplicationManager>();
sb.AddServiceEndpoint<ApplicationManager, IApplicationManager>(
    HttpEndpointFactory.CreateCoreWCFBinding( SecurityMode.Transport, HttpClientCredentialType.None ),
    "/ApplicationManagerService" );
sb.ConfigureServiceHostBase<ApplicationManager>( serviceHost =>
{
    new StandardServiceHostConfiguration().Configure( serviceHost );
} );

sb.AddService<Adapter.LocalDataService.LocalDataService>();
sb.AddServiceEndpoint<Adapter.LocalDataService.LocalDataService, ILocalData>(
    HttpEndpointFactory.CreateCoreWCFBinding( SecurityMode.Transport, HttpClientCredentialType.Certificate ),
    "/LocalDataService" );
sb.ConfigureServiceHostBase<Adapter.LocalDataService.LocalDataService>( serviceHost =>
{
    new StandardServiceHostConfiguration().Configure( serviceHost );
    new SecureServiceHostConfiguration().Configure( serviceHost );
    new UserContextServiceHostConfiguration().Configure( serviceHost );
} );
```

mconnew:
[...]

PR #583, opened 2022-03-30
Add test for adding multiple services

jonlouie:
Added a test to verify multiple services can be added using subsequent calls to `IServiceBuilder.AddService<TService>()`

Closes: #428


jonlouie:
@mconnew Following up here per our brief exchange last week on issue #428

g7ed6e (approved):
Thanks!

PR #590, opened 2022-04-04
Introduce CoreWCF project templates

### LINKED ISSUE #541: [Visual Studio] Project Templates for CoreWCF
There is a bit of a chicken and egg issue with templates, but the net story as of this moment is that templates installed via `dotnet new install <package.id>` will show up in VS once they are installed. Instructions for creating templates are at https://github.com/sayedihashimi/template-sample.

* CoreWCF should have a nuget package with project & item templates in it.
* Those templates can then have references to the other CoreWCF nuget packages as required.
* Users will be able to install then using something like `dotnet new install CoreWCF.Templates`
* Those templates will then show up in Visual Studio for the Project New dialog, and will be usable from cmd line and VS code.

g7ed6e:
Resolves #541


src/CoreWCF.Templates/src/CoreWCF.Templates.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageId>CoreWCF.Templates</PackageId>
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+  </PropertyGroup>
+  <PropertyGroup>
+    <PackageType>Template</PackageType>
+    <PackageVersion>1.0</PackageVersion>
+    <!--<PackageId>AdatumCorporation.Utility.Templates</PackageId>
```
g7ed6e, line 10: clean

azure-pipelines-ci.yml
```diff
@@ -14,7 +14,7 @@ pr: none
 variables:
   _solution: 'src/CoreWCF.sln'
   _libraryProjects: 'src/CoreWCF.*/src/CoreWCF.*.csproj'
-  _testProjects: '**/*.Tests.csproj'
+  _libraryTestProjects: '**/*.Tests.csproj'
```
g7ed6e, line 17: '**/*.Tests.csproj;!**/*.Templates.Tests.csproj'

src/CoreWCF.Templates/src/templates/.editorconfig
```diff
@@ -0,0 +1,133 @@
+# Pour en savoir plus sur .editorconfig, consultez https://aka.ms/editorconfigdocs
```
samsp-msft, line 1: Is there one of these for the project as a whole - if so why is this different, and if not, should this be moved up to apply at the root of the repo?

src/CoreWCF.Templates/src/templates/CoreWCFService/IService.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+            if (composite.BoolValue)
+            {
+                composite.StringValue += "Suffix";
+            }
+            return composite;
+        }
+    }
+
+    // Utilisez un contrat de données comme indiqué dans l'exemple ci-après pour ajouter les types composites aux opérations de service.
```
samsp-msft, line 40: Can we have the comment in English please - I don't think we have localization support for templates. I'm assuming this is the french version of "// Use a data contract as illustrated in the sample below to add composite types to service operations."

src/CoreWCF.Templates/src/templates/CoreWCFService/Program.cs
```diff
@@ -0,0 +1,39 @@
+﻿var builder = WebApplication.CreateBuilder();
```
samsp-msft, line 1: Are the \#if's resolved at template creation time or is this the code expected to be seen by the user?

If the goal is to enable all the different scenarios, then I think we should reconsider. There is a balance needed in templates between:
* The least number of lines
* The simplest form of the code
* Handling every possibility

Generally we favor the simplest form of the code - so as to not to scare users off, and to provide a good starting point.

I think we should pick a default for the template, and have that for both the modern form and classic program.cs formats. I would suggest:

| Feature | Included in template |
| --- | --- |
| WSDL | Yes |
| NetTCP | No |
| BasicHttpBinding | Yes |
| Https BasicHttpBinding | Yes |

src/CoreWCFTemplates.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
samsp-msft, line 1: I understand the need for this while you are developing the templates, but should these be incorporated into the main sln so that as people add support, the templates are also under their noses, so they know to update them too?

src/CoreWCF.Templates/src/templates/CoreWCFService/IService.cs
```diff
@@ -0,0 +1,61 @@
+﻿#if(!impliciteUsings)
+using System.Runtime.Serialization;
+using CoreWCF;
+using System;
```
mconnew, line 4: Order of usings should be alphabetical. So:
```c#
using CoreWCF;
using System;
using System.Runtime.Serialization;
```

src/CoreWCF.Templates/src/templates/.editorconfig
g7ed6e: This file overrides is there to override the one from CoreWCF repo as i guess template users does not want the MIT license header.

src/CoreWCF.Templates/src/templates/CoreWCFService/IService.cs
g7ed6e: Sure

src/CoreWCF.Templates/src/templates/CoreWCFService/Program.cs
g7ed6e: > Are the #if's resolved at template creation time or is this the code expected to be seen by the user?

Yes please see https://github.com/dotnet/templating/wiki/Conditional-processing-and-comment-syntax
> 
> If the goal is to enable all the different scenarios, then I think we should reconsider. There is a balance needed in templates between:
> 
> * The least number of lines
> * The simplest form of the code
> * Handling every possibility
> 
> Generally we favor the simplest form of the code - so as to not to scare users off, and to provide a good starting point.
> 
> I think we should pick a default for the template, and have that for both the modern form and classic program.cs formats. I would suggest:
> 
> Feature	Included in template
> WSDL	Yes
> NetTCP	No
> BasicHttpBinding	Yes
> Https BasicHttpBinding	Yes

Then i will drop netTcp binding support

src/CoreWCFTemplates.sln
g7ed6e: Running unit test requires the current code to be built and packaged thus i moved them in a separate solution to avoid embarassing users who want to work on CoreWCF runtime code. Please see https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/README.md#running-projecttemplate-tests , i'm trying to do something similar but without the hive dependency.

src/CoreWCF.Templates/src/templates/CoreWCFService/IService.cs
samsp-msft: I took it from the .NET Framework project, so I'm assuming its the same.

README.md
```diff
@@ -33,6 +35,17 @@ To keep up to date on what's going on with CoreWCF, you can subscribe to the [an
...
+#### *Using `dotnet new` project templates*
+
+* Install CoreWCF project templates, create a directory for your project and cd inside the directory and initialize your project
+```cmd
+dotnet new --install CoreWCF.Templates ^
+mkdir MyService && cd MyService ^
+dotnet new corewcf
+```
+* `CoreWCFService` project template supports the following arguments
+  * `--binding`: `nettcp` or `basic` (default)
```
g7ed6e, line 47: this is no longer true.
add supported options and default values

templates/TestTemplatesStage.yml
```diff
@@ -0,0 +1,56 @@
...
+      displayName: 'Use .NET 6 sdk'
+      inputs:
+        packageType: sdk
+        version: '6.0.x'
+        installationPath: $(Agent.ToolsDirectory)/dotnet          
+
+    - task: PowerShell@2
+      inputs:
+        targetType: 'filePath'
+        filePath: src/CoreWCF.Templates/Run-Templates-UnitTests.ps1
```
g7ed6e, line 56: @mconnew  I ensured the script return a non zero exit code to fail the build unit tests are failing

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/dotnetcli.host.json
```diff
@@ -0,0 +1,33 @@
+{
```
g7ed6e, line 1: add the vs support file like here https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/.template.config/ide.host.json

README.md
g7ed6e: done

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/dotnetcli.host.json
g7ed6e: image

g7ed6e: image

src/CoreWCF.Templates/src/templates/CoreWCFService/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+    serviceBuilder.AddService<Service>();
+#if (NoHttps)
+    serviceBuilder.AddServiceEndpoint<Service, IService>(new BasicHttpBinding(), "/Service.svc");
+#else
+    serviceBuilder.AddServiceEndpoint<Service, IService>(new BasicHttpBinding(BasicHttpSecurityMode.Transport), "/Service.svc");
+#endif
+
+#if (!NoWsdl)
+    var serviceMetadataBehavior = app.Services.GetRequiredService<ServiceMetadataBehavior>();
+
```
g7ed6e, line 25: drop this line

src/CoreWCF.Templates/tests/BasicTests.cs
```diff
@@ -0,0 +1,212 @@
...
+            return this;
+        }
+
+        public TestVariation Framework(string framework)
+        {
+            Arguments.Add($"--framework {framework}");
+            return this;
+        }
+
+        public TestVariation Net6()
```
g7ed6e, line 70: drop

README.md
```diff
@@ -29,6 +29,19 @@ To keep up to date on what's going on with CoreWCF, you can subscribe to the [an
...
+
+* Install CoreWCF project templates, create a directory for your project and cd inside the directory and initialize your project
+```cmd
+dotnet new --install CoreWCF.Templates && mkdir MyService && cd MyService && dotnet new corewcf
+```
+* `Core WCF Service` project template creates a minimal ASP.NET Core web application thats exposes the well-known WCF default service using a `BasicHttpBinding`.
+It supports the following arguments:
+  * `--framework`: `net6.0` (default), `net5.0`, `netcoreapp3.1`, `net48`, `net472` and `net462` are valid values.
+  * `--use-program-main`: whether to turn off ASP.NET Core minimal API hosting. This option only affects `net6.0` projects as other target require a `Startup` and a `Program` class. 
+  * `--no-https`: whether to turn off HTTPS and use BasicHttpSecurityMode.None. Default is HTTPS enabled a `BasicHttpBinding` using `BasicHttpSecurityMode.Transport`.
```
g7ed6e, line 42: Missing ˋ around BasicHttpSecurityMode.None

g7ed6e: Missing `using` a ..

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
```diff
@@ -0,0 +1,189 @@
+{
+  "$schema": "http://json.schemastore.org/template",
+  "author": "CoreWCF",
```
g7ed6e, line 3: @mconnew author is displayed in VS new project wizard.

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/ide.host.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "http://json.schemastore.org/vs-2017.3.host",
+    "icon": "ide/Empty.png",
```
g7ed6e, line 3: This is a default empty C# project logo

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
```diff
@@ -0,0 +1,189 @@
...
+  "sourceName": "CoreWCFService",
+  "preferNameDirectory": true,
+  "shortName": "corewcf",
+  "tags": {
+    "language": "C#",
+    "type": "project"
+  },
+  "name": "Core WCF Service",
+  "defaultName": "CoreWCFService",
+  "description": "A project template to host a Core WCF service.",
```
g7ed6e, line 15: Is this wording correct?

Kralizek:
As someone who is extremely n00b on GitHub and is just lurking the repo. Is it normal that whenever I land on this PR from the notification summary, I see nothing in the "files changed" panel?

g7ed6e:
> As someone who is extremely n00b on GitHub and is just lurking the repo. Is it normal that whenever I land on this PR from the notification summary, I see nothing in the "files changed" panel?

i already encountered the issue you are describing with other PR. But with this one it works for me. I see all modifications there https://github.com/CoreWCF/CoreWCF/pull/590/files

Kralizek:
@g7ed6e 
If I click the link you posted, it works. Same is if i click again on the "modified files".

Could it be that the notification points to commits that have been removed from the PR path via forced pushes?

g7ed6e:
> @g7ed6e If I click the link you posted, it works. Same is if i click again on the "modified files".
> 
> Could it be that the notification points to commits that have been removed from the PR path via forced pushes?

yes i force pushed several times

Kralizek:
Oh okay. In my noobness i thought it was very bad to force push. Sorry for the annoyance, I'll get back to lurk.

mconnew:
@Kralizek, it's bad to force push to a branch other people will be working from. The model we use here is make your own fork of the repo, make your branch there, then make a pull request from that branch in your fork to the CoreWCF repo. When you force push to your own branch, GitHub updates the PR and reruns the tests. The problem with force pushing comes if I were to force push to main. That changes the history of the branch and when other people are making changes based off of main, it makes it very hard to reconcile their changes when the base branch has had its history changed. But it's perfectly normal and common to force push to a temporary branch used for feature development.

mconnew:
@g7ed6e, just an FYI, I'm going to take my time reviewing this and will have multiple people try it out and give their thoughts on it. I want to make sure the initial release has an intuitive user experience. It might take a few weeks to fully review this.

Kralizek:
@mconnew thanks for the explanation

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
samsp-msft: I would suggest either:
* CoreWCF Project 
* or if it doesn't balk at having a URL as the author, then https://github.com/CoreWCF/CoreWCF

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
samsp-msft, line 15:
```suggestion
  "description": "A project template for creating a server application which hosts a Core WCF service.",
```

src/CoreWCF.Templates/src/templates/CoreWCFService/IService.cs
```diff
@@ -0,0 +1,61 @@
+﻿#if(!impliciteUsings)
+using CoreWCF;
+using System;
+using System.Runtime.Serialization;
+#endif
+
+namespace CoreWCFService
```
samsp-msft, line 7: Its not usual for templates to have multiple classes etc in one file. But in this case, I think its appropriate. It makes it very clear that this is all the contract stuff, and so you don't have to go fishing to find it.

src/CoreWCF.Templates/src/templates/CoreWCFService/Properties/launchSettings.json
```diff
@@ -0,0 +1,16 @@
+{
+  "profiles": {
+    "CoreWCFService": {
+      "commandName": "Project",
+      "launchBrowser": true,
```
samsp-msft, line 5: IMHO this line should be removed. We don't have the fancy test page that used to exist for soap services. The browser comes up saying it can't find a url, and it controls the lifetime of the process - so closing the browser stops debugging. Without it the process stays running and you can then test with a client, query for wsdl etc.

samsp-msft (approved):

samsp-msft:
This is freaking cool, I love it. I created a couple of variations to see how they look and they all look good. The main change I'd ask for is to remove the launch browser from launch.settings, but otherwise is good.

@mconnew & @g7ed6e- Is there a way that we could publish a preview of the template package (that refs 1.0 of corewcf) and ask for user feedback. I think this is one place where people really need to try it to find issues. That and a quick post in the announcement repo asking people to try it?

src/CoreWCF.Templates/src/templates/CoreWCFService/Properties/launchSettings.json
g7ed6e: i dropped the launchBrowser

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
g7ed6e: I used https://github.com/CoreWCF as author.

samsp-msft:
@g7ed6e - For the <PackageReference ... > in the resulting csproj it seems to be picking up a version from the current build. Should this actually be a wildcard so to start with its would be 
```xml
<PackageReference Include="CoreWCF.Http" Version="1.*" />
```
That means it would pick up the latest version of 1.x but wouldn't pickup a 2.0. The goals for API compat that we have stated for incremental releases is that they won't break existing APIs.

g7ed6e:
1. Add a separate pipeline to release the templates indenpendently
2. Exclude modifications done to files matching the templates folder from triggers

g7ed6e:
> @g7ed6e - For the <PackageReference ... > in the resulting csproj it seems to be picking up a version from the current build. Should this actually be a wildcard so to start with its would be
> 
> ```
> <PackageReference Include="CoreWCF.Http" Version="1.*" />
> ```
> 
> That means it would pick up the latest version of 1.x but wouldn't pickup a 2.0. The goals for API compat that we have stated for incremental releases is that they won't break existing APIs.

done

g7ed6e:
@mconnew i had a look to the path exclusion you requested to be added in the TestTemplates stage I introduced in the PR pipeline and I think this is not natively supported by azure devops. 
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml
https://stackoverflow.com/questions/61213779/azure-devops-condition-on-sources-paths
Thus i added a new pipeline. azure-pipelines-templates.yml. However the new pipeline is not ran in my PR, I think a new build definition should be created in azure devops.

src/CoreWCF.Templates/src/templates/CoreWCFService/CoreWCFService.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <!--#if(nullableEnabled)-->
```
mconnew, line 4: We have support for targeting .NET Framework, but this looks hard coded to .net6.0. Should there be an if block around this?

src/CoreWCF.Templates/src/CoreWCF.Templates.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <!-- clear SymbolPackFormat: https://github.com/NuGet/Home/issues/10372 -->
+    <IncludeSymbols>false</IncludeSymbols>
+    <SymbolPackageFormat></SymbolPackageFormat>
+  </PropertyGroup>
+  <ItemGroup>
+    <Content Include="templates\**\*" Exclude="templates\**\bin\**;templates\**\obj\**" CopyToOutputDirectory="Always" />
+    <Content Remove="templates\Directory.Build.props"/>
+    <Content Remove="templates\Directory.Build.targets"/>
+    <Content Remove="templates\.editorconfig"/>
+    <Compile Remove="templates\**\*"/>
```
mconnew, line 20: I'm having difficulty understanding some of these. You are removing the .editorconfig, but I thought we had a separate one so that the generated project included an editorconfig?

README.md
```diff
@@ -29,6 +29,20 @@ To keep up to date on what's going on with CoreWCF, you can subscribe to the [an
 * The [Samples](src/Samples/) folder has examples for .NET Framework, .NET Core 3.1, .NET 5 & 6 
 * The [Blog](https://corewcf.github.io/) has details on the design philosophy and a deep dive into the features are included in each new release.
 
+### Use Core WCF project templates (dotnet CLI or VisualStudio)
+
```
mconnew, line 33: There was some ambiguity with the name where it was sometimes referred to as "Core WCF" and other times "CoreWCF". We settled on "CoreWCF" and the rest of the documentation has been updated to reflect that. Can you fix the few references here?

src/CoreWCF.Templates/src/templates/CoreWCFService/Program.Main.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.UseKestrel(options =>
+                    {
+                        options.AllowSynchronousIO = true;
+                    });
```
mconnew, line 28: This isn't needed any more unless you are using streamed transfer mode.

g7ed6e: it dropped it in [32d1a91](https://github.com/CoreWCF/CoreWCF/pull/590/commits/32d1a917a5c8ef67a8b585e09aa7a6b0febd0058)

src/CoreWCF.Templates/src/templates/CoreWCFService/CoreWCFService.csproj
g7ed6e: I double checked this is correctly valued dynamically using the optional `--framework` arg when using dotnet CLI. Using the VS GUI you may also use the drop down menu allowing to choose the target framework. https://github.com/CoreWCF/CoreWCF/pull/590#discussion_r861507711

src/CoreWCF.Templates/src/CoreWCF.Templates.csproj
mconnew: I see my confusion now. I didn't notice there was separate Content and Compile items. The general convention is to not mix different item types in a single ItermGroup. That always confused me why, but I think I just learnt why. Can you move the Compile into it's own ItemGroup to make things clearer.

src/CoreWCF.Templates/src/templates/CoreWCFService/CoreWCFService.csproj
mconnew: Sounds like it's just a placeholder then. That's really non-intuitive. Thanks for checking.

src/CoreWCF.Templates/src/CoreWCF.Templates.csproj
g7ed6e: done. i also added comments

src/CoreWCF.Templates/src/templates/CoreWCFService/CoreWCFService.csproj
g7ed6e: indeed. You can have a look to the integration tests here:
https://github.com/CoreWCF/CoreWCF/blob/3f3e41801daeb787f1cfc86bfc202fcce12fd6b4/src/CoreWCF.Templates/tests/BasicTests.cs
Basically CoreWCF template is installed and then the unit test code is issuing all variations of `dotnet new corewcf`. Then the project is built and ran. Finally an HttpClient ensure it is working appropriately.

src/CoreWCF.Templates/src/templates/CoreWCFService/Program.Main.cs
mconnew: @FreeAndNil, it depends on what your contract looks like. If you are using data types which will use a serializer, then you will need to turn on `AllowSynchronousIO`. This is because we create an XmlWriter which writes directly to the network and pass that to the serializer and tell it to serialize. The serializers only have a synchronous implementation so write to the XmlWriter using synchronous api's only, which means the output stream only gets written to synchronously. If you are using `Stream` in your contract, I believe this doesn't need this option, but I haven't tested it. If it does still need it, I might be able to improve that.

FreeAndNil: We use a stream as response.
We get no error, but after some time  (12 to 30 hours in our development environment) the server does no longer respond to new requests. We see many, many threads when looking at the dump.

I would propose opening a new issue.

mconnew: Please do open a new issue. Are you able to provide a process dump? You can contact me privately on gitter to arrange making a dump available.

PR #600, opened 2022-04-12
Updated release guide with support details

samsp-msft:


Documentation/release-guide.md
```diff
@@ -4,12 +4,76 @@ This document details the strategy for releases and the process to manage them.
...
+- CoreWCF has major and minor versions. Major versions can take breaking changes, minor versions will not (and we will fix accidental breaks in minor versions.)
+  - E.g. 1.0 -> 2.0 is a major version update with breaking changes while 2.0 -> 2.1 is a minor update and no intentional breaking changes.
+
+- Major versions will declare what .NET runtime & platform is supported and may drop support for an older .NET runtime that is still under support. 
+  - E.g. CoreWCF 1.0 may support .NET Core 3.1 and .NET 5.0 while CoreWCF 2.0 may choose to only support .NET 6.0.
+
+- Minor versions always support the same .NET runtime & platform as the parent major, they do not drop support for a .NET runtime that was supported by the parent major as long as the .NET runtime & ASP.NET Core itself is still in support. 
+  - E.g. If CoreWCF  1.0 supported .NET Core 3.1 then Core WCF 1.1 will also support .NET Core 3.1 as long as .NET Core 3.1 remains in support. Core WCF 2.x will follow whatever was declared supported when 2.0 shipped. 
+
+- Minor versions will be supported for 6 months after the successor ships.
```
mjrousos, line 56: We're missing verbiage on how long major versions will be supported. After 3.0 is released, how long will we support 2.x? I think that should be called out more clearly here.

mjrousos (changes requested):
This looks good but I think we need some more detail about how long major versions of CoreWCF will be supported.

mconnew (approved):

PR #601, opened 2022-04-12
Throw InvalidOperationException if Service Model Web ServiceNotRegistered

tallalnparis4ev:
This PR throws an InvalidOperationException, with a message, if the Service Model Web Service is not registered.

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Configuration/ServiceModelWebServiceBuilderExtensions.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Configuration/ServiceModelWebServiceBuilderExtensions.cs
@@ -230,11 +230,7 @@ namespace CoreWCF.Configuration
             builder.AddServiceEndpoint(service, implementedContract, binding, address, listenUri, serviceEndpoint =>
             {
                 KeyedByTypeCollection<IEndpointBehavior> behaviors = (KeyedByTypeCollection<IEndpointBehavior>)serviceEndpoint.EndpointBehaviors;
-                WebHttpBehavior webHttpBehavior = behaviors.Find<WebHttpBehavior>();
-                if (webHttpBehavior is null)
-                {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.ServiceModelWebServiceNotRegistered)));
-                }
+                WebHttpBehavior webHttpBehavior = behaviors.Find<WebHttpBehavior>() ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.ServiceModelWebServiceNotRegistered, address)));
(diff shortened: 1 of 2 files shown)


src/CoreWCF.WebHttp/src/Resources/Strings.resx
```diff
@@ -375,6 +375,9 @@
   <data name="SerializingRequestNotSupportedByFormatter" xml:space="preserve">
     <value>The formatter '{0}' does not support serializing requests.</value>
   </data>
+  <data name="ServiceModelWebServiceNotRegistered" xml:space="preserve">
+    <value>Service Model Web Service not registered, call IServiceCollection.AddServiceModelWebServices() in Startup.cs.</value>
```
mconnew, line 379: This is a little obscure, and not quite accurate. The error message should say what the problem is (WebHttpBehavior endpoint behavior is missing) and suggest a possible solution. Also a Startup class isn't necessary, or it could be called something different. This would be a better error message:
```
Unable to find the endpoint behavior WebHttpBehavior for endpoint address "{0}". This is possibly caused by not calling  ServiceModelWebServiceCollectionExtensions.AddServiceModelWebServices(IServiceCollection services).
```
Then in the code, passing address as a parameter to SR.Format.

src/CoreWCF.WebHttp/src/CoreWCF/Configuration/ServiceModelWebServiceBuilderExtensions.cs
```diff
@@ -231,6 +231,10 @@ public static class ServiceModelWebServiceBuilderExtensions
             {
                 KeyedByTypeCollection<IEndpointBehavior> behaviors = (KeyedByTypeCollection<IEndpointBehavior>)serviceEndpoint.EndpointBehaviors;
                 WebHttpBehavior webHttpBehavior = behaviors.Find<WebHttpBehavior>();
+                if (webHttpBehavior is null)
```
mconnew, line 234: You could use the null coalescing feature here:
```c#
WebHttpBehavior webHttpBehavior = behaviors.Find<WebHttpBehavior>() ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.ServiceModelWebServiceNotRegistered, address)));
```

src/CoreWCF.WebHttp/src/Resources/Strings.resx
```diff
@@ -375,6 +375,9 @@
   <data name="SerializingRequestNotSupportedByFormatter" xml:space="preserve">
     <value>The formatter '{0}' does not support serializing requests.</value>
   </data>
+  <data name="ServiceModelWebServiceNotRegistered" xml:space="preserve">
```
mconnew, line 378: Small typo on the resource name, WebServices should be pluralized. E.g. `ServiceModelWebServicesNotRegistered`

mconnew (approved):

PR #605, opened 2022-04-13
Workaround runtime DataContract bugs

mconnew:
Worked around issue where equivalent data contracts were failing to add. E.g. `IEnumerable<string>` and `List<string>`
Also worked around issue where adding `Dictionary<K,V>` was adding spurious extra schema `KeyValuePair<K,V>`.


mconnew:
I've been thinking about this more and I think I can solve the problem generally and remove the edge cases.

mconnew:
Latest commit now fixes for all types added whether directly or indirectly.

birojnayak (approved):

PR #607, opened 2022-04-13
Updated to properly fetch wsdl and xsd files with a query string value

### LINKED ISSUE #604: WSDL import URLs loading default WSDL instead of correct document
When there are imports in the WSDL, e.g. `<wsdl:import namespace="myns" location="https://localhost:8088/MyService?wsdl=wsdl0" />`, the import URL returns the same document as the default.

It looks like there might be a problem in `ServiceMetadataExtension.HttpGetImpl.FindQuery` when there is a Value. The returned query does not include the value, so it cannot find the correct doc:
```
. . .
if (string.Compare(q.Key, WsdlQueryString, StringComparison.OrdinalIgnoreCase) == 0)
{
    query = q.Key;
}
. . .
```
should be something like:
```
. . .
if (string.Compare(q.Key, WsdlQueryString, StringComparison.OrdinalIgnoreCase) == 0)
{
    query = string.IsNullOrWhiteSpace(q.Value) ? q.Key : $"{q.Key}={q.Value}";
}
. . .
```
This is true for at least `?wsdl=wsdlN` and `?xsd=xsdN`.

It look like the way it's done is slightly different from System.ServiceModel, which passes in the queries, e.g. `"wsdl=wsdl0"`, as a `string[]` and then checks a substring for `"wsdl"`, etc.

roend83:
Fixes #604

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
@@ -760,28 +760,27 @@ namespace CoreWCF.Description
 
             private string FindQuery(IQueryCollection queries)
             {
-                string query = null;
-                foreach (var q in queries)
+                if (queries.TryGetValue(WsdlQueryString, out var wsdlValues))
                 {
-                    if (string.Compare(q.Key, WsdlQueryString, StringComparison.OrdinalIgnoreCase) == 0)
-                    {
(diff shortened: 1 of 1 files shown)


mconnew:
@roend83, we have one test where there's a race condition I haven't been able to track down which causes a 500 response to be returned with a particular type of bad request instead of a SOAP Fault response. I've kicked that run off again and it should pass.

mconnew (approved):

PR #609, opened 2022-04-14
Support for listening on IPv6; fixes #606

### LINKED ISSUE #606: NetTcpBinding won't listen on IPV6
Hi,

we use 

IWebHostBuilder result = WebHost.CreateDefaultBuilder().UseNetTcp(805);

void Configure(IApplicationBuilder app, IHostingEnvironment env)
=> app.UseServiceModel(builder =>
{
  builder.AddService(typeof(MyService));
  builder.AddServiceEndpoint(typeof(MyService), typeof(IMyService), netTcpBinding, "nettcp://<hostname>:805/MyService.svc", "nettcp://<hostname>:805/MyService.svc");
}

After starting the server, it won't listen on IPV6.

Now listening on: http://**0.0.0.0**:805
Application started. Press Ctrl+C to shut down.

When using HTTP we get:
Now listening on: http://**[::]**:805

But the client tries to connect via IPV6 and won't get a response.

Is this a known problem?
How can this be fixed?

Regards.

FreeAndNil:
fixes #606 (NetTcpBinding won't listen on IPV6)

@mconnew could you please review

### DIFF
--- a/src/CoreWCF.ConfigurationManager/tests/IntegrationOfDependencyInjectionTests.cs
+++ b/src/CoreWCF.ConfigurationManager/tests/IntegrationOfDependencyInjectionTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Linq;
+using System.Net;
 using CoreWCF.Configuration;
 using Microsoft.AspNetCore.Hosting;
 using Xunit;
@@ -21,18 +22,26 @@ namespace CoreWCF.ConfigurationManager.Tests
         [Fact]
         public void AddOneEndpointInContainer()
         {
-            IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output).Build();
-            using (host)
-            {
-                host.Start();
-                var resolver = new DependencyResolverHelper(host);
-                IServiceBuilder serviceBuilder = resolver.GetService<IServiceBuilder>();
+            AddOneEndpointInContainerCore(IPAddress.Any, 6687, "net.tcp://0.0.0.0:6687/");
+        }
 
-                Assert.Single(serviceBuilder.Services);
-                Assert.Single(serviceBuilder.BaseAddresses);
-                Assert.Equal(typeof(SomeService).FullName, serviceBuilder.Services.Single().FullName);
-                Assert.Equal("net.tcp://0.0.0.0:6687/", serviceBuilder.BaseAddresses.Single().AbsoluteUri);
-            }
+        [Fact]
+        public void AddOneEndpointInContainerIPv6()
+        {
+            AddOneEndpointInContainerCore(IPAddress.IPv6Any, 6688, "net.tcp://[::]:6688/");
+        }
+
+        private void AddOneEndpointInContainerCore(IPAddress ipAddress, int port, string expectedBaseUri)
+        {
+            IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output, ipAddress, port).Build();
+            host.Start();
+            var resolver = new DependencyResolverHelper(host);
+            IServiceBuilder serviceBuilder = resolver.GetService<IServiceBuilder>();
+
+            Assert.Single(serviceBuilder.Services);
+            Assert.Single(serviceBuilder.BaseAddresses);
+            Assert.Equal(typeof(SomeService).FullName, serviceBuilder.Services.Single().FullName);
+            Assert.Equal(expectedBaseUri, serviceBuilder.BaseAddresses.Single().AbsoluteUri);
         }
     }
 }
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
@@ -101,7 +101,10 @@ namespace CoreWCF.Configuration
             foreach (ListenOptions listenOptions in ListenOptions)
             {
                 IPEndPoint endpoint = listenOptions.IPEndPoint;
-                var baseAddress = new Uri($"net.tcp://{endpoint.Address}:{endpoint.Port}/");
+                string address = endpoint.Address.ToString();
+                if (endpoint.Address.AddressFamily == System.Net.Sockets.AddressFamily.InterNetworkV6)
+                    address = $"[{address}]";
+                var baseAddress = new Uri($"net.tcp://{address}:{endpoint.Port}/");
                 _logger.LogDebug($"Adding base address {baseAddress} to ServiceBuilderOptions");
                 _serviceBuilder.BaseAddresses.Add(baseAddress);
             }
--- a/src/CoreWCF.ConfigurationManager/tests/ServiceHelper.cs
+++ b/src/CoreWCF.ConfigurationManager/tests/ServiceHelper.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Net;
 using CoreWCF.Configuration;
 using Helpers;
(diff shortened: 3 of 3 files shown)


FreeAndNil:
@mconnew are the remaining failed tests my fault?

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder)
 
         public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder, int port)
         {
-            return webHostBuilder.UseNetTcp(IPAddress.Any, port);
+            return webHostBuilder.UseNetTcp(IPAddress.IPv6Any, port);
```
mconnew, line 28: This change is a breaking change and will break some existing users of CoreWCF. If someone doesn't have IPv6 enabled on just one of their clients, your change will break them. You don't need to modify the behavior of the UseNetTcp method though, instead in your service you can use the overload which takes an IPAddress and pass IPv6Any. So you could modify the example code in your original issue comment to this:

```c#
IWebHostBuilder result = WebHost.CreateDefaultBuilder().UseNetTcp(IPAddress.IPv6Any, 805);
```

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
```diff
@@ -101,7 +101,10 @@ private void OnServiceBuilderOpening(object sender, EventArgs e)
             foreach (ListenOptions listenOptions in ListenOptions)
             {
                 IPEndPoint endpoint = listenOptions.IPEndPoint;
-                var baseAddress = new Uri($"net.tcp://{endpoint.Address}:{endpoint.Port}/");
+                string address = endpoint.Address.ToString();
```
mconnew, line 104: This makes sense due to Uri understandable not being able to parse `net.tcp://:::805/` as a valid Uri. 😁

src/CoreWCF.ConfigurationManager/tests/IntegrationOfDependencyInjectionTests.cs
```diff
@@ -21,18 +21,15 @@ public IntegrationOfDependencyInjectionTests(ITestOutputHelper output)
         [Fact]
         public void AddOneEndpointInContainer()
         {
-            IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output).Build();
-            using (host)
-            {
-                host.Start();
-                var resolver = new DependencyResolverHelper(host);
-                IServiceBuilder serviceBuilder = resolver.GetService<IServiceBuilder>();
+            using IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output).Build();
```
mconnew, line 24: As you won't be modifying the UseNetTcp method, this test will need to be reverted. You could add a new test which explicitly passes `IPAddress.IPv6Any` to CoreWCF to make sure it continues to work in the future with the base address fix.

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
FreeAndNil: My observation is that the server listens on IPv4 and IPv6 (as does Legacy WCF) when using AnyIPV6.
Can you confirm this?

mconnew: Are you sure about Legacy WCF? I think when you use TCP port sharing it might, but I just checked [the code](https://github.com/microsoft/referencesource/blob/master/System.ServiceModel/System/ServiceModel/Channels/SocketConnection.cs#L2105-L2113) and it doesn't set the option to use both IPv4 and IPv6. To make it listen on both, you need to set `Socket.DualMode` to true, which then sets an underlying socket option. You can see where this is done in .NET 6 [here](https://github.com/dotnet/aspnetcore/blob/5b1238f1f5f4485e85790ce57de5208790129022/src/Servers/Kestrel/Transport.Sockets/src/SocketTransportOptions.cs#L116-L122).

The thing I'm concerned about is the scenario where the OS doesn't support IPv4 and you use DualMode with IPv6Any. I can't find any clear documentation that says what the behavior is but I found some code in ASP.NET Core which suggests it will fail. The [code is here](https://github.com/dotnet/aspnetcore/blob/5b1238f1f5f4485e85790ce57de5208790129022/src/Servers/Kestrel/Core/src/AnyIPListenOptions.cs#L23-L34), I'm pasting it below for convenience:
```c#
        try
        {
            await base.BindAsync(context, cancellationToken).ConfigureAwait(false);
        }
        catch (Exception ex) when (!(ex is IOException))
        {
            context.Logger.LogDebug(CoreStrings.FormatFallbackToIPv4Any(IPEndPoint.Port));

            // for machines that do not support IPv6
            EndPoint = new IPEndPoint(IPAddress.Any, IPEndPoint.Port);
            await base.BindAsync(context, cancellationToken).ConfigureAwait(false);
        }
```
This is in the class `AnyIPListenOptions` which initializes the base class to use `IPAddress.IPv6Any`. You can see here that an exception can be thrown when the OS doesn't support IPv6, in which case it retries with `IPAddress.Any` to only bind to IPv4 addresses. Your proposed change is a breaking change. If someone is using UseNetTcp with just a port number to  have CoreWCF listen on all IPv4 addresses and are running on an OS which doesn't support IPv6 (not uncommon when using VM's) and they upgrade to use a package with your proposed change, their service will fail to start.  

Basically it comes down to two options.
1. Accept your change and break some people who were previously working and now have to modify their code to explicitly specify `IPAddress.Any`  
2. Don't accept your change, and if you want to listen on all IPv4 and IPv6 addresses, you need to pass `IPAddress.IPv6Any`  

Breaking existing users between releases has a very high bar (generally only acceptable for security fixes) and changing the default which will break some users in order to save other users having to pass an extra parameter doesn't meet that bar.  

I did consider suggesting using `Socket.OSSupportsIPv6` but it seems strange that the ASP.NET Core team didn't use that instead of the try/catch approach which leads me to think there might be scenarios where `OSSupportsIPv6` returns true and it still fails so I'm not comfortable with that approach this close to the final build of 1.0. Once we switch to developing CoreWCF 2.0 (which will be soon after the 1.0 final release) that might be an option as it will have a long amount of time before we release 2.0 so plenty of opportunity for people to tell us it broke them. We also accept reasonable breaking changes between major versions and the risk that the `OSSupportsIPv6` approach would bring feels like it would be okay to do in 2.0.

mconnew (changes requested):
Need to revert the behavior change for UseNetTcp as it's a breaking change.

src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
FreeAndNil: Thank you very much for your detailed explanation. I will adjust the code as proposed tomorrow.

FreeAndNil: I've changed the code as requested.

src/CoreWCF.ConfigurationManager/tests/IntegrationOfDependencyInjectionTests.cs
FreeAndNil: Test added

mconnew (approved):

PR #616, opened 2022-04-18
Fixes to support WSDL generation after updating to preview 2

### LINKED ISSUE #615: Errors Generating WSDL in Preview2 Package in .NET 6
The errors are thrown in DataContractEx.GetGetOnlyCollectionDataContractStub.  The first error I got was here:

`var serializationModeParam = Expression.Constant(intModeParam, SerializationModeType); `

The message is:

`System.ArgumentException
  HResult=0x80070057
  Message=Argument types do not match
  Source=System.Linq.Expressions
  StackTrace:
   at System.Linq.Expressions.Expression.Constant(Object value, Type type)
   at CoreWCF.Runtime.Serialization.DataContractEx.GetGetOnlyCollectionDataContractStub(Int32 id, RuntimeTypeHandle typeHandle, Type type, Int32 mode) in C:\repos\inspection\apsl\Apsl\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\Runtime\Serialization\DataContractEx.cs:line 121`

I can fix this by changing it to:

`var serializationModeParam = Expression.Constant(Enum.ToObject(SerializationModeType, mode), SerializationModeType); `

However then I get an error on this line:

`Expression callExpr = Expression.Call(methodInfo, idParam, typeHandleParam, typeParam, serializationModeParam);`

The message is: 

`System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null. (Parameter 'method')
  Source=System.Linq.Expressions
  StackTrace:
   at System.Linq.Expressions.Expression.Call(MethodInfo method, Expression arg0, Expression arg1, Expression arg2, Expression arg3)
   at CoreWCF.Runtime.Serialization.DataContractEx.GetGetOnlyCollectionDataContractStub(Int32 id, RuntimeTypeHandle typeHandle, Type type, Int32 mode) in C:\repos\inspection\apsl\Apsl\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\Runtime\Serialization\DataContractEx.cs:line 124`

It appears that the `GetGetOnlyCollectionDataContract` method does not exist on the type `System.Runtime.Serialization.DataContract, System.Private.DataContractSerialization, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`.

roend83:
Fixes #615

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
@@ -113,12 +113,12 @@ namespace CoreWCF.Runtime.Serialization
         private static DataContractEx GetGetOnlyCollectionDataContractStub(int id, RuntimeTypeHandle typeHandle, Type type, int mode)
         {
             var methodInfo = DataContractType.GetMethod("GetGetOnlyCollectionDataContract", BindingFlags.Static | BindingFlags.NonPublic, null,
-                new Type[] { typeof(int), typeof(RuntimeTypeHandle), typeof(Type),  }, null);
+                new Type[] { typeof(int), typeof(RuntimeTypeHandle), typeof(Type), SerializationModeType }, null);
             var idParam = Expression.Parameter(typeof(int), "id");
             var typeHandleParam = Expression.Parameter(typeof(RuntimeTypeHandle), "typeHandle");
             var typeParam = Expression.Parameter(typeof(Type), "type");
             var intModeParam = Expression.Parameter(typeof(int), "mode");
-            var serializationModeParam = Expression.Constant(intModeParam, SerializationModeType); 
+            var serializationModeParam = Expression.Constant(Enum.ToObject(SerializationModeType, mode), SerializationModeType); 
 
             // Passing null as instance expression as static method call
             Expression callExpr = Expression.Call(methodInfo, idParam, typeHandleParam, typeParam, serializationModeParam);
--- a/src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
@@ -118,7 +118,7 @@ namespace CoreWCF.Runtime.Serialization
             var typeHandleParam = Expression.Parameter(typeof(RuntimeTypeHandle), "typeHandle");
             var typeParam = Expression.Parameter(typeof(Type), "type");
             var intModeParam = Expression.Parameter(typeof(int), "mode");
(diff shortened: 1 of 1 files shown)


mconnew:
Thank you for this. I really don't like this type of code living in the code base because it's so easy to make a small mistake like this and really hard to catch in code review. I'm expecting a few more cases here and there after we release so I'm expecting to do some patch releases with more wsdl fixes. For now this is planned to go into the final GA release unless we get enough other things need fixing this week.

src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
```diff
@@ -113,12 +113,12 @@ private static DataContractEx GetDataContractStub(Type clrType)
...
         {
             var methodInfo = DataContractType.GetMethod("GetGetOnlyCollectionDataContract", BindingFlags.Static | BindingFlags.NonPublic, null,
-                new Type[] { typeof(int), typeof(RuntimeTypeHandle), typeof(Type),  }, null);
+                new Type[] { typeof(int), typeof(RuntimeTypeHandle), typeof(Type), SerializationModeType }, null);
             var idParam = Expression.Parameter(typeof(int), "id");
             var typeHandleParam = Expression.Parameter(typeof(RuntimeTypeHandle), "typeHandle");
             var typeParam = Expression.Parameter(typeof(Type), "type");
             var intModeParam = Expression.Parameter(typeof(int), "mode");
-            var serializationModeParam = Expression.Constant(intModeParam, SerializationModeType); 
+            var serializationModeParam = Expression.Constant(Enum.ToObject(SerializationModeType, mode), SerializationModeType);
```
mconnew, line 121: This isn't quite the correct solution. The first time this method gets called, it will take the passed in mode value and convert it to SerializationModeType and store that as a constant and always use the same value.  This happens to work because we only pass in the one `SerializationModeType` value in a single call point so it happens to always be that one same value, but this is a potential future bug waiting to be tripped up on. The original code was wrong as I misread what the intellisense auto complete was offering. It should be `Expression.Convert` and not `Expression.Constant`.  
```suggestion
            var serializationModeParam = Expression.Convert(intModeParam, SerializationModeType);
```

mconnew (changes requested):
Use `Expression.Convert` to case to Enum so that the passed in value always gets used.

src/CoreWCF.Primitives/src/CoreWCF/Runtime/Serialization/DataContractEx.cs
roend83: This makes a lot more sense.  I was wondering why it was just using constant.  I've updated the PR.

mconnew (approved):

Issue #595, opened 2022-04-07
Implement a compression format option for binary messages sent over a TCP transport


glsolaria:
I get an NotSupportedException when I try to enable GZip compression on a NetTcpBinding ...
```csharp
var netTcpBinding = new NetTcpBinding();
netTcpBinding.Security.Mode = SecurityMode.Transport;
netTcpBinding.Security.Transport.ClientCredentialType = TcpClientCredentialType.Certificate;
                    
var customBinding = new CustomBinding(netTcpBinding);

var binaryEncoding = customBinding.Elements.Find<BinaryMessageEncodingBindingElement>();
binaryEncoding.CompressionFormat = CompressionFormat.GZip;
```

> Unhandled exception. System.NotSupportedException: The transport configured on this binding does not appear to support the CompressionFormat specified (GZip) on the message encoder.  To resolve this issue, set the CompressionFormat on the BinaryMessageEncodingBindingElement to 'None' or use a different transport. 
> at CoreWCF.Channels.BinaryMessageEncodingBindingElement.VerifyCompression(BindingContext context) in C:\corewcf\ReferenceSource\CoreWCF-1.0.0-preview1\src\CoreWCF.Primitives\src\CoreWCF\Channels\BinaryMessageEncodingBindingElement.cs:line 163

jonlouie:
I'm interested in helping with this issue

birojnayak:
@jonlouie assigning to you

PR #622, opened 2022-04-21
Enable GZip and Deflate compression for binary messages over TCP

jonlouie:
### Description
Closes: #595 

This PR allows server bindings to be configured for GZip and Deflate compression formats for binary messages over TCP.

---

### Testing
* Unit tests show end-to-end communication works as expected
* Manual inspection of network activity showed the messages sent in the unit tests were in fact compressed:
Uncompressed message: `new string('a', 10000)  // string of length 10,000`

Compression | Message size
--- | --- 
Uncompressed | 10198 bytes
GZip |  233 bytes
Deflate | 216 bytes

---

### Concerns/Follow-up questions:
* ~This PR only enables recognition of GZip and Deflate compression formats if the content type is `application/soap+msbinsession1`. It was the only content type I was able to figure out how to test so the content type of `application/soap+msbin1` was left out. Are there any concerns leaving this content type unsupported? If so, what is the best way to test this case?~

### DIFF
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingDecoder.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingDecoder.cs
@@ -386,6 +386,14 @@ namespace CoreWCF.Channels.Framing
                     return FramingEncodingString.Binary;
                 case FramingEncodingType.BinarySession:
                     return FramingEncodingString.BinarySession;
+                case FramingEncodingType.ExtendedBinaryGZip:
+                    return FramingEncodingString.ExtendedBinaryGZip;
+                case FramingEncodingType.ExtendedBinarySessionGZip:
+                    return FramingEncodingString.ExtendedBinarySessionGZip;
+                case FramingEncodingType.ExtendedBinaryDeflate:
+                    return FramingEncodingString.ExtendedBinaryDeflate;
+                case FramingEncodingType.ExtendedBinarySessionDeflate:
+                    return FramingEncodingString.ExtendedBinarySessionDeflate;
                 default:
                     return "unknown" + ((int)type).ToString(CultureInfo.InvariantCulture);
             }
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingEncoder.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingEncoder.cs
@@ -130,7 +130,23 @@ namespace CoreWCF.Channels.Framing
 
         public static EncodedContentType Create(string contentType)
         {
-            if (contentType == FramingEncodingString.BinarySession)
+            if (contentType == FramingEncodingString.ExtendedBinarySessionDeflate)
+            {
+                return new EncodedContentType(FramingEncodingType.ExtendedBinarySessionDeflate);
+            }
+            else if (contentType == FramingEncodingString.ExtendedBinaryDeflate)
+            {
+                return new EncodedContentType(FramingEncodingType.ExtendedBinaryDeflate);
+            }
+            else if (contentType == FramingEncodingString.ExtendedBinarySessionGZip)
+            {
+                return new EncodedContentType(FramingEncodingType.ExtendedBinarySessionGZip);
+            }
+            else if (contentType == FramingEncodingString.ExtendedBinaryGZip)
+            {
+                return new EncodedContentType(FramingEncodingType.ExtendedBinaryGZip);
+            }
+            else if(contentType == FramingEncodingString.BinarySession)
             {
                 return new EncodedContentType(FramingEncodingType.BinarySession);
             }
--- a/src/CoreWCF.NetTcp/tests/CompressionFormatTests.cs
+++ b/src/CoreWCF.NetTcp/tests/CompressionFormatTests.cs
@@ -1,0 +1,132 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using CoreWCF.Channels;
+using CoreWCF.Configuration;
+using Helpers;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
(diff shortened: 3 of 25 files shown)


src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingDecoder.cs
```diff
@@ -386,6 +386,14 @@ public static string GetString(FramingEncodingType type)
                     return FramingEncodingString.Binary;
                 case FramingEncodingType.BinarySession:
                     return FramingEncodingString.BinarySession;
+                //case FramingEncodingType.ExtendedBinaryGZip:
+                //    return FramingEncodingString.ExtendedBinaryGZip;
```
mconnew, line 390: I believe these commented out variations are needed to support compression when using TransferMode Streamed as that doesn't maintain a session between requests. So I think you need to include the non-session variants too.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingEncoder.cs
```diff
@@ -130,7 +130,23 @@ private EncodedContentType(string contentType)
 
         public static EncodedContentType Create(string contentType)
         {
-            if (contentType == FramingEncodingString.BinarySession)
+            if (contentType == FramingEncodingString.ExtendedBinarySessionDeflate)
+            {
+                return new EncodedContentType(FramingEncodingType.ExtendedBinarySessionDeflate);
+            }
+            //else if (contentType == FramingEncodingString.ExtendedBinaryDeflate)
+            //{
```
mconnew, line 138: Same comment as previous, we need to non-session variations too.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingDecoder.cs
jonlouie: Thanks for pointing out TransferMode.Streamed. I've uncommented those variations and added appropriate test cases.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/Framing/FramingEncoder.cs
jonlouie: Uncommented these non-session variants as well

jonlouie:
It looks like this test is failing for `Debug Windows_net5.0` and `Release Linux_net5.0`: CoreWCF.NetTcp.Tests.TransportWithMessageCredentialNetTCPTest.NetTCPRequestReplyWithTransportMessageEchoStringDemuxFailure

@mconnew Any idea if this is a flaky test? I'm not able to reproduce the failure locally.

mconnew:
It's a flakey test.

mconnew:
I'm really starting to get annoyed by that flakey test. I can't reproduce it locally though as my computer is too fast. It's a race condition on a failure case causing the server to return a 500 error instead of the correct fault message. It's a rare edge case for a very narrow scenario and it's causing the wrong failure on the client but isn't causing a passing scenario to fail so I haven't put the time in to try really hard to reproduce locally (need to set up a VM and artificially slow it down). So for now I just rerun the test and it usually passes the next time.

src/CoreWCF.NetTcp/tests/CompressionFormatTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                var result = channel.EchoString(testString);
+                Assert.Equal(testString, result);
+                ((System.ServiceModel.IClientChannel)channel).Close();
+            }
+        }
+
+        public static IEnumerable<object[]> GetTestVariations()
+        {
+            yield return new object[]
+            {
```
mconnew, line 62: You can simplify this by using loops. You can also cast an enum between types as they are backed by an integer. You could do something like this:
```c#
foreach (CompressionFormat compression in Enum.GetValues(typeof(CompressionFormat)))
{
    foreach (TransferMode transferMode in Enum.GetValues(typeof(TransferMode))
    {
        yield return new object[]
        {
                MessageVersion.Soap12WSAddressing10,
                System.ServiceModel.Channels.MessageVersion.CreateVersion(System.ServiceModel.EnvelopeVersion.Soap12, System.ServiceModel.Channels.AddressingVersion.WSAddressing10),
                compression,
                (System.ServiceModel.Channels.CompressionFormat)compression,
                transferMode,
                (System.ServiceModel.TransferMode)transferMode
        };
        yield return new object[]
        {
                MessageVersion.Soap12WSAddressingAugust2004,,
                System.ServiceModel.Channels.MessageVersion.Soap12WSAddressingAugust2004,
                compression,
                (System.ServiceModel.Channels.CompressionFormat)compression,
                transferMode,
                (System.ServiceModel.TransferMode)transferMode
        };
    }
}
```
I haven't tested this code so you might need to cast to int first when casting between the CoreWCF and the SM variations, but I think the above will work. Also GetValues returns an Array which has a signature of returning objects. I think the foreach does an implicit cast but it might need the iterator variable to be type object and then cast where it's used. I believe it should work as written, but there's a chance you might need to make these changes.

jonlouie: Thanks for the suggestion- I've added it in my latest commit. Looks great, worked perfectly, and increased the test coverage from 8 to 24 cases.

mconnew:
@jonlouie, this looks good now. Do you want to rebase and squash these down to a single commit before I merge or should I squash and merge it on github? The difference is it tends to loose the authorship if I squash and merge it on github. It ends up as part of the commit comment and not on the commit metadata so I'm not sure if it shows up on your github activity graph. I believe it doesn't.

birojnayak:
> It looks like this automated test is failing for `Debug Windows_net5.0` and `Release Linux_net5.0`: CoreWCF.NetTcp.Tests.TransportWithMessageCredentialNetTCPTest.NetTCPRequestReplyWithTransportMessageEchoStringDemuxFailure
> 
> @mconnew Any idea if this is a flaky test? I'm not able to reproduce the failure locally.

it's flaky @jonlouie

jonlouie:
Closing this PR to replace it with #627

Issue #621, opened 2022-04-20
SecurityTokenManager internal abstract method prevents implementation of class for custom tokens


gougeman:
I'm prototyping a port of WCF service to Core WCF 1.0.0-preview2, and it uses a custom token type, which then has a custom derived type for TokenSerializer, ServiceCredentials etc.

The CoreWCF equivalent of SecurityTokenManager seems to have an internal abstract method, and this blocks derivation from outside the CoreWCF assembly.

Is support for custom tokens targeted for part of the 1.0.0 feature set? Looks like most of the other security binding stuff is....

```
namespace CoreWCF.IdentityModel.Selectors
{
    public abstract class SecurityTokenManager
    {
        internal abstract SecurityTokenSerializer CreateSecurityTokenSerializer(SecurityTokenVersion version);
    }
}
```

mconnew:
This looks to be a bug. This should be public. Any chance you can submit a quick PR within the next 24 hours to fix that? If so, it will make it into 1.0.0. We haven't explicitly targeted that scenario, but I would expect it to work. If it doesn't, we can fix whatever is missing for a 1.1 release if extra API's are needed to be exposed, otherwise it would be possible to do in a patch release.

gougeman:
Thanks for the quick response, Matt.  

I have a PR from changes I made locally to work on my scenario with custom tokens,, but it looks like to enable it E2E I need ServiceCredentials.CreateSecurityTokenManager exposed also so that my custom credentials can create the right token manager. 

I have no context on where this area of the project is w.r.t to test coverage of these methods, but I'll submit the PR and you can decide whether it works for 1.0.0.0 or 1.0.1.0

PR #630, opened 2022-04-27
Add support for custom tokens and credentials by exposing internal methods

gougeman:
Add support for custom tokens and credentials by exposing internal methods
 - Change ServiceCredentials.CreateSecurityTokenManager from internal to public to allow custom credential types
 - Change SecurityTokenManager.CreateSecurityTokenSerializer from internal to public to allow custom serializers.
 - Adjust derived classes to match.

No functionality changes in this PR. 

#621

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceCredentials.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceCredentials.cs
@@ -109,7 +109,7 @@ namespace CoreWCF.Description
             return new ServiceCredentials();
         }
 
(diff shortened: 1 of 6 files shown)


mconnew:
This is one of the rare cases where unit tests aren't needed. As you are aware as you had to make visibility changes in them, the extensibility of this type is used by internal types in CoreWCF. This means we are indirectly testing the ability to provide a custom implementation via testing those other features. I'm going to accept this without tests.

gougeman:
Yeah, I figured it was pretty extensively tested already via internal class tests. My tests run timed out and it looks like I don't have rights to the resources to re-run them. Thanks for re-running.

PR #633, opened 2022-04-28
Adding a getting started guide

samsp-msft:
Re-using content from the blog post


Documentation/Walkthrough.md
```diff
@@ -0,0 +1,206 @@
...
+Using the package Manager console, add: 
+
+- CoreWCF.Primitives
+- CoreWCF.Http
+
+![Package Manager Console](images/nuget.png)
+
+*Command Line:*
+
+Edit the project file and add:
```
mconnew, line 32: For command line you can do:
```
dotnet add package CoreWCF.Primitives --version 1.0.0
dotnet add package CoreWCF.Http --version 1.0.0
```

I think you could miss out the version and it will grab the latest version which might be better.

Documentation/Walkthrough.md
```diff
@@ -0,0 +1,206 @@
...
+
+```xml
+<ItemGroup>
+  <PackageReference Include="CoreWCF.Primitives" Version="1.0.0" />
+  <PackageReference Include="CoreWCF.Http" Version="1.0.0" />
+</ItemGroup>
+```
+### 1. Create the Service Contract and Data Contract definitions
+
+These are defined the same as with WCF. When modernizing projects, this code can remain largely unchanged.
```
mconnew, line 42: Would it be worth mentioning that if you need to reuse your contracts for server and client in a shared assembly, that CoreWCF supports using the System.ServiceModel versions of the attributes that can be found in the System.ServiceModel.Primitives client package? Or is that over complicating things?

Documentation/Walkthrough.md
```diff
@@ -0,0 +1,206 @@
...
+{
+    [DataContract]
+    public class EchoFault
+    {
+        [AllowNull]
+        private string _text;
+
+        [DataMember]
+        [AllowNull]
+        public string Text
```
mconnew, line 61: Why not use an auto property?

README.md
```diff
@@ -4,6 +4,8 @@ Core WCF is a port of Windows Communication Foundation (WCF) to .NET Core. The g
 
 ### Package Status
 
+The latest released packages can be found at Nuget.org:
+
```
mconnew, line 8: Could you update the table with the WebHttp package as I realized it's missing.

Documentation/Walkthrough.md
mconnew: It's in the video you linked to at the end of your blog and is in a blog post.

Documentation/Walkthrough.md
samsp-msft: I think I stole this code from the samples.

PR #634, opened 2022-04-29
Cleaning up and simplifying the samples

samsp-msft:


mconnew (approved):

g7ed6e (commented):
Is Newtonsoft really used ?

src/Samples/DesktopServer/Program.cs
```diff
@@ -1,31 +1,32 @@
 ﻿using System;
 using System.Linq;
 using System.ServiceModel;
-using CoreWCF.Samples.StandardCommon;
 
 namespace DesktopServer
```
g7ed6e, line 5: DesktopServer could be renamed to NetFrameworkCoreWCFServer

src/Samples/WsHttpBindings/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            IWebHost host = CreateWebHostBuilder(args).Build();
+            host.Run();
+        }
+
+        // Listen on 8088 for http, and 8443 for https, 8089 for NetTcp.
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+            .UseKestrel(options => {
+                options.ListenLocalhost(WSHttpUserPassword.HTTP_PORT);
```
g7ed6e, line 22: Isn’t AllowSynchronousIO flag missing here ?

src/Samples/NetCoreServer/Program.cs
```diff
@@ -15,25 +15,21 @@ static void Main(string[] args)
             host.Run();
         }
 
+        // Listen on 8088 for http, and 8443 for https, 8089 for NetTcp.
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
             .UseKestrel(options => {
-                options.ListenLocalhost(8088);
-                options.ListenLocalhost(8443, listenOptions =>
+                options.ListenLocalhost(Startup.HTTP_PORT);
```
g7ed6e, line 22: AllowSynchronousIO flag should be set here also I think.

src/Samples/NetCoreServer/Properties/launchSettings.json
```diff
@@ -1,27 +1,12 @@
 {
-  "iisSettings": {
```
g7ed6e, line 2: I would be great to have an IIS sample with net.tcp binding

src/Samples/WsHttpBindings/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+
+namespace NetCoreServer
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            IWebHost host = CreateWebHostBuilder(args).Build();
```
g7ed6e, line 14: on NetCoreServer we could use the netcoreapp3.1/net5.0 variation which uses the generic host.

src/Samples/NetCoreServer/Program.cs
mconnew: I'm digging in to that now to find the conditions when that's the case. We only synchronously read in 2 situations.  
1. Using StreamedMode transfer
2. Using BufferedMode and the client does NOT send a ContentLength header. We read a single byte of the request body to see if there is a request body. I've almost got a fix ready where AllowSynchronousIO only needs to be set when using StreamedMode.

src/Samples/WsHttpBindings/WSHttpUserPassword.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        private static readonly TimeSpan s_debugTimeout = TimeSpan.FromMinutes(20);
+
+        private static CoreWCF.Channels.Binding ApplyDebugTimeouts(CoreWCF.Channels.Binding binding)
+        {
+            if (System.Diagnostics.Debugger.IsAttached)
+            {
+                binding.OpenTimeout =
+                    binding.CloseTimeout =
```
mconnew, line 47: I know this was copied, but the indentation is inconsistent.

src/Samples/NetCoreServer/Properties/launchSettings.json
mconnew: That doesn't work yet. That depends TCP port sharing which needs client side named pipes support to tell SMSvcHost to listen on your behalf, and also needs WAS support to tell SMSvcHost to launch the IIS worker process.

src/Samples/DesktopServer/Program.cs
mconnew: Is the term NetFx known and understood outside of Microsoft?

src/Samples/WsHttpBindings/Program.cs
samsp-msft: Yes - Is it needed for all bindings or just specific ones?

src/Samples/NetCoreClient/NetCoreClient.csproj
```diff
@@ -12,9 +12,9 @@
     <PackageReference Include="System.ServiceModel.Primitives" Version="4.8.1" />
   </ItemGroup>
 
-  <ItemGroup>
+  <!--<ItemGroup>
```
samsp-msft, line 15: Needs to be removed.

src/Samples/NetCoreClient/Program.cs
```diff
@@ -1,30 +1,191 @@
...
 
             Console.Title = "WCF .Net Core Client";
-            Settings settings = ClientLogic.BuildClientSettings(hostname);
+            PrintScenarioList();
 
-            static void log(string value) => Console.WriteLine(value);
-            ClientLogic.InvokeEchoServiceUsingWcf(settings, log);
+            while (true)
+            {
+                Console.WriteLine("Type scenario number, ? for scenario list or Hit enter to exit");
```
samsp-msft, line 26: @mconnew, @g7ed6e  The previous form of this sample was to me a bit overly convoluted. Does it make it more readable (and directly stealable from) to break it up into self contained scenarios?
Should each of them be split into a file per binding?

src/Samples/NetCoreClient/Program.cs
```diff
@@ -1,30 +1,191 @@
...
+            factory.Open();
+            try
+            {
+                IEchoService client = factory.CreateChannel();
+                channel = client as IClientChannel;
+
+                var result = client.FailEcho("Hello Fault!");
+                channel.Close();
+            }
+            catch (FaultException e)
```
samsp-msft, line 107: @mconnew  - I am not getting a FaultException<EchoFail> here, which is what the old code was expecting. Why?

src/Samples/NetCoreServer/Program.cs
samsp-msft: When you have the answer, I will fix these up to be consistent.

src/Samples/dotnet6server/Program.cs
```diff
@@ -1,24 +1,42 @@
...
+app.Urls.Add(BASE_HTTPS_URL);
+
 app.UseServiceModel(builder =>
 {
-    builder.AddService<EchoService>()
-    .AddServiceEndpoint<EchoService, IEchoService>(new BasicHttpBinding(), "/EchoService/basichttp");
+    builder.AddService<EchoService>((serviceOptions) =>
+    {
+        // Set the base addressrs that will be used for the service and acts as the WSDL endpoint 
+        serviceOptions.BaseAddresses.Add(new Uri($"{BASE_HTTP_URL}/EchoService"));
```
samsp-msft, line 30: @mconnew - The way that asp.net core works means that if you configure URLs via config or the cmd line, you don't know what base will be. That's why I added them explicitly. It would be better if the base address could be set using a relative path to the host, and not need to know the full uri here.

samsp-msft:
> Is Newtonsoft really used ?

Yes, and is probably the right approach for this audience rather than System.Text.Json.

src/Samples/WsHttpBindings/Program.cs
samsp-msft: Can you point me to what you mean?

g7ed6e: sure. please see below
```c#
public class Program
{
    public static void Main(string[] args) => CreateHostBuilder(args).Build().Run();
    
    public static IHostBuilder CreateHostBuilder(string[] args) =>
        Host.CreateDefaultBuilder(args)
            .ConfigureWebHostDefaults(webBuilder =>
            {
                webBuilder.UseKestrel(options =>
                {
                    options.AllowSynchronousIO = true;
                });
                webBuilder.UseStartup<Startup>();
            });
}
```

src/Samples/NetCoreClient/Program.cs
g7ed6e: i think samples should :
- contain only the required code to required for its purpose
- allow user to copy/paste and get started

i would go with several folders each containing a working project.

src/Samples/dotnet6server/Program.cs
g7ed6e: @samsp-msft i think that you can simply let the runtime get the urls from `launchsettings.json` and/or env variable or `--urls` argument. Please see https://github.com/g7ed6e/CoreWCF/blob/feature/project-template/src/CoreWCF.Templates/src/templates/CoreWCFService/Program.cs

src/Samples/dotnet6server/Program.cs
```diff
@@ -1,24 +1,42 @@
 using CoreWCF;
 using CoreWCF.Configuration;
+using CoreWCF.Channels;
+using CoreWCF.Description;
 using dotnet6Server;
 using MyContracts;
 
+const string BASE_HTTP_URL = "http://localhost:5000";
```
alexey-kostylev, line 8: what about running on Azure? Is it possible to avoid hardcoded urls and use a startdard way of asp.net? At the moment I'm having my project running locally but it does not start on azure. I suspect because I don't set urls

samsp-msft: The BaseURL is a bit funky. The host:port is only used for the generation of wsdl, specifying the address of the endpoint. The path is critical as that specifies a base address that is prepended to the paths as part of the ServiceEndpoint registration.
The problem is that at this part of the process activation, ASP.NET hasn't figured out what its listening on yet, so we can't ask it, or I'd have put that in the sample.

That's all why this PR is not yet merged.

What URLs have you told kestrel to listen on? Do you have a Urls specified in app.config or via a command line option?

PR #637, opened 2022-04-29
Fix TimeSpan format of CustomBinding UT relying on en-US

g7ed6e:


mconnew:
We should probably find a way to run at least one of our CI test run variations with different cultures to catch this sort of thing.

PR #643, opened 2022-05-02
Fetch a DataProtector from DI in DataProtectionSecurityStateEncoder

### LINKED ISSUE #624: Modify DataProtectionSecurityStateEncoder to use DataProtection instead of ProtectedData
We switched the CookieTransform to use DataProtection as it works cross platform, but we still are using ProtectedData in DataProtectionSecurityStateEncoder. We need to get the DataProtection instance from DI and plumb it through.  

@birojnayak, it would be great to get this done before GA. If there isn't time, I'd like to make DataProtectionSecurityStateEncoder be internal so we can modify it later without any API breakage.

g7ed6e:
Resolves #624

### DIFF
--- a/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
+++ b/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
@@ -230,8 +230,8 @@ namespace WSHttp
         }
 #endif
 
-        [Fact(Skip="Temporarily failing in Azure Devops due to corrupted IIS Express dev certificate in test environment image."),
-            Description("transport-security-with-windows-authentication-httpsys")]
+        [Fact]
+        [Description("transport-security-with-windows-authentication-httpsys")]
         [Trait("Category", "WindowsOnly")]  // HttpSys not supported on Linux
 #if NET5_0_OR_GREATER
         [System.Runtime.Versioning.SupportedOSPlatform("windows")]
@@ -284,8 +284,8 @@ namespace WSHttp
             }
         }
 
-        [Fact(Skip = "Failing in Azure Devops due to corrupted IIS Express dev certificate"),
-            Description("no-security-with-an-anonymous-client-using-impersonation-httpsys")]
+        [Fact]
+        [Description("no-security-with-an-anonymous-client-using-impersonation-httpsys")]
         [Trait("Category", "WindowsOnly")]  // HttpSys not supported on Linux
 #if NET5_0_OR_GREATER
         [System.Runtime.Versioning.SupportedOSPlatform("windows")]


src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
```diff
@@ -3,64 +3,26 @@
...
-        {
-            StringBuilder result = new StringBuilder();
-            result.Append(GetType().ToString());
-            result.AppendFormat("{0}  UseCurrentUserProtectionScope={1}", Environment.NewLine, UseCurrentUserProtectionScope);
-            result.AppendFormat("{0}  Entropy Length={1}", Environment.NewLine, (_entropy == null) ? 0 : _entropy.Length);
-            return result.ToString();
-        }
+        private static IDataProtector GetDataProtectorFromDI()
+            => OperationContext.Current.InstanceContext.Extensions.Find<IServiceProvider>().GetDataProtector(nameof(DataProtectionSecurityStateEncoder));
+
```
mconnew, line 18: This won't work unless the service instance has been added to DI. The `Find<IServiceProvider>()` will throw null in that case.

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
```diff
@@ -3,64 +3,26 @@
...
-        public DataProtectionSecurityStateEncoder() : this(true)
-        {
-            // empty
-        }
-
-        public DataProtectionSecurityStateEncoder(bool useCurrentUserProtectionScope) : this(useCurrentUserProtectionScope, null)
-        { }
-
-        public DataProtectionSecurityStateEncoder(bool useCurrentUserProtectionScope, byte[] entropy)
-        {
```
mconnew, line 24: Removing a public api is a breaking change and would need moving to a 2.0 release. Even though the data is called entropy, it wasn't actually used as Entropy by ProtectedData, it was just a way to make the encryption unique and prevent other code running on the same machine/user from being able to decode data encoded in other places. There are some security attacks which target that scenario. This entropy acts in a similar way to what is passed to DataProtector as a purpose. I think we can support this api by converting the byte array into a base64 encoded string and using that as the purpose.

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
```diff
@@ -3,64 +3,26 @@
...
-            }
-            else
-            {
-                _entropy = Fx.AllocateByteArray(entropy.Length);
-                Buffer.BlockCopy(entropy, 0, _entropy, 0, entropy.Length);
-            }
-        }
-
-        public bool UseCurrentUserProtectionScope { get; }
+        public DataProtectionSecurityStateEncoder() =>  _dataProtector = new Lazy<IDataProtector>(GetDataProtectorFromDI);
```
mconnew, line 15: This needs to stay, but can always return false.

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
```diff
@@ -3,64 +3,26 @@
...
 
-        public byte[] GetEntropy()
-        {
```
mconnew, line 40: If the overload wasn't used where entropy was passed in, then we could return the purpose we used encoded as UT8. Ie Encoding.UTF8.GetBytes(purposeString).

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
mconnew: While it's really unlikely, we need to avoid the possibility of accidental naming collisions. If you use the fully qualified name of this class, the only way to get a collision is if someone else explicitly chooses "CoreWCF.Security.DataProtectionSecurityStateEncoder", which would have to be on purpose.

g7ed6e: > This won't work unless the service instance has been added to DI. The `Find<IServiceProvider>()` will throw null in that case.

Indeed. I'm wondering what is the best way to get a `IServiceProvider` here ? Should i try to pass it from call sites ? Despite a kind of ServiceLocator would i think do the job here I'm wondering if you would be open to it due to potential misuse that happen in the future.

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
g7ed6e: i reverted to ensure there is no breaking changes in public api

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
mconnew: We can add new api's when increasing the minor version so I think adding a constructor overload which takes a `IDataProtectionProvider` is probably the cleanest way to do this. Then in the code you can call CreateProtector(purpose). When we go to 2.0, we can remove the other constructors. I'm then not sure what to do about the other constructors though. I can think of two options.   

1. Mark them as obsolete and throw an exception saying to use the new constructor passing an IDataProtectionProvider
2. On non-windows, throw PlatformNotSupportedException (which is what would happen if you tried to use the class on Linux today), but on Windows, if the existing constructor is used, still use ProtectedData.

The problem with option 1 is we intend to maintain backwards compatibility for compilation and if someone is referencing the older constructor they will get a compiler warning due to the obsolete attribute, and if they have warn as error turned on, it's breaking that promise. I doubt anyone today is creating an instance, but it is still technically going against the policy. I want to avoid making a precedent so early on. The problem with option 2 is that we keep a dependency on the ProtectedData package for a theoretical scenario which is almost certain to not ever be an issue.  

Having typed all this out, my current preference is a modified version of 1, but without the obsolete attribute. It maintains compilation compatibility, I'm 99.9999999% certain it won't break anyone, and if it does, we can go option 2 after to fix them. I'm not wanting to be a dictator here, so feel free to express your thoughts.

Issue #486, opened 2021-10-29
ImpersonationOption.Required does not work in wsHttpBinding


vladascibulskis:
I'm trying to set   ` [CoreWCF.OperationBehavior(Impersonation = CoreWCF.ImpersonationOption.Required)]` attribute on one of the WCF service method.
I'm using wsHttpBinding set up as follows:
```
            var binding = new WSHttpBinding();
            binding.Security.Mode = SecurityMode.Transport;
            binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.Windows;
```

Getting following error:

> 10:46:21 ERR] First chance exception occured
> System.InvalidOperationException: SecurityContextProperty is missing from the request Message, this may indicate security is configured incorrectly.
>    at CoreWCF.Dispatcher.SecurityImpersonationBehavior.RunImpersonated2[T](MessageRpc rpc, ServiceSecurityContext securityContext, Boolean isSecurityContextImpersonationOn, Func`1 func) in D:\Odesk\Eddie\Radian_TFS\CoreWCF\src\CoreWCF.Primitives\src\CoreWCF\Dispatcher\SecurityImpersonationBehavior.cs:line 226
> 

I might be doing something wrong, but client works with other service created on .Net Framework, but does not work with CoreWCF (.Net Core 5).

Attached is a patch with unit test that reproduces same issue.

[0001-486-Unit-test.zip](https://github.com/CoreWCF/CoreWCF/files/7441929/0001-486-Unit-test.zip)

mconnew:
Notes for when this is addressed. The fix is to retrieve the WindowsIdentity from ASP.NET Core and create a ServiceSecurityContext to add to the incoming message which holds the WindowsIdentity. So we need code in the HTTP middleware which does this when security mode is transport and the credential type is Windows.

jonlouie:
@birojnayak I'm interested in helping with this

birojnayak:
@jonlouie assigning to you

birojnayak:
@mconnew can you assign.. I can't assign to @jonlouie

mconnew:
@birojnayak, it looks like you assigning it to Jon was successful.

PR #647, opened 2022-05-03
Enable impersonation with WSHttpBinding

jonlouie:
### Description
Closes: #486

This PR adds the current WindowsIdentity to the incoming message when impersonation is required.

---

### Testing
* Add test to ensure that a request succeeds when the Operation Contract method requires impersonation, SecurityMode is Transport, and CredentialType is Windows

### Concerns/Follow-up questions:
* Is there a better place to modify the ServiceSecurityContext on the message?
* Should there be different or additional checks before adding the WindowsIdentity to the ServiceSecurityContext?


jonlouie:
Investigating test failures...

src/CoreWCF.Primitives/src/CoreWCF/Security/MessageSecurityOverHttp.cs
```diff
@@ -34,16 +34,6 @@ public MessageCredentialType ClientCredentialType
                     throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException(nameof(value)));
                 }
 
-                if(value is MessageCredentialType.Windows)
-                {
```
mconnew, line 38: You can't remove this check yet. The Windows auth implementation depends on reflecting over an internal platform abstraction layer implementation in .NET Core which works on Windows and Linux. That internal implementation does not exist on .NET Framework so we throw to prevent the reflection based code from being hit when running on .NET Framework.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityImpersonationBehavior.cs
```diff
@@ -329,7 +329,14 @@ private ServiceSecurityContext GetAndCacheSecurityContext(MessageRpc rpc)
                 SecurityMessageProperty securityContextProperty = rpc.Request.Properties.Security;
                 if (securityContextProperty == null)
                 {
-                    securityContext = null; // SecurityContext.Anonymous
+                    if (rpc.Operation.Impersonation == ImpersonationOption.Required)
+                    {
```
mconnew, line 333: This code was correct. If you don't have a SecurityMessageProperty, then that means it wasn't populated when the client authenticated. What this code now does is assume the identity that the service was launched with and impersonate that user. This means if a method is attributed to require impersonation and the user didn't authenticate (could have multiple service endpoints with one having no authentication), they can now call the method, and worse, they will execute it with the credentials that the service is running as. If the Security property is null, it means it wasn't populated earlier by the authentication code, and that needs to be fixed. In this case, that means the ServicemodelHttpMiddleware.

jonlouie: Thanks, Matt. That makes sense- I will take a look ServiceModelHttpMiddleware

mconnew:
There are a couple of pieces which are missing in CoreWCF. The first is making sure that the client has authenticated, and if they haven't, to send the client a challenge. You do this using the extension class `AuthenticationHttpContextExtensions`. The first thing to do is call `AuthenticateAsync` passing the auth scheme that's configured on the binding. Check the AuthenticationResult to see if None is true, and if it is, call ChallengeAsync with the auth scheme and return (client authentication will come in a later request so we're done). If None is false, then get the ClaimsPrincipal from the result and use it to create the claims set. Get the Identity property from the ClaimsPrincipal and check if it's a WindowsIdentity. If it is, you can use that to create a WindowsClaimSet. You can get the ClaimsPrincipal from HttpContext.User but as we need to call AuthenticateAsync to make sure the client authenticated with the correct authentication scheme anyway, we might as well get it from AuthenitcationResult.

mconnew:
I forgot to mention, on Linux you will get a GenericIdentity. Find the code where this is handled for TransportWithMessageCredentials and you'll see how this is handled slightly differently. I can't remember the detail off the top of my head. Look for existing code which instantiates WindowsClaimSet and you should find it easily.

mconnew (changes requested):
Implementation isn't correct.

src/CoreWCF.Http/src/CoreWCF/Security/SecurityUtils.cs
```diff
@@ -16,5 +17,16 @@ public static void ValidateAnonymityConstraint(WindowsIdentity identity, bool al
                     new SecurityTokenValidationException(SR.AnonymousLogonsAreNotAllowed));
             }
         }
+
+        public static bool ShouldUseAuthentication(Binding binding)
+        {
+            if (binding is WSHttpBinding wsHttpBinding
+                && wsHttpBinding.Security.Mode != SecurityMode.None
```
mconnew, line 24: This isn't just needed for WSHttpBinding. Also you might have a CustomBinding which is equivalent. You need to call CreateBindingElements and on the resulting collection, use `Find<HttpTransportBindingElement>()` to get the transport binding element. Then you can check AuthenticationScheme to see if authentication is required.

src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
```diff
@@ -171,6 +172,10 @@ private RequestDelegate BuildBranch()
                         {
                             wcfApp.UseWebSockets(requestHandler.WebSocketOptions);
                         }
+                        if (SecurityUtils.ShouldUseAuthentication(dispatcher.Binding))
+                        {
```
mconnew, line 176: Which should go before the WebSocketOptions so that WebSockets is authenticated too.

mconnew: Take a look at RequestDelegateHandler.BuildHandler, it already extracts the HttpTransportBindingElement and stores the AuthenticationScheme in HttpTransportSettings (in the field _httpSettings) so I think ShouldUseAuthentication would be better to exist on RequestDelegateHandler as it already has all the information available.

src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
```diff
@@ -97,6 +98,21 @@ private WebSocketOptions CreateWebSocketOptions(HttpTransportBindingElement tbe)
 
         internal async Task HandleRequest(HttpContext context)
         {
+            if (SecurityUtils.ShouldUseAuthentication(_serviceDispatcher.Binding))
+            {
```
mconnew, line 102: As mentioned elsewhere, ShouldUseAuthentication should live in this class and be driven by _httpSettings.AuthenticationScheme.

src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
```diff
@@ -97,6 +98,21 @@ private WebSocketOptions CreateWebSocketOptions(HttpTransportBindingElement tbe)
...
+                AuthenticateResult authenticateResult = await context.AuthenticateAsync(scheme);
+                if (authenticateResult.None)
+                {
+                    await context.ChallengeAsync(scheme);
+                    return;
+                }
+                else
+                {
+                    context.Features.Set(authenticateResult);
+                }
```
mconnew, line 113: Features isn't the best place to stash this. Features is really about "do this to the request for me". You definitely shouldn't be putting a class in there that you don't own the type for. It's intended for middleware to put things in there for end user service implementations to use. There's an Items dictionary on HttpContext if you want to stash things. Use a CoreWCF specific name so there's no risk of name clashes. But I don't think you even need to stash this anyway. I'll comment more when I get to it.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -375,10 +382,34 @@ public override HttpOutput GetHttpOutput(Message message)
                 return HttpOutput.CreateHttpOutput(_aspNetContext, HttpTransportSettings, message, HttpMethod);
             }
 
-            protected override SecurityMessageProperty OnProcessAuthentication()
+            protected override async Task<SecurityMessageProperty> OnProcessAuthenticationAsync()
             {
-                // TODO: Wire up authentication for ASP.Net Core
-                //return Listener.ProcessAuthentication(listenerContext);
+                if (HttpTransportSettings.IsAuthenticationRequired)
+                {
```
mconnew, line 388: This property can be used in place of ShouldUseAuthentication elsewhere

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -375,10 +382,34 @@ public override HttpOutput GetHttpOutput(Message message)
...
+                    AuthenticateResult authenticateResult = _aspNetContext.Features.Get<AuthenticateResult>();
+                    if (authenticateResult == null)
+                    {
+                        // No authentication occurred
+                        return null;
+                    }
+
+                    if (authenticateResult.None)
+                    {
```
mconnew, line 397: This code is unreachable as RequestDelegateHandler.HandleRequest checks this and calls ChallengeAsync if it's true.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -375,10 +382,34 @@ public override HttpOutput GetHttpOutput(Message message)
...
+                        throw new Exception("The request requires authentication, but the user was not authenticated.");
+                    }
+                    else
+                    {
+                        ServiceSecurityContext securityContext = await CreateSecurityContextAsync(authenticateResult);
+                        SecurityMessageProperty securityMessageProperty = new SecurityMessageProperty
```
mconnew, line 403: You don't need to stash the AuthenticationResult as CreateSecurityContext only accesses the Principal property. You can get this from HttpContext.User

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -375,10 +382,34 @@ public override HttpOutput GetHttpOutput(Message message)
...
+                        {
+                            ServiceSecurityContext = securityContext
+                        };
+                        return securityMessageProperty;
+                    }
+                }
+
+                // Authentication is required but we were unable to authenticate.
+                // Should we throw instead?
```
mconnew: You only get here if HttpTransportSettings.IsAuthenticationRequired returns false, so it makes sense to not have a SecurityMessageProperty.

src/CoreWCF.Http/src/CoreWCF/Channels/AspNetCoreReplyChannel.cs
```diff
@@ -120,6 +120,12 @@ internal async Task HandleRequestCore(HttpContext context)
             }
 
             var requestContext = HttpRequestContext.CreateContext(_httpSettings, context);
+            bool isAuthenticated = await requestContext.ProcessAuthenticationAsync();
+            if (!isAuthenticated)
+            {
+                // throw exception?
+            }
```
mconnew, line 127: I'm not seeing a way for ProcessAuthentication to return false. It either returns true or throws an exception. This is because ValidateAuthentication always returns OK. I need to spend a bit more time looking at the original WCF code and work out what the equivalent conditions are, but I think there's enough feedback for you to be getting on with for now.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace CoreWCF.IdentityModel.Claims
 {
-    internal class WindowsClaimSet : ClaimSet, IIdentityInfo, IDisposable
+    public class WindowsClaimSet : ClaimSet, IIdentityInfo, IDisposable
     {
```
mconnew, line 15: We need to be thoughtful about how we make this class public. There's a lot of overloads of the constructor and we might not want to expose them all. I don't see a reason to make this public right now. New api surface also triggers bumping the minor version up to 1.1 and I want to avoid that for now if possible.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/UnconditionalPolicy.cs
```diff
@@ -15,7 +15,7 @@ internal interface IIdentityInfo
         IIdentity Identity { get; }
     }
 
-    internal class UnconditionalPolicy : IAuthorizationPolicy, IDisposable
+    public class UnconditionalPolicy : IAuthorizationPolicy, IDisposable
     {
```
mconnew, line 19: Why are you making this public? It's internal on .NET Framework.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Claims/WindowsClaimSet.cs
jonlouie: Thanks for catching this. This change is an artifact from a previous implementation which I eventually scrapped. Will revert.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Policy/UnconditionalPolicy.cs
jonlouie: Same as above, the accessor changes to this file are from a previous implementation which I eventually scrapped. Will revert.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/GenericSecurityToken.cs
```diff
@@ -8,14 +8,14 @@
 
 namespace CoreWCF.IdentityModel.Tokens
 {
-    internal class GenericSecurityToken : SecurityToken
+    public class GenericSecurityToken : SecurityToken
     {
```
mconnew, line 12: I understand why you need to make this public, you need to create an instance of this in the Http project. As this is becoming public, it needs a little bit of design work. Can you rename it to GenericIdentitySecurityToken. Can you add a constructor which only takes a name and it auto generates the Id. Have a look at WindowsSecurityToken for an example. I'm also not sure how I feel about the expiration time being 10 hours out. Other token types which don't have an expiration time with them use SecurityUtils.MaxUtcDateTime. I think that's probably the most common sense option to use. Make it non-expiring.  
Also can you make the constructor take a GenericIdentity instead of the name? Constructing a fresh GenericIdentity instance from just the name loses a lot of data. It contains things like IsAuthenticated and AuthenticationType which we lose. There is a Clone() method on GenericIdentity so you can make a copy.

src/CoreWCF.Http/src/CoreWCF/Security/SecurityUtils.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Net;
 using System.Security.Principal;
```
mconnew, line 5: You left behind a coupe of new usings when you backed out the code.

jonlouie: Good catch, will search for any others

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/GenericIdentitySecurityToken.cs
```diff
@@ -8,21 +8,38 @@
...
+            : this(name, SecurityUniqueId.Create().Value)
+        {
+        }
+        public GenericIdentitySecurityToken(GenericIdentity genericIdentity)
+            : this(genericIdentity, SecurityUniqueId.Create().Value)
+        {
+        }
+
+        public GenericIdentitySecurityToken(string name, string id)
         {
```
mconnew, line 28: I don't think we should have constructors which take a name. If someone only has a name, they can construct a GenericIdentity themselves with that name and pass that.

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);</TargetFrameworks>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
```
mconnew, line 5: Why did you remove net472 from the target frameworks? This is what's making the build fail.

jonlouie: This was a quick and dirty fix to get auth from aspnetcore working/building. Will push a commit that fixes this properly.

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -55,4 +56,10 @@
     <ProjectReference Include="..\..\CoreWCF.Primitives\src\CoreWCF.Primitives.csproj" />
     <ProjectReference Include="..\src\CoreWCF.Http.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Negotiate">
```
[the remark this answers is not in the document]
jonlouie: Will clean this up

jonlouie:
The auth changes in this PR rely on aspnetcore, so any test cases requiring authentication will fail when the project is targeting .NET Framework. Preprocessor directives were added to prevent execution of these tests from a net472 context; is there a better way to handle these cases? Or would it be better to hold off on these changes until CoreWCF targets .NET 6?

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -1,7 +1,7 @@
-<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0;net472</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);</TargetFrameworks>
```
mconnew, line 4: This is going to break on linux. The original xml had it conditional on the OS being Windows_NT so it wouldn't add net472 when compiling and running on linux.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
```diff
@@ -163,6 +167,57 @@ internal void WSHttpRequestReplyWithTransportMessageCertificateEchoString()
             }
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        [Fact, Description("transport-security-with-windows-authentication")]
+        [Trait("Category", "NetCoreOnly")] // Windows auth not supported on NetFx
+        internal void WSHttpRequestReplyWithTransportMessageImpersonateEchoString()
```
mconnew, line 173: I think you can get away with not conditionally compiling this test as the conditional piece is dealt with in the Startup class. The Trait on the test will prevent it running on NetFx.

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
jonlouie: Will fix

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
jonlouie: Will fix

jonlouie: One thing to note, this test still fails locally for `net472`. The test request requires authentication since the SecurityMode is Transport, but the auth service is only added for netcore runtimes since we're using aspnetcore.

For now, I have removed the preprocessor condition from this test. Let me know if I should re-add it based on the above information.

jonlouie:
Tests now passing on Windows platforms but failing on Linux:
> System.ComponentModel.Win32Exception : GSSAPI operation failed with error - Unspecified GSS failure.  Minor code may provide more information (SPNEGO cannot find mechanisms to negotiate).

Is additional handling required to impersonate from Linux? Or does the Linux test env need to be domain-joined to Active Directory for this to pass?

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
mconnew: I spoke with the asp.net core team and this is expected. You could switch to using the HttpSys host instead of Kestrel and configuring it to use Windows auth and apparently that should work find in net472. HttpSys won't work on Linux, although you can reference it from Linux. It will just throw at runtime. Maybe you could make the code have different behavior based on what runtime framework it's running on and either use Kestrel with the existing code for .NET Core/.NET and use HttpSys configured for Windows auth on net472.

mconnew:
I don't believe you can actually impersonate on Linux. You need a WindowsIdentity object to do impersonation, and on Linux you get a GenericIdentity. So for Linux the change is limited to populating the ServiceSecurityContext for identification purposes.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
jonlouie: If my understanding is correct, this test uses SecurityMode.Transport which requires an `https` endpoint. It seems like getting HttpSys to work with https [requires SSL cert registration](https://github.com/dotnet/aspnetcore/issues/4965#issuecomment-447352643).

Is there an effective way to handle this with unit tests? Or would it make more sense to throw a PlatformNotSupported exception if we're trying to use authentication with net472? Or let the test fail?

mconnew: If the machine has VS installed, it will generally have IIS express also installed, which sets up an ssl cert registration on some default ports. The IIS express test certificate is configured on ports 44300-44399.  So pick a port in that range somewhere and configure HttpSys to use that port for HTTPS.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
```diff
@@ -163,6 +169,105 @@ internal void WSHttpRequestReplyWithTransportMessageCertificateEchoString()
...
+                string result = channel.EchoForImpersonation(testString);
+                Assert.Equal(testString, result);
+                ((IChannel)channel).Close();
+            }
+        }
+#endif
+
+        [Fact, Description("transport-security-with-windows-authentication-httpsys")]
+        internal void WSHttpRequestImpersonateWithHttpSys()
+        {
```
mconnew, line 201: This test is failing on Linux due to Linux not supporting HttpSys. You need to add a windows only trait. You do that by adding the following attribute:
```c#
[Trait("Category", "WindowsOnly")]
```

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
```diff
@@ -163,6 +169,105 @@ internal void WSHttpRequestReplyWithTransportMessageCertificateEchoString()
...
+                Assert.Throws<System.ServiceModel.FaultException<System.ServiceModel.ExceptionDetail>>(() =>
+                {
+                    channel.EchoForImpersonation(testString);
+                });
+            }
+        }
+
+        [Fact, Description("no-security-with-an-anonymous-client-using-impersonation-httpsys")]
+        public void WSHttpRequestImpersonateWithHttpSysFailsWithoutAuthentication()
+        {
```
mconnew, line 249: This test also needs the WindowsOnly trait added.

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
```diff
@@ -198,6 +198,53 @@ public static Binding GetBufferedModHttp1Binding()
...
+                    logging.AddProvider(new XunitLoggerProvider(outputHelper));
+                logging.AddFilter("Default", LogLevel.Debug);
+                logging.AddFilter("Microsoft", LogLevel.Debug);
+                logging.SetMinimumLevel(LogLevel.Debug);
+            })
+#endif // DEBUG
+            .UseHttpSys(options =>
+            {
+                options.AllowSynchronousIO = true;
+                options.Authentication.AllowAnonymous = false;
```
mconnew, line 216: I don't believe you need this any more.

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
```diff
@@ -198,6 +198,53 @@ public static Binding GetBufferedModHttp1Binding()
...
+                options.Authentication.Schemes = Microsoft.AspNetCore.Server.HttpSys.AuthenticationSchemes.Negotiate
+                                                 | Microsoft.AspNetCore.Server.HttpSys.AuthenticationSchemes.NTLM;
+                options.MaxConnections = null;
+                options.MaxRequestBodySize = 30000000;
+                options.UrlPrefixes.Add("https://localhost:5001");
+            })
```
mconnew, line 222: You need to use one of the IIS Express preconfigured ports. This is failing in the PR test run because the certificate isn't configured on port 5001. Use a port number between 44300 and 44399 and the SSL cert will be configured correctly.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
jonlouie: Added the WindowsOnly attribute

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
jonlouie: Added the WindowsOnly attribute

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
jonlouie: Removed unnecessary config

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
jonlouie: Fixed

jonlouie: Looks like SSL is failing after changing to port 44300: https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=72369&view=logs&j=61b3ff45-772e-578d-ad5f-a6aa4ee834dd&t=73006cae-d2fd-57c9-bf9e-49e14522176c&l=310

jonlouie:
@mconnew Updated the HttpSys tests to be skipped. Looks like the impersonation tests on Linux are failing with the following error: `System.ComponentModel.Win32Exception : GSSAPI operation failed with error - Unspecified GSS failure.  Minor code may provide more information (SPNEGO cannot find mechanisms to negotiate).`

Should these tests be skipped as well?

mconnew:
Did you run the tests manually on Linux? That error basically means the necessary libraries haven't been installed and/or configured on the host. If the tests pass on Windows, add the trait which makes them Windows only.

mconnew (approved):

PR #656, opened 2022-05-07
test: skip failing .NET framework WSHttpBinding test

### LINKED ISSUE #518: WSHttpBinding_WithDefaultSetting() test failing on .NET Framework
The `WSHttpBinding_WithDefaultSetting()` test is failing with a `PlatformNotSupportedException` exception when executing the test on Windows. I'm running on Windows 10 version 21H1, 64 bit.

Debugging the test narrows me down to the `MessageSecurityOverHttp.ClientCredentialType` setter, where an exception is being trown if the `MessageCredentialType` is `Windows`, and the `RuntimeInformation.FrameworkDescription` contains `.NET Framework`.

There is a comment in the code to remove it after .net 5, and if I skip this platform test, then the test succeeds.

It is the only test failing in the entire solution on my computer, and I would expect to see all tests passing.

randomnoise:
`WSHttpBinding_WithDefaultSetting()` test is not flaky and it fails locally with every test run. This should fix that issue.

Closes #518

### DIFF
--- a/src/CoreWCF.ConfigurationManager/tests/WSHttpBindingTests.cs
+++ b/src/CoreWCF.ConfigurationManager/tests/WSHttpBindingTests.cs
@@ -50,7 +50,7 @@ namespace CoreWCF.ConfigurationManager.Tests
 
                     var actualBinding = settingHolder.ResolveBinding(nameof(WSHttpBinding), expectedName) as WSHttpBinding;
                     Assert.Equal(expectedName, actualBinding.Name);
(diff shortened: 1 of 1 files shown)


mconnew:
This isn't needed. There is a trait on the test which says it's NetCoreOnly. The automatic PR test run passes a filter when running tests on .NET Framework to say exclude tests with this trait. We have a similar thing for Windows only tests when running on Linux. We'll soon be switching to developing for CoreWCF 2.0 which drops .NET Framework support so it will be a non issue. If you are running tests in VS, you can apply the trait filter in the test explorer UI. You can also pass a filter to the dotnet test command if using the CLI.

PR #664, opened 2022-05-11
Response honors the request Connection header values

### LINKED ISSUE #369: CoreWCF does not honor the 'Connection: close' header if set in the request
# Problem
CoreWCF appears to always send `Connection: keep-alive` in the response HTTP headers, even if the request headers specified `Connection: close`. This is a regression from how WCF worked on .NET Framework.

# Steps to repro:
1. Run the sample server application that comes with the CoreWCF code (src\Samples\NetCoreServer\NetCoreServer.csproj).
2. Make the following web request:

```http
POST http://localhost:802/basichttp HTTP/1.1
Host:localhost:802
SOAPAction:http://tempuri.org/IEchoService/Echo
Connection:close
Content-Type:text/xml
Content-Length:219

<soapenv:Envelope xmlns:soapenv='http://schemas.xmlsoap.org/soap/envelope/'>
    <soapenv:Body>
    <Echo xmlns='http://tempuri.org/'>
        <text>Hello</text>
    </Echo>
    </soapenv:Body>
</soapenv:Envelope>
```

# Expected result:
The server should return a response with `Connection: close` set in the HTTP headers.

# Actual result:
The server sends the following response. Note that `Connection: keep-alive` is returned.

```html
HTTP/1.1 200 OK
Connection: keep-alive
Date: Fri, 04 Jun 2021 19:12:43 GMT
Content-Type: text/xml; charset=utf-8
Server: Kestrel
Content-Length: 181

<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"><s:Body><EchoResponse xmlns="http://tempuri.org/"><EchoResult>Hello</EchoResult></EchoResponse></s:Body></s:Envelope>
```

# Notes
I used the 'DesktopServer' sample to verify that the old behavior of .NET Framework WCF does indeed return the `Connection: close` header, so this definitely seems to be a regression.

randomnoise:
Connection response header is set according to the connection request header.

Also, request header can be case-insensitive (reference: https://www.rfc-editor.org/rfc/rfc7230#section-3.2)

Closes #369

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
@@ -358,11 +358,11 @@ namespace CoreWCF.Channels
                     {
                         return base.KeepAliveEnabled;
                     }
-                    else if (_aspNetContext.Request.Headers[connectionHeader] == "keep-alive")
+                    else if (StringComparer.OrdinalIgnoreCase.Equals("keep-alive", _aspNetContext.Request.Headers[connectionHeader]))
                     {
                         return true;
                     }
-                    else if (_aspNetContext.Request.Headers[connectionHeader] == "close")
+                    else if (StringComparer.OrdinalIgnoreCase.Equals("close", _aspNetContext.Request.Headers[connectionHeader]))
                     {
                         return false;
                     }
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
@@ -31,7 +31,7 @@ namespace CoreWCF.Channels
             _replySentTcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
         }
 
-        public bool KeepAliveEnabled
+        public virtual bool KeepAliveEnabled
         {
             get
             {
@@ -348,6 +348,31 @@ namespace CoreWCF.Channels
 
             public override string HttpMethod => _aspNetContext.Request.Method;
 
+            public override bool KeepAliveEnabled
+            {
+                get
+                {
+                    string connectionHeader = HttpRequestHeader.Connection.ToString();
+
+                    if (!_aspNetContext.Request.Headers.ContainsKey(connectionHeader))
+                    {
+                        return base.KeepAliveEnabled;
+                    }
+                    else if (_aspNetContext.Request.Headers[connectionHeader] == "keep-alive")
+                    {
+                        return true;
+                    }
+                    else if (_aspNetContext.Request.Headers[connectionHeader] == "close")
+                    {
+                        return false;
+                    }
+                    else
+                    {
+                        return base.KeepAliveEnabled;
+                    }
+                }
+            }
+
             protected override HttpInput GetHttpInput()
             {
                 return new AspNetCoreHttpInput(this);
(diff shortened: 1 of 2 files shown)


src/CoreWCF.Http/tests/HttpResponseMessagePropertyTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                        httpBinding,
+                        new System.ServiceModel.EndpointAddress(
+                            new Uri("http://localhost:8080/BasicWcfService/basichttp.svc")));
+
+                    channel = factory.CreateChannel();
+                    using (new System.ServiceModel.OperationContextScope(channel as System.ServiceModel.IContextChannel))
+                    {
+                        var httpRequestMessageProperty = new System.ServiceModel.Channels.HttpRequestMessageProperty();
+                        httpRequestMessageProperty.Headers[HttpRequestHeader.Connection] = connectionRequestHeader;
+                        System.ServiceModel.OperationContext.Current.OutgoingMessageProperties.Add(
```
mconnew, line 54: Do you know what this does on the wire when setting it to null?

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -348,6 +348,31 @@ public AspNetCoreHttpContext(IHttpTransportFactorySettings settings, HttpContext
 
             public override string HttpMethod => _aspNetContext.Request.Method;
 
+            public override bool KeepAliveEnabled
+            {
+                get
+                {
+                    string connectionHeader = HttpRequestHeader.Connection.ToString();
+
```
mconnew, line 356: I think using nameof(HttpRequestHeader.Connection) would achieve the same thing but no incur the cost of calling Enum.ToString() which historically has been really expensive although I think they did improve it. If you use nameof, I think you can make this into a private const string. You're calculating the string value on every request. If you use nameof though, it will just be turned into a const string by the compiler anyway so you could keep it inline.

mconnew: Can you add a TODO comment in here as .NET 6 expanded IHeaderDictionary to have properties for each of the common headers. Once we drop support for earlier versions, we can do _aspNetContext.Request.Headers.Connection instead.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -348,6 +348,31 @@ public AspNetCoreHttpContext(IHttpTransportFactorySettings settings, HttpContext
...
+                    if (!_aspNetContext.Request.Headers.ContainsKey(connectionHeader))
+                    {
+                        return base.KeepAliveEnabled;
+                    }
+                    else if (StringComparer.OrdinalIgnoreCase.Equals("keep-alive", _aspNetContext.Request.Headers[connectionHeader]))
+                    {
```
mconnew, line 362: Is the binding has been configured to not use keep-alive, we should always return false otherwise the client can control resource usage on the server by creating a large number of persistent connections. I think starting with
```c#
if (!base.KeepAliveEnabled) return false;
```
That way if the binding has it disabled, we always return "close".

mconnew: I think this whole method can be simplified.
```c#
if (!base.KeepAliveEnabled) return false;
// We know base.KeepAliveEnabled is true
if (_aspNetContext.Request.Headers.ContainsKey(connectionHeader))
{
    return !StringComparer.OrdinalIgnoreCase.Equals("close", _aspNetContext.Request.Headers[connectionHeader])
}

return true;
```
I think that's sufficient. If base.KeepAliveEnabled is false, we always return false. Otherwise it's enabled so if the client provided a value, use that. If the client didn't supply a value, return true (as we know it's enabled).

PR #665, opened 2022-05-11
OpenAPI documentation generation improvements:

JonathanHopeDMRC:
- Adding tags to OpenAPI document with sorting callback.
 - Added min length, max length, and format for properties.
 - Fix for request body description not showing up.
 - Fix for bug with references.

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/Attributes/OpenApiPropertyAttribute.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/Attributes/OpenApiPropertyAttribute.cs
@@ -20,5 +20,20 @@ namespace CoreWCF.OpenApi.Attributes
         /// Whether the property is required or not.
         /// </summary>
         public bool IsRequired { get; set; }
+
+        /// <summary>
+        /// Minimum length of the property.
+        /// </summary>
+        public int MinLength { get; set; }
+
+        /// <summary>
+        /// Maximum length of the property.
+        /// </summary>
+        public int MaxLength { get; set; }
+
+        /// <summary>
+        /// How the property should be formatted.
+        /// </summary>
+        public string Format { get; set; }
     }
 }
--- a/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiOptions.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiOptions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.OpenApi.Models;
 
 namespace CoreWCF.OpenApi
 {
@@ -70,5 +71,10 @@ namespace CoreWCF.OpenApi
         /// Any tags to hide.
         /// </summary>
         public IEnumerable<string> TagsToHide { get; set; } = new List<string>();
+
+        /// <summary>
+        /// Optional callback to sort tags.
+        /// </summary>
+        public IComparer<OpenApiTag> TagsSorter { get; set; }
     }
 }
--- a/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
@@ -58,6 +58,13 @@ namespace CoreWCF.OpenApi
             PopulateOpenApiInfo(document, info);
             PopulateOpenApiPathsOperations(document, contracts, info.TagsToHide);
 
+            if (info.TagsSorter != null)
+            {
+                var tags = document.Tags as List<OpenApiTag> ?? document.Tags.ToList();
+                tags.Sort(info.TagsSorter);
+                document.Tags = tags;
+            }
+
             return document;
         }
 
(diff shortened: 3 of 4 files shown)


src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/Attributes/OpenApiPropertyAttribute.cs
```diff
@@ -20,5 +20,55 @@ public sealed class OpenApiPropertyAttribute : Attribute
...
+            {
+                _minLength = value;
+                MinLengthSet = true;
+            }
+        }
+
+        /// <summary>
+        /// Maximum length of the property.
+        /// </summary>
+        public int MaxLength
```
mconnew, line 43: What do you think about defining MaxLength and MinLength as nullables to know if a value was set? So define them like this:
```c#
public int? MaxLength { get; set; }
```

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiOptions.cs
```diff
@@ -70,5 +71,10 @@ public sealed class OpenApiOptions
         /// Any tags to hide.
         /// </summary>
         public IEnumerable<string> TagsToHide { get; set; } = new List<string>();
+
+        /// <summary>
+        /// Optional callback to sort tags.
+        /// </summary>
+        public Func<IEnumerable<OpenApiTag>, IEnumerable<OpenApiTag>> SortTags { get; set; }
```
mconnew, line 78: Did you consider having a property of type `IComparable<OpenApiTag>` so that a comparer can be provided? That way you can instantiate a `List<OpenApiTag>` and once they've all been populated, you can call `List<OpenApi>.Sort(IComparer<OpenApi> comparer)` if a comparer has been provided.  
I can think of a couple of reasons you might not want to do that.  If the required order might be non-trivial to represent in an IComparer (although I can't think of a scenario where I couldn't write it without a small amount of code). You might want to have this be a back door mechanism to filter the OpenApiTag instances or to enable extra ones being added. But that seems like a bad idea to me as you are encouraging breaking someone who depends on that if the implementation changes.

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/Attributes/OpenApiPropertyAttribute.cs
JonathanHopeDMRC: That was the initial route I went down, but I got a "'MaxLength' is not a valid named attribute argument because it is not a valid attribute parameter type".

mconnew: I just thought of another way to do this. Don't use GetCustomAttributes, use GetCustomAttributesData instead. That doesn't give you an instance of the attribute, it gives you an instance of CustomAttributeData. This is a representation of how the attribute was used in the original code. If has a NamedArguments list which only contains the arguments specified on the attribute. If they didn't specify MaxLength, there won't be an entry in NamedArguments for MaxLength. It's a lot cleaner way of handling this.

JonathanHopeDMRC: Oooh that is clever. Changes have been pushed.

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
```diff
@@ -58,6 +58,13 @@ public static OpenApiDocument BuildOpenApiSpecificationDocument(OpenApiOptions i
             PopulateOpenApiInfo(document, info);
             PopulateOpenApiPathsOperations(document, contracts, info.TagsToHide);
 
+            if (info.TagsSorter != null)
+            {
+                var tags = document.Tags.ToList();
+                tags.Sort(info.TagsSorter);
```
mconnew, line 64: You can avoid an allocation here as Tags defaults to a `List<OpenApiTag>`. In case someone else replaces it or they change the implementation, you need the ToList as a backup, but the following code should avoid an allocation.
```c#
var tags = document.Tags as List<OpenApiTag> ?? document.Tags.ToList();
```

src/CoreWCF.WebHttp/tests/OpenApiWcfTests.cs
```diff
@@ -1089,7 +1090,7 @@ public void PropertyCanBeHiddenWithTag()
         private class SimplePropertyClass
         {
             [DataMember(Name = "Property")]
-            [OpenApiProperty(Description = "description", IsRequired = true)]
+            [OpenApiProperty(Description = "description", IsRequired = true, MinLength = 1, MaxLength = 2, Format = "format")]
             public string Property { get; set; }
```
mconnew, line 1094: Is there a negative test to make sure that when MinLength and MaxLength are not set, that they aren't in the generated json file? I couldn't find one, but it's possible it's implicitly tested somewhere.

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
JonathanHopeDMRC: I pushed up the changes for this.

src/CoreWCF.WebHttp/tests/OpenApiWcfTests.cs
JonathanHopeDMRC: I added some additional asserts for the negative case.

PR #682, opened 2022-05-23
Add sensible names to test runs to improve readability in DevOps

mconnew:
Currently when there's a test failure, you can't tell which platform it failed on. This should add nice names in the UI so you can tell which tests ran on which platform.


birojnayak (approved):
looks ok... it's the same test failing ?

g7ed6e (approved):

Issue #676, opened 2022-05-19
FaultExceptions thrown from UserNamePasswordValidators do not get serialised correctly


marlon-tucker:
Using CoreWCF version `1.0.2` 

We have a couple of custom classes that inherit from the `UserNamePasswordValidator` base class. Inside the `ValidateAsync` method we throw `FaultExceptions` to signify various authentication failures. CoreWCF does not serialise these back into a WCF response, which in our case should be a SOAP response. 

The response from AspNetCore indicates the exception is not being handled by CoreWCF so it's falling through the middleware pipeline eventually hitting the developer exception handling page, which generates a `text/plain` response.

I have double checked our NET Framework code and I have confirmed FaultExceptions thrown from `UserNamePasswordValidator` are correctly serialised.

birojnayak:
assigning to @jonlouie  we do have test https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs#L110 and we can throw exception https://github.com/CoreWCF/CoreWCF/blob/a6531a5d237e79c645fa7670b70763438139bafb/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs#L166 and see  why not getting fault exception..

jonlouie:
@birojnayak Looks like we have a negative test case where auth is set to fail: https://github.com/CoreWCF/CoreWCF/blob/a6531a5d237e79c645fa7670b70763438139bafb/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs#L77-L107

I stepped through the code and we do attempt to [return a fault SOAP message](https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/src/CoreWCF/Channels/AspNetCoreReplyChannel.cs#L139-L140), but I'll need to reproduce the issue to get a better understanding on where the issue is.

marlon-tucker:
That test looks like it is specifically testing the behaviour of the time out option of the binding, rather than testing exceptions thrown by the CustomUsernameValidator are getting serialised correctly.

jonlouie:
@marlon-tucker Thanks for the clarification. I wrote a new test to throw within the `UserNamePasswordValidator` class and found that the security exception was not being set, preventing a fault message reply. PR #684 addresses this, currently pending review.

PR #684, opened 2022-05-24
Send fault message on security exception

jonlouie:
### Description
Closes: #676

This PR sets the security exception if there is a failure during security validation. Previously, this exception was set to null and never updated, resulting in the fault message not being sent.

---

### Testing
* Add negative test to ensure that the correct fault exception is thrown when user name validation fails

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
@@ -398,7 +398,6 @@ namespace CoreWCF.Dispatcher
             {
                 return null;
             }
-            Exception securityException = null;
             TimeSpan timeout = ServiceDefaults.ReceiveTimeout;
             Message message = requestContext.RequestMessage;
             TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
@@ -418,14 +417,14 @@ namespace CoreWCF.Dispatcher
                 }
                 return new SecurityRequestContext(message, requestContext, SecurityProtocol, correlationState, ServiceDefaults.SendTimeout, ServiceDefaults.CloseTimeout);
             }
-            catch
+            catch (Exception securityException)
             {
-                SendFaultIfRequired(securityException, requestContext, timeoutHelper.RemainingTime());
+                await SendFaultIfRequiredAsync(securityException, requestContext, timeoutHelper.RemainingTime());
                 throw;
             }
         }
 
-        private void SendFaultIfRequired(Exception e, RequestContext innerContext, TimeSpan timeout)
+        private async Task SendFaultIfRequiredAsync(Exception e, RequestContext innerContext, TimeSpan timeout)
         {
             if (!_sendUnsecuredFaults)
             {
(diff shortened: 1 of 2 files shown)


birojnayak (approved):
looks ok

mconnew:
I think you just helped find a bug which has been causing a flakey test. There's another small change that's needed. `SendFaultIfRequired` needs to be made async and the calls to `innerContext.ReplyAsync(faultMessage);` and `innerContext.CloseAsync(timeoutHelper.GetCancellationToken());` need to be awaited. We've had a flakey test which sometimes works and sometimes doesn't in the failure code path, and that's because `CloseAsync` has been aborting the connection before the fault message had finsihed sending. There's a second similar case on line 642 (where it's called) and line 648 where there's a similarly named method which isn't awaiting on the SendAsync (and should also be calling CloseAsync after sending).

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
```diff
@@ -640,13 +639,13 @@ private async ValueTask<Message> ProcessInnerItemAsync(Message innerItem, TimeSp
...
             {
-                SendFaultIfRequired(securityException, unverifiedMessage, timeoutHelper.RemainingTime());
+                await SendFaultIfRequired(securityException, unverifiedMessage, timeoutHelper.RemainingTime());
                 return null;
             }
             return innerItem;
         }
 
-        private void SendFaultIfRequired(Exception e, Message unverifiedMessage, TimeSpan timeout)
+        private async Task SendFaultIfRequired(Exception e, Message unverifiedMessage, TimeSpan timeout)
```
mconnew, line 648: Can you change the method names that are now async to have the suffix Async?

jonlouie: Ah, good catch. WIll fix

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/SecurityServiceDispatcher.cs
```diff
@@ -663,7 +662,9 @@ private void SendFaultIfRequired(Exception e, Message unverifiedMessage, TimeSpa
                 {
                     if (unverifiedMessage.Headers.MessageId != null)
                         faultMessage.InitializeReply(unverifiedMessage);
-                    ((IDuplexChannel)InnerDuplexChannel).SendAsync(faultMessage);
+                    TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+                    await ((IDuplexChannel)InnerDuplexChannel).SendAsync(faultMessage);
+                    await ((IDuplexChannel)InnerDuplexChannel).CloseAsync(timeoutHelper.GetCancellationToken());
                 }
```
mconnew, line 668: There are some test failures with the error message "Expected dispatcher state to be Opened or Faulted, instead it's Closed", so I think I was mistaken in saying that CloseAsync needs to be added here. The difference between this method and the earlier one is this is for duplex channels so it's not appropriate to close as the channel itself didn't fault, it was just the message was bad.

jonlouie: That makes sense. Fixed and squashed the commits in the latest push

PR #695, opened 2022-06-01
Test pipeline to problem installed ssl certs for Http.SYS tests

mconnew:
This is not intended to be merged. I've added an extra pipeline step to the test stage to try and discover what SSL certs are bound. We're expecting as IIS Express is installed that the IIS Express test certificate is installed and bound to ports 44300 to 44399. This PR is intended to help diagnose PR failures in #647.


mconnew:
I've verified that the IIS certificate is bound to ports 44300-44399 on the devops windows build agents so that's not the problem.

PR #711, opened 2022-06-09
Removes an sync over async call to improve closing behavior of the ServiceChannel

### LINKED ISSUE #710: ThreadPool exhaustion with NetHttpBinding (WebSockets)
When using a `NetHttp(s)Binding` with web socket transfer and a lot of concurrent requests, the ThreadPool can get exhausted quickly. The dotnet-counters tool will show this, e.g.:

>    ThreadPool Queue Length                                      185
>    ThreadPool Thread Count                                       70

All thread pool worker threads are waiting at the same call site:

> System.Private.CoreLib.dll!System.Threading.ManualResetEventSlim.Wait
> System.Private.CoreLib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 570
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 2985
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.InternalWaitCore(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 2920
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task task) Line 141
> [Waiting on Async Operation, double-click or press enter to view Async Call Stacks]
> CoreWCF.Primitives.dll!CoreWCF.Channels.ServiceChannel.DecrementActivity() Line 719
> CoreWCF.Primitives.dll!CoreWCF.Channels.ServiceChannel.HandleReceiveComplete(CoreWCF.Channels.RequestContext context) Line 789
> CoreWCF.Primitives.dll!CoreWCF.Dispatcher.ChannelHandler.HandleReceiveCompleteAsync(CoreWCF.Channels.RequestContext request) Line 402
> CoreWCF.Primitives.dll!CoreWCF.Dispatcher.ChannelHandler.DispatchAsync(CoreWCF.Channels.RequestContext requestContext) Line 226
> CoreWCF.Primitives.dll!CoreWCF.Dispatcher.DuplexChannelBinder.DispatchAsync(CoreWCF.Channels.Message message) Line 402
> CoreWCF.Http.dll!CoreWCF.Channels.TransportDuplexSessionChannel.StartReceivingAsync() Line 80
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<StartReceivingAsync>d__37>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<StartReceivingAsync>d__37>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<StartReceivingAsync>d__37>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<(CoreWCF.Channels.Message, bool)>.TrySetResult((CoreWCF.Channels.Message, bool) result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<(CoreWCF.Channels.Message, bool)>.SetExistingTaskResult(System.Threading.Tasks.Task<(CoreWCF.Channels.Message, bool)> task, (CoreWCF.Channels.Message, bool) result) Line 446
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<(CoreWCF.Channels.Message, bool)>.SetResult((CoreWCF.Channels.Message, bool) result) Line 432
> [Completed] CoreWCF.Http.dll!CoreWCF.Channels.TransportDuplexSessionChannel.TryReceiveAsync(System.Threading.CancellationToken token) Line 137
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<(CoreWCF.Channels.Message, bool)>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<TryReceiveAsync>d__39>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<(CoreWCF.Channels.Message, bool)>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<TryReceiveAsync>d__39>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<(System.__Canon, bool)>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<TryReceiveAsync>d__39>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.__Canon>.TrySetResult(System.__Canon result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.SetExistingTaskResult(System.Threading.Tasks.Task<System.__Canon> task, System.__Canon result) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.SetResult(CoreWCF.Channels.Message result) Line 432
> [Completed] CoreWCF.Http.dll!CoreWCF.Channels.TransportDuplexSessionChannel.ReceiveAsync(System.Threading.CancellationToken token) Line 118
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<ReceiveAsync>d__38>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<ReceiveAsync>d__38>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.AsyncStateMachineBox<CoreWCF.Channels.TransportDuplexSessionChannel.<ReceiveAsync>d__38>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.__Canon>.TrySetResult(System.__Canon result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.SetExistingTaskResult(System.Threading.Tasks.Task<System.__Canon> task, System.__Canon result) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.SetResult(CoreWCF.Channels.Message result) Line 432
> [Completed] CoreWCF.Http.dll!CoreWCF.Channels.SynchronizedMessageSource.ReceiveAsync(System.Threading.CancellationToken token) Line 71
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.SynchronizedMessageSource.<ReceiveAsync>d__4>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.SynchronizedMessageSource.<ReceiveAsync>d__4>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.AsyncStateMachineBox<CoreWCF.Channels.SynchronizedMessageSource.<ReceiveAsync>d__4>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.__Canon>.TrySetResult(System.__Canon result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.SetExistingTaskResult(System.Threading.Tasks.Task<System.__Canon> task, System.__Canon result) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.SetResult(CoreWCF.Channels.Message result) Line 432
> [Completed] CoreWCF.Http.dll!CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.ReceiveAsync(System.Threading.CancellationToken token) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReceiveAsync>d__30>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<CoreWCF.Channels.Message>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReceiveAsync>d__30>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.__Canon>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReceiveAsync>d__30>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.__Canon>.TrySetResult(System.__Canon result) Line 401
> System.Private.CoreLib.dll!System.Threading.Tasks.TaskCompletionSource<object>.SetResult(object result) Line 198
> CoreWCF.Http.dll!CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.StartNextReceiveAsync() Line 689
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<StartNextReceiveAsync>d__36>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<StartNextReceiveAsync>d__36>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<StartNextReceiveAsync>d__36>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.Threading.Tasks.VoidTaskResult>.TrySetResult(System.Threading.Tasks.VoidTaskResult result) Line 401
> CoreWCF.Http.dll!CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.ReadBufferedMessageAsync() Line 545
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReadBufferedMessageAsync>d__32>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReadBufferedMessageAsync>d__32>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.AsyncStateMachineBox<CoreWCF.Channels.WebSocketTransportDuplexSessionChannel.WebSocketMessageSource.<ReadBufferedMessageAsync>d__32>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.__Canon>.TrySetResult(System.__Canon result) Line 401
> System.Private.CoreLib.dll!System.Threading.Tasks.ValueTask<System.Net.WebSockets.WebSocketReceiveResult>.ValueTaskSourceAsTask..cctor.AnonymousMethod__4_0(object state) Line 638
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<System.Net.WebSockets.WebSocketReceiveResult>.SignalCompletion() Line 233
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<System.Net.WebSockets.WebSocketReceiveResult>.SetResult(System.Net.WebSockets.WebSocketReceiveResult result) Line 63
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Net.WebSockets.WebSocketReceiveResult>.StateMachineBox.SetResult(System.Net.WebSockets.WebSocketReceiveResult result) Line 250
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Net.WebSockets.WebSocketReceiveResult>.SetResult(System.Net.WebSockets.WebSocketReceiveResult result) Line 62
> System.Net.WebSockets.dll!System.Net.WebSockets.ManagedWebSocket.ReceiveAsyncPrivate<System.Net.WebSockets.WebSocketReceiveResult>(System.Memory<byte> payloadBuffer, System.Threading.CancellationToken cancellationToken) Line 861
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Net.WebSockets.WebSocketReceiveResult>.StateMachineBox<System.Net.WebSockets.ManagedWebSocket.<ReceiveAsyncPrivate>d__63<System.Net.WebSockets.WebSocketReceiveResult>>.ExecutionContextCallback(object s) Line 394
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Net.WebSockets.WebSocketReceiveResult>.StateMachineBox<System.Net.WebSockets.ManagedWebSocket.<ReceiveAsyncPrivate>d__63<System.Net.WebSockets.WebSocketReceiveResult>>.MoveNext() Line 416
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<System.Threading.Tasks.VoidTaskResult>.SignalCompletion() Line 233
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder.SetResult() Line 51
> System.Net.WebSockets.dll!System.Net.WebSockets.ManagedWebSocket.EnsureBufferContainsAsync(int minimumRequiredBytes, System.Threading.CancellationToken cancellationToken, bool throwOnPrematureClosure) Line 1374
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.StateMachineBox<System.Net.WebSockets.ManagedWebSocket.<EnsureBufferContainsAsync>d__74>.ExecutionContextCallback(object s) Line 394
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.Threading.Tasks.VoidTaskResult>.StateMachineBox<System.Net.WebSockets.ManagedWebSocket.<EnsureBufferContainsAsync>d__74>.MoveNext() Line 416
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<int>.TrySetResult(int result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.SetExistingTaskResult(System.Threading.Tasks.Task<int> task, int result) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncValueTaskMethodBuilder<int>.SetResult(int result) Line 56
> [Completed] Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.ReadAsyncInternal(System.Memory<byte> destination, System.Threading.CancellationToken cancellationToken) Line 153
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.<ReadAsyncInternal>d__29>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.<ReadAsyncInternal>d__29>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.<ReadAsyncInternal>d__29>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<System.IO.Pipelines.ReadResult>.SignalCompletion() Line 233
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<System.IO.Pipelines.ReadResult>.SetResult(System.IO.Pipelines.ReadResult result) Line 63
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.IO.Pipelines.ReadResult>.StateMachineBox.SetResult(System.IO.Pipelines.ReadResult result) Line 250
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.IO.Pipelines.ReadResult>.SetResult(System.IO.Pipelines.ReadResult result) Line 62
> Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1UpgradeMessageBody.ReadAsyncInternalAwaited(System.Threading.Tasks.ValueTask<System.IO.Pipelines.ReadResult> readTask, System.Threading.CancellationToken cancellationToken) Line 108
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.IO.Pipelines.ReadResult>.StateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1UpgradeMessageBody.<ReadAsyncInternalAwaited>d__12>.ExecutionContextCallback(object s) Line 394
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<System.IO.Pipelines.ReadResult>.StateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1UpgradeMessageBody.<ReadAsyncInternalAwaited>d__12>.MoveNext() Line 416
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<System.IO.Pipelines.ReadResult>.TrySetResult(System.IO.Pipelines.ReadResult result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.IO.Pipelines.ReadResult>.SetExistingTaskResult(System.Threading.Tasks.Task<System.IO.Pipelines.ReadResult> task, System.IO.Pipelines.ReadResult result) Line 446
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncValueTaskMethodBuilder<System.IO.Pipelines.ReadResult>.SetResult(System.IO.Pipelines.ReadResult result) Line 56
> [Completed] System.IO.Pipelines.dll!System.IO.Pipelines.StreamPipeReader.ReadAsync.__Core|36_0(System.IO.Pipelines.StreamPipeReader reader, System.Threading.CancellationTokenSource tokenSource, System.Threading.CancellationToken cancellationToken) Line 271
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.IO.Pipelines.ReadResult>.AsyncStateMachineBox<System.IO.Pipelines.StreamPipeReader.<<ReadAsync>g__Core|36_0>d>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.IO.Pipelines.ReadResult>.AsyncStateMachineBox<System.IO.Pipelines.StreamPipeReader.<<ReadAsync>g__Core|36_0>d>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<System.IO.Pipelines.ReadResult>.AsyncStateMachineBox<System.IO.Pipelines.StreamPipeReader.<<ReadAsync>g__Core|36_0>d>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__12_0(System.Action innerContinuation, System.Threading.Tasks.Task innerTask) Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(System.Action action, bool allowInlining) Line 743
> System.Private.CoreLib.dll!System.Threading.Tasks.Task.RunContinuations(object continuationObject) Line 3381
> System.Private.CoreLib.dll!System.Threading.Tasks.Task<int>.TrySetResult(int result) Line 401
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.SetExistingTaskResult(System.Threading.Tasks.Task<int> task, int result) Line 443
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncValueTaskMethodBuilder<int>.SetResult(int result) Line 56
> [Completed] System.Net.Security.dll!System.Net.Security.SslStream.ReadAsyncInternal<System.Net.Security.AsyncReadWriteAdapter>(System.Net.Security.AsyncReadWriteAdapter adapter, System.Memory<byte> buffer) Line 1061
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<System.Net.Security.SslStream.<ReadAsyncInternal>d__186<System.Net.Security.AsyncReadWriteAdapter>>.ExecutionContextCallback(object s) Line 287
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<System.Net.Security.SslStream.<ReadAsyncInternal>d__186<System.Net.Security.AsyncReadWriteAdapter>>.MoveNext(System.Threading.Thread threadPoolThread) Line 324
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder<int>.AsyncStateMachineBox<System.Net.Security.SslStream.<ReadAsyncInternal>d__186<System.Net.Security.AsyncReadWriteAdapter>>.MoveNext() Line 302
> System.Private.CoreLib.dll!System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore<int>.SignalCompletion() Line 233
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<int>.SetResult(int result) Line 62
> Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.ReadAsyncInternal(System.Memory<byte> destination, System.Threading.CancellationToken cancellationToken) Line 151
> [Resuming Async Method]
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<int>.StateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.<ReadAsyncInternal>d__27>.ExecutionContextCallback(object s) Line 394
> System.Private.CoreLib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Line 183
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<int>.StateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.<ReadAsyncInternal>d__27>.MoveNext() Line 416
> System.Private.CoreLib.dll!System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder<int>.StateMachineBox<Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.<ReadAsyncInternal>d__27>.System.Threading.IThreadPoolWorkItem.Execute() Line 400
> System.Private.CoreLib.dll!System.Threading.ThreadPoolWorkQueue.Dispatch() Line 729
> System.Private.CoreLib.dll!System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart() Line 63  

This happens because `DecrementActivity` is automically closing the `ServiceChannel` but is doing it in a sync over async way. 
Note that this behavior can easily be exploited for an DoS attack.

gvkries:
When a ServiceChannel gets closed automatically when receiving is done, it will be closed synchronously by calling CloseAsync(...).GetAwaiter().GetResult(). This causes problems e.g. when using NetHttpBinding with WebSockets.

Note: This fix does not remove the sync over async call in all circumstances, but just fixes the error in #710.

Fixes #710


src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannel.cs
```diff
@@ -709,14 +709,20 @@ internal void DecrementActivity()
             }
 
             if (updatedActivityCount == 0 && _autoClose)
+            {
+                return AutoCloseAsync();
+            }
```
mconnew, line 714: Can you lift the `if (State == CommunicationState.Opened)` check to be around the call to AutoCloseAsync instead? I understand that you made a private inline method to avoid the async state machine and Task allocation when closing if it's not needed. Moving the if block to here will help with that more.

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannel.cs
```diff
@@ -764,7 +770,7 @@ private TimeoutException GetOpenTimeoutException(TimeSpan timeout)
             }
         }
 
-        internal void HandleReceiveComplete(RequestContext context)
+        internal Task HandleReceiveComplete(RequestContext context)
         {
```
mconnew, line 774: Can you rename the method to have the Async suffix? Are there other call points which now need to await this method elsewhere in the code?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -857,7 +857,7 @@ private void ProcessMessageCleanup(MessageRpc rpc)
                 {
                     try
                     {
-                        rpc.Channel.DecrementActivity();
+                        rpc.Channel.DecrementActivity().GetAwaiter().GetResult();
                     }
```
mconnew, line 861: This method is called from ProcessError which is async so you can make this method async and await here.

mconnew:
There's another trick you can do to improve things a lot. We can move the WebSocket receive continuation code into the separate CoreWCF thread pool mechanism. It's as simple as modifying WebSocketMessageSource.ReceiveAsync to wrap the entire method in this using block:
```c#
public async Task<Message> ReceiveAsync(CancellationToken token)
{
    using (TaskHelpers.RunTaskContinuationsOnOurThreads())
    {
        // Existing implementation body of ReceiveAsync
    }
}
```
We have a mechanism to basically hijack an IO Thread from the CLR and use that for running task continuations. The purpose of this is that there can be times that application code can clog up the thread pool (I know in this case it was CoreWCF, but any code can do that) and this enables us to make forward progress on IO. You can do the same for `WebSocketTransportDuplexSessionChannel.OnSendCoreAsync` too, although that's not as important.

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannel.cs
gvkries: There is only one call site to this method and it's internal. If I rename this, what about all the other methods, e.g. `DecrementActitivity`, `ProcessMessageCleanup` in the `ImmutableDispatchRuntime` etc. Should I rename all of those too, at least the internal ones?

mconnew:
I'm suggesting this in addition to the changes you made, not instead of. The problem you can get into is you block a large number of threads waiting for the async work to complete. The async work completes and the continuation to continue execution of an async method (deep down in the stack) gets scheduled to the thread pool. The problem is the thread pool can't run the continuation because there aren't any threads available, which means the sync over async blocking wait won't complete. This helper code we have pushes the async work to complete on an IO Thread instead. The continuation after an async operation completes (that's deep down in the call stack) gets scheduled to our own scheduler which runs it on the IO Thread. It doesn't get stuck waiting to run in the exhausted general-purpose thread pool. So now the Task can complete, and the blocked synchronous thread gets unblocked and continues execution.  

This basically makes a bad coding practice (sync over async) have lighter consequences. I find it interesting how sometimes an issue won't affect anyone for a long time, and then you start seeing multiple cases of it. We've had an internal team at Microsoft hit the sync over async problem with the WCF client (their issue this time, not a problem in the WCF client) and I've coded a workaround for them using this exact technique, it's currently in PR in the WCF repo. This technique that I developed for WCF/CoreWCF can come in handy.

src/CoreWCF.Primitives/src/CoreWCF/Channels/ServiceChannel.cs
mconnew: If they are async, they should have the Async suffix so rename anything internal which you are touching that is async and doesn't match that name.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
gvkries: Yes, I saw that but I wasn't sure how far I should go. This method calls some additional `Close` methods which than can be made async as well.

gvkries: I went ahead and removed some more cases of sync over async in `ProcessMessageCleanup`. Please check if these changes should be incldued.

gvkries:
I understand what you are suggesting. But I'm not feeling that I grok this part deeply enough to be sure to make those changes.  At least in the receive case, most Tasks are not scheduled back on the context because of `ConfigureAwait(false)` and it also uses `async void` - what kind of impact will this have? Maybe a separate issue/PR is better for these changes. 
But if you want to have that added, I can do it anyway 😊

gvkries:
Btw, this test `NetTCPRequestReplyWithTransportMessageEchoStringDemuxFailure` is annoying... 🙄 Works on my machine...

mconnew:
I believe we've tracked down the cause of this. One the demux failure, the fault message is being sent, but there was a missing await. Then the channel is closed. There's a race condition that the close closes the HttpContext (and so aborts the response) before the response has been sent. The PR fixing this got blocked by an investigation into why HTTP.SYS with SSL doesn't work in Azure DevOps. We're going to disable the HTTP.SYS tests for now until the DevOps windows agent bug is fixed. Once that's done, the fix will get merged and that annoying flakey test will should stop being flakey.

mconnew:
ConfigureAwait(false) causes the continuation to be scheduled back to the main thread pool which is actually the opposite of what we want to do here. The scenario where you would normally use ConfigureAwait is when you have a limited resource SynchronizationContext such as a single threaded sync context in a UI app such as WPF or WinForms. In that case you would have some code which starts on the UI thread as the result of an action like a button press, calls something async such as an http request, then updates the UI based on the result. After you await the http request, you want to return back to the UI thread to be able to make UI changes (UI apps require UI changing calls to be done on the UI thread). But the internal implementation of the http request will consist of multiple asynchronous calls. E.g. an async socket connection, an async send for the headers, async send for a body, multiple async receives to receive the entire response. You don't want the continuation of each of these async calls to happen on the UI thread for many reasons. It might be busy with something else which slows things down, or you might have a lot of things going on and then will have contention on that single thread. So the internals of the http client will use ConfigureAwait(false) on every IO call to have it drop the sync context and post continuations to the thread pool. This is to free up the single threaded UI sync context.  
We want to do the opposite in CoreWCF. If the general purpose thread pool is contended (lots of blocking calls consuming threads), then CoreWCF IO can't complete as the continuations don't have an available thread pool thread to run on. By using out own synchronization context, we cause continuations to be posted to our own thread scheduler which hijacks an IO thread. This means our own IO can make progress and complete tasks and our async methods can complete even though the main thread pool is overloaded and contended.  
This isn't new to CoreWCF, the same trick is done in WCF on .NET Framework. The difference is that we aren't using Task's there. When you use Task's, there's a tendency to get kicked back to the thread pool for continuations more often. APM (Begin/End style async calls) don't capture the sync context and run continuations on them so when an IO operation completes, the callback executes on an IO thread (where the completion happens). But asp.net core has a habit of bumping your continuation back to the thread pool because it uses Task's. A few strategic uses of `RunTaskContinuationsOnOurThreads` gets things back running where we want them to.

gvkries:
Thank you very much for taking the time to summarize this again. I may be mistaken, but I thought that ConfigureAwait(false) just skips scheduling the remainder of the method back to the original context. It should just continue on the current thread, similar to the old APM End-methods. This may happen on a ThreadPool thread of cause, but most often it should already be an IO thread in networking calls.
I have the feeling that this is quite important for performance in hot paths like the receive loops, because it avoids posting a callback back and also some additional checks if a SynchronizationContext is in place etc.. And the more I think about this, I'm not convinced to proactively add `RunTaskContinuationsOnOurThreads` within the send/receive methods. But as said, I do not know enough about CoreWCF.

mconnew:
I'm going to merge this asis as `RunTaskContinuationsOnOurThreads` isn't needed to directly address this. But I am going to talk a little about the behavior of `ConfigureAwait`. Every await in a call hierarchy chooses its own behavior, so using `ConfigureAwait` only affects the immediate method call. When you have multiple methods being called inside each other, you end up having to using ConfigureAwait at every level and on every call and end up littering the code everywhere with `ConfigureAwait`. All it takes is to miss one await and you can easily end up losing the benefit of `ConfigureAwait`. I'll provide some example code to demonstrate.
```c#
public async Task FooAsync()
{
    await FooCoreAsync();
    await BarAsync();
}

public async Task FooCoreAsync()
{
    await _webSocket.ReceiveAsync(...).ConfigureAwait(false);
    // More code
}

public async Task BarAsync()
{
    await _webSocket.SendAsync(...).ConfigureAwait(false);
    // More code
}
```
When you call `FooAsync`, the first thing it does is call `FooCoreAsync`, which then calls `ReceiveAsync` on the websocket. This call has a `ConfigureAwait(false)` used, which will do one of two things. If the `ReceiveAsync()` call completes synchronously (we already had data waiting in the socket buffer), the "More Code" will execute on the original calling thread. If instead it goes async, when the `ReceiveAsync` call completes it does so on an IO thread as a native callback happened on an IO Completion Port. The `ConfigureAwait(false)` means execution of `More Code` will happen on this IO thread. But once `FooCoreAsync` completes, something different happens. The completion registered against the task returned from `FooCoreAsync` will be executed when that method completes, but because `ConfigureAwait(false)` wasn't used for the call to `FooCoreAsync` in `FooAsync`, the continuation which will run `BarAsync` gets scheduled to the main thread pool. In the case where you have thread pool contention, you might be waiting a long time for `BarAsync` to be called. The "typical" solution is to add `ConfigureAwait` everywhere in your code. There is a non-zero cost to using `ConfigureAwait`.  

If you miss a single location where you need to use `ConfigureAwait`, if that method call ends up completing asynchronously, that continuation hops back across to the thread pool, even if the callback happened from an IO thread. From then on none of your `ConfigureAwait` calls helps keep work off of the thread pool until you next do some more IO and get a completion happening on an IO thread. You've lost the benefit of using `ConfigureAwait` (keeping execution on an IO thread) but kept the cost of all those `ConfigureAwait` calls.  

Then you have the problem of asynchronously waiting on non-IO. For example, asynchronously waiting on a `SemaphoreSlim` (to use as a naive async lock mechanism). Even if you use `ConfigureAwait(false)`, the continuation is still getting executed on the thread pool as the `SemaphoreSlim.Release` call happened on a theead pool thread.  

With the `RunTaskContinuationsOnOurThreads`, we keep execution off of the thread pool, even in cases where it's a non-IO completion. We don't need to make sure we haven't missed decorating any call points with `ConfigureAwait(false)` as it's ambient and will apply to all child calls. It means you don't need to be constantly vigilant to make sure you never miss using `ConfigureAwait(false)` and allows more forgiving code to be written.  

We should definitely measure the performance of using `RunTaskContinuationsOnOurThreads` as there are some call patterns where it would result in worse performance. I don't believe this is a case where it would hurt performance, but it's wise to measure before applying it. Generally though you will see a performance improvement when under load due to the faster scheduling of code on an IO thread (Windows treats IO threads as something special and won't preemptively swap the thread out unless it yields).

gvkries:
Thanks again for all your input here. I really appreciate the work you put into all of this, so I apologize if I seem to be a nay-sayer here. I understand the behavior of `ConfigureAwait` and its complications, my original questioning was more about the `async void` call. I remember this interacts with a `SynchronizationContext` differently and I don't remember the details.
Now I've quickly skimmed over the `IOThreadScheduler` and I'm now even more hesitant as it looks like each callback will spawn a new thread on non windows OSes. That is problematic in tight receive and send loops in my mind.

mconnew:
That's not what it does. If the IOThreadScheduler queue is empty, it creates a new thread. Once that thread has executed the work item, if nothing else has been scheduled, the thread exits. While that thread is processing the work item, if another item comes along and gets placed in the queue, the thread won't exit and will instead pick up the next work item to execute. You shouldn't have more than one IOThreadScheduler thread running at a time. The exception to this is if the work queue is full, then a second instance is created with a new queue and a new thread launched to handle that work load. The first instance won't have any more work added to it and the thread will be removed once all the pending work has been completed.  
On Windows, we hijack an IO Thread by posting an IO Completion item to the IOCP to process items from the queue instead of creating a new thread.
On Linux, we could improve things by holding on to the thread for a while after the queue has been drained. Getting the signaling correct to know the thread has exited (or is about to exit) after a timeout will be tricky and there's a high risk of race conditions resulting in a work queue not getting drained. The only time when you are creating and exiting (just one thread at a time) a lot is when the server is under low load. If it's under any kind of load, the queue will generally have a couple of items pending and the Thread will persist. It's basically a self solving problem, the high cost operation is only done when there's close to zero load, but apply so load and the high cost operation stops happening.

gvkries:
Very cool, thanks for the explanation.

PR #729, opened 2022-06-23
Fixing possible null reference exception during OpenAPI doc gen.

JonathanHopeDMRC:
This fixes a possible null reference exception during OpenAPI doc gen.


mconnew:
Have you verified that the Method property now being nullable isn't going to cause a NRE elsewhere?

JonathanHopeDMRC:
I did, but you bring up  a fair point. I added a test.

mconnew:
Sorry this took so long to get to. I'll bundle the fix with some other pending stuff when I get a chance.

PR #733, opened 2022-06-28
Fix #732: Services without default constructor generate misleading error message

### LINKED ISSUE #732: Services without default constructor generate misleading error message
After adding a non-default constructor to a service registered only with `IServiceBuilder.AddService<ServiceType>`, the following error message pops up:
`add a default constructor to the type, or pass an instance of the type to the host.`
However there is no way to "pass an instance of the type to the host."

Suggested message: 
`add a default constructor to the type, or register a service of the type (e.g. IServiceCollection.AddSingleton<ServiceType>())`

tom-englert:
Fix #732: Services without default constructor generate misleading error message


mconnew:
Thank you

Issue #164, opened 2020-07-04
Will there be MSMQ Support?


Grauenwolf:
Now that .NET's MSMQ drivers have been open-sourced, are there plans to incorporate NetMsmqBinding? 

ref: https://github.com/microsoft/referencesource/pull/141

mconnew:
There isn't any immediate plans. It would be great to eventually support pretty much everything that WCF supports, but that's a really long list. Currently CoreWCF is based on ASP.NET Core 2.1 to make life easier for an initial port from WCF on .NET Framework. It allows you to move your service to CoreWCF and still run on Framework before you have all your dependencies and/or implementation code moved across. So CoreWCF v1 will be running on ASP.NET Core 2.1.  
The hosting model has been changed in later versions of ASP.NET Core to make it easier to have non-socket based transports such as MSMQ. After v1 is released and the dust settles, I will be focusing on moving to the latest version of ASP.NET Core. Then we can start doing some interesting things such as support Net.Tcp port sharing, and MSMQ.  
There is one potential issue with MSMQ. As MSMQ isn't a single hop transport (the client directly talks to the service), security becomes more of an issue. People often use full message security when the message itself is signed and encrypted. Because some dependencies are missing in .NET Core, we can't yet do full message security. If you are unsure whether you are using full message security, if you can provide your binding config I should be able to work it out.

weloytty:
Another big issue is that (having been in the System.Messaging code quite a bit getting it to run under netcore) is that it relies heavily on mqrt.dll, which is never going to run on anything other than windows.  Getting System.Messaging running on netcore on windows was fairly simple, but making it run on any other platform would be a huge undertaking.

Grauenwolf:
I don't think anyone is seriously thinking about MSMQ for future projects, but they need it for upgrading legacy systems to Core.

mconnew:
Everything I've done has been with an idea to embrace other technologies. So if MSMQ support does get to the front of the line for bringing over, I would make an abstract messaging transport implementation with MSMQ being an implementation with the idea being it would be trivial to add RabbitMQ support or even one of the cloud based queue services.  
There are many scenarios where using a queued transport is the best option. If it wasn't, the main cloud providers wouldn't be providing a message queue service.

Grauenwolf:
> I would make an abstract messaging transport implementation with MSMQ being an implementation with the idea being it would be trivial to add RabbitMQ support or even one of the cloud based queue services

Yes please. 

[...]

PR #736, opened 2022-06-30
Design document for generic queue transport in CoreWCF

birojnayak:
Design document for generic queue transport in CoreWCF #164


Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+# Class Diagram
+![CoreWCF Generic Queue](/Documentation/DesignDocs/generic_queue_class_diagram.png?raw=true)
+
+# Class Details
+#### IQueueMiddlewareBuilder and QueueMiddlewareBuilder
+- IQueueMiddlewareBuilder provides signatures for 2 important methods, ```Use``` and ```Build```. 
+```csharp
+internal interface IQueueMiddlewareBuilder 
+{ 
+   IServiceProvider HandshakeServices { get; set; }
```
mconnew, line 16: The name HandshakeServices came from the NetTcp implementation which has a handshake performed on connection. This should be named something different, maybe QueueServices or maybe just Services.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+```csharp
+internal interface IQueueMiddlewareBuilder 
+{ 
+   IServiceProvider HandshakeServices { get; set; } 
+   IDictionary<string, object> Properties { get; } 
+   IQueueMiddlewareBuilder Use(Func<QueueMessageDispatcherDelegate, QueueMessageDispatcherDelegate> middleware); 
+   QueueMessageDispatcherDelegate Build(); 
+} 
+```
+- ```Use``` adds middleware delegates to the Queue handshake pipeline
```
mconnew, line 22: Drop the word handshake.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+internal interface IQueueMiddlewareBuilder 
+{ 
+   IServiceProvider HandshakeServices { get; set; } 
+   IDictionary<string, object> Properties { get; } 
+   IQueueMiddlewareBuilder Use(Func<QueueMessageDispatcherDelegate, QueueMessageDispatcherDelegate> middleware); 
+   QueueMessageDispatcherDelegate Build(); 
+} 
+```
+- ```Use``` adds middleware delegates to the Queue handshake pipeline 
+- ```Build``` builds the delegate used by this application to process fetching request hold up in queue and process them.
```
mconnew, line 23: ...to process requests received from the queue.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+#### QueueMiddleWareBuilderExtension
+- QueueMiddleWareBuilderExtension defines ```UseMiddleware``` method which specifically search for ```InvokeAsync``` method in each middleware class and construct it as delegrate to be invoked middleware pipeline. Simillar to the way implemented in UseMiddlewareFramingConnectionHandshakeExtensions for NetTCP.
```
mconnew, line 25: Typo: delegrate->delegate  
..to be involed by the middleware pipeline. This is similar to the NetTcp class UseMiddlewareFramingConnectionHandshakeExtensions and the ASP.NET Core UseMiddlewareExtensions.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+```csharp
+public class QueueMiddleWareBuilderExtension
+{
+    internal const string InvokeAsyncMethodName = "InvokeAsync";
+    public static IQueueMiddlewareBuilder UseMiddleware<TMiddleware>(this IQueueMiddlewareBuilder app, params object[] args)
```
mconnew, line 30: You don't need to stub out the methods or provide internal details. Should this class be static?

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
+# Objective 
+The objective of this document is to provide details on how to build a generic queue transport inside CoreWCF so that it can enable various queue binding. This document is based on the discussion outlined here https://github.com/CoreWCF/CoreWCF/issues/164 .
```
mconnew, line 2: Nit: various queue binding**s**

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Why are we building
+When WCF was build there were not many queing technique and there was no concept of cloud. With multiple cloud providers supporting queing technique and many queing technique widely available, we wanted to build a generic transport, so that various queuing mechanism can be supported in CoreWCF.
```
mconnew, line 5: Spelling/grammar/phrasing:  
When WCF was built there were not many queuing technologies....

mconnew: Those same changes should be made through the entire paragraph and not just the one sentence.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
+# Objective 
+The objective of this document is to provide details on how to build a generic queue transport inside CoreWCF so that it can enable various queue binding. This document is based on the discussion outlined here https://github.com/CoreWCF/CoreWCF/issues/164 .
+
+# Why are we building
```
mconnew, line 4: Cleaner section title would be "Problem overview" as current title is a sentence fragment.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+   IServiceProvider QueueServices { get; set; } 
+   IDictionary<string, object> Properties { get; } 
+   IQueueMiddlewareBuilder Use(Func<QueueMessageDispatcherDelegate, QueueMessageDispatcherDelegate> middleware); 
+   QueueMessageDispatcherDelegate Build(); 
+} 
+```
+- ```Use``` adds middleware delegates to the Queue pipeline 
+- ```Build``` builds the delegate used by this application to process requests received from the queue.
+#### QueueMiddleWareBuilderExtension
+- QueueMiddleWareBuilderExtension defines ```UseMiddleware``` method which specifically search for ```InvokeAsync``` method in each middleware class and construct it as delegate to be invoked by the middleware pipeline. This is similar to the NetTcp class UseMiddlewareFramingConnectionHandshakeExtensions and the ASP.NET Core UseMiddlewareExtensions..
```
mconnew, line 25: I just noticed you use triple backticks when inserting inline code fragments. Inline you use a single back tick \`like this\` to produce `like this`.  Can you fix throughout the document. I just noticed we don't define `QueueMessageDispatcherDelegate` anywhere. You should probably add it's definition in the code snippet which contains the `IQueueMiddlewareBuilder` delegate.

```suggestion
The `QueueMiddleWareBuilderExtension` class defines the `UseMiddleware` method which searches the provided `TMiddleware` generic parameter type for an `InvokeAsync` method with a method signature that matches the `QueueMessageDispatcherDelegate` delegate. It expects a constructor to exist which takes as the first parameter an instance of a `QueueMessageDispatcherDelegate` delegate, and any remaining parameters to match the optional additional args. The `QueueMessageDispatcherDelegate` delegate passed to the constructor is the next middleware to call when the middleware doesn't terminate the dispatch chain. This is similar to the NetTcp class `UseMiddlewareFramingConnectionHandshakeExtensions`, and the ASP.NET Core `UseMiddlewareExtensions` class.
```

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+```csharp
+public class QueueMiddleWareBuilderExtension
```
mconnew, line 27: I just checked the the NetTcp and asp.net core equivalents are static classes, and this needs to be static as it contains extension methods.
```suggestion
public static class QueueMiddleWareBuilderExtension
```

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+    }
+    
+    public static IQueueMiddlewareBuilder UseMiddleware(this IQueueMiddlewareBuilder app, Type middleware, params object[] args)
+    {
+        //TODO
+     }
+    }
+```
+#### QueueServiceWebHostBuilderExtension
+- QueueServiceWebHostBuilderExtension is the WebHostBuilder extension which stiches all the middleware pipeline and ```QueuePollingService``` which will be called during service startup for queuing mechanism to work. e:g:-
```
mconnew, line 42: Make sure any specified types are quoted as inline code. Eq `QueueServiceWebHostBuilderExtension` as opposed to QueueServiceWebHostBuilderExtension

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+- ```Build``` builds the delegate used by this application to process requests received from the queue.
+#### QueueMiddleWareBuilderExtension
+- QueueMiddleWareBuilderExtension defines ```UseMiddleware``` method which specifically search for ```InvokeAsync``` method in each middleware class and construct it as delegate to be invoked by the middleware pipeline. This is similar to the NetTcp class UseMiddlewareFramingConnectionHandshakeExtensions and the ASP.NET Core UseMiddlewareExtensions..
+```csharp
+public class QueueMiddleWareBuilderExtension
+{
+    internal const string InvokeAsyncMethodName = "InvokeAsync";
+    public static IQueueMiddlewareBuilder UseMiddleware<TMiddleware>(this IQueueMiddlewareBuilder app, params object[] args)
+    {
+         return app.UseMiddleware(typeof(TMiddleware), args);
```
mconnew, line 32: Remove implementation detail and leave it as definition only. Just put a semi-colon at the end of the method name like you would an abstract method, this doesn't need to be compilable.

Documentation/DesignDocs/corewcfqueue-design.md
mconnew: You should provide the class design here. Also the example should be updated to .NET 6 style syntax. See the sample app [here](https://github.com/CoreWCF/samples/blob/main/Scenarios/Logging-and-Dependency-Injection/LoggingSample/Program.cs) for an example.

mconnew: We don't need to add CoreWCF into the method name. It's not `UseAspNetCoreKestrel`. We should just be bold and call it `UseQueues`. This is also the wrong place to call the method, as Use verbs are applied to builders, Add verbs to collections. I've been trying to work out where we should apply the `UseQueues` method and I'm beginning to think for CoreWCF 2.0 we need to take a bit of a step back to look at the api's and how you configure CoreWCF in the future. As of right now we have NetTcp (which uses UseConnectionHandler and is based on Kestrel) and HTTP based transports, so everything builds off of a WebHost. We're about to add queue's, and then we have named pipe's coming up, neither of which will build on top of Kestrel. Then there's TCP Port Sharing, and the IIS hosted model in the future where the IIS Integration IServer type is used, both of which will have us using sockets outside of an IServer. There's also been some demand for UDP based bindings too. There's a new model in .NET 6 where a Web based application (mean HTTP based) is just another IHostedService implementation. I think we need to start thinking about configuring CoreWCF transports outside of the IWebHostBuilder world (as you might not have an HTTP endpoint) and have some mechanism build around `IHostBuilder`. I don't want to hold up merging this based on that (as it will take some designing), this is more of a heads up that this portion is likely to evolve. It shouldn't affect the rest of the design as IHostedService is still the right way to go. Basically everything should become an IHostedService and this is around how do we add/configure the various transports in the host. We also need to solve the problem of not wanting to configure your services and endpoints in multiple places.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+```csharp
+internal sealed class Startup
+{
+    public void ConfigureServices(IServiceCollection services)
+    {
+      services.AddServiceModelServices();
+      services.UseCoreWCFQueue();
+ }
+ ```
+ - It first keeps an instance of IQueueMiddlewareBuilder for chaining all the middleware components to be used
```
mconnew, line 52: This isn't making a lot of sense. Do you mean it adds an implementation of IQueueMiddlewareBuilder to DI so that a Queue based binding is able to build the transport layer stack by adding middleware? It will add other needed components to DI such as QueueHandShakeMiddleWare and QueuePollingService which are necessary to build and run a Queue based service.  

We also need to rename QueueHandShakeMiddle. The NetTcp one is called that because there is a connection handshake which is completed in stages (here's my endpoint, here's my framing mode, here's the stream upgrade mime type I wish to upgrade etc). There is no handshake with Queue. One name that seems appropriately short but descriptive is (and I'm not the only person who can have an opinion on this, so feel free to suggest other names) QueueDispatchMiddleware? QueuePollingService seems like a good name, although QueuePollingHostedService might be a better fit as it describes what kind of service it is (it implements IHostedService).

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+ - Afterwards it initalizes the ```QueueHandShakeMiddleWare``` and ```QueuePollingService```.
+#### QueueHandShakeMiddleWare
+- The QueueHandShakeMiddleWare has 2 methods, ```InitQueueMiddleWare``` and ```Build```.
```
mconnew, line 55: We need to provide the class description as code. The name `InitQueueMiddleware` is very redundant. Also as it's called from the constructor, it's an implementation detail and should be a private method. There's no reason for user code to call it.  
Please specify the type of delegate which is returned.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+- The InitQueueMiddleWare initalizes middlewares to use and is called from constructor
+- Build method builds the middleware pipeline and returns delegate.
+#### QueueBaseBindingElement , QueueTransportContext and IQueueTransport
+- IQueueTransport holds the signature for method ```ReceiveQueueMessageContextAsync```.
```
mconnew, line 59:
```suggestion
- `IQueueTransport` defines a single method, `ReceiveQueueMessageContextAsync`. The QueuePollingService will call this method to fetch the next incoming queue message.

File: Documentation/DesignDocs/corewcfqueue-design.md [hunk 16]
```diff
@@ -0,0 +1,190 @@
...
+- Build method builds the middleware pipeline and returns delegate.
+#### QueueBaseBindingElement , QueueTransportContext and IQueueTransport
+- IQueueTransport holds the signature for method ```ReceiveQueueMessageContextAsync```. 
+```csharp
+   public interface IQueueTransport
+    {
+        Task<QueueMessageContext> ReceiveQueueMessageContextAsync(CancellationToken cancellationToken);
+    }
+```
+- QueueBaseBindingElement is an abstract class with ```CreateQueueTransport``` as abstract method. Any concrete QueueTransport part of the CoreWCF, needs to implement and provide a concrete implementation of IQueueTransport.
```
mconnew, line 66: The second sentence needs to rewording.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+            : base(elementToBeCloned)
+        {
+        }
+
+        public abstract IQueueTransport CreateQueueTransport(BindingContext context);
+        public virtual int MaxPendingReceives { get { return 1; } }
+    }
+ ```
+- The concrete implementation of IQueueInternalTransport will be used by CoreWCF to retrieve messages from the queue storage service. 
+- When ReceiveQueueMessageContextAsync method is called, it should return a a concrete implementation of QueueMessageContext object.
```
mconnew, line 84: Weren't we going to try to have QueueMessageContext be usable for many scenarios as is and the only reason to derive it would be if a particular implementation needs to do something different?

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+- QueueTransportContext is a holder class to store InternlTransport, QueueBaseBindingElement, ServiceDispatcher
```
mconnew, line 85: I thought we fixed the naming here. InternalTransport is the wrong name, I thought it was the just the IQueueTransport?

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+#### QueuePollingService
+- QueuePollingService is the center for queue implementation, it starts when the polling messages would start and when it would end.
```
mconnew, line 87: For the second half of this sentence, maybe "it's started when the host starts and continues fetching messages until the host is stopped."

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+- The concrete implementation of IQueueInternalTransport will be used by CoreWCF to retrieve messages from the queue storage service. 
+- When ReceiveQueueMessageContextAsync method is called, it should return a a concrete implementation of QueueMessageContext object.
+- QueueTransportContext is a holder class to store InternlTransport, QueueBaseBindingElement, ServiceDispatcher
+#### QueuePollingService
+- QueuePollingService is the center for queue implementation, it starts when the polling messages would start and when it would end. 
+- At the initialization of the class, 
+```csharp
+private async Task Init()
+{
+   foreach (Type serviceType in serviceBuilder.Services)
```
mconnew, line 92: Nit, indentation

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+   var tasks = _queueTransportContexts.Select(_queueTransport => StartFetchingMessage(_queueTransport));
+   await Task.WhenAll(tasks); // handle cancellation
+}
+private async Task StartFetchingMessage(QueueTransportContext queueTransport)
+{
+    List<Task> tasks = new List<Task>();
+    for (int i = 0; i < queueTransport.QueueBindingElement.MaxPendingReceives; i++)
+    {
+        tasks.Add(FetchAndProcessAsync(queueTransport, _cancellationTokenSource.Token));
+
```
mconnew, line 144: Formatting, you have extra blank lines in some of these methods

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+        queueMessageContext.QueueTransport = queueTransport;
+        await queueTransport.QueueHandShakeDelegate(queueMessageContext);
+        
+    }
+}
+        
+```
+- For StartAsync
+   - First we create a list of tasks```(StartFetchingMessage)``` for each ```QueueTransportContext``` (populated during Init). 
+   -  Each Task uses internal transport concurrency (MaxPendingReceives) to create another list of subtask to fetch message from each internal transport concurrently```( By FetchAndProcessAsync)
```
mconnew, line 163: You have an unpaired backtick on this line for code formatting.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,190 @@
...
+ - When the ASP.NET Core WebHost starts, it calls IHostedService.StartAsync on the QueuePollingService, which creates list os tasks (StartFetchingMessage) for each transport and run them in a loop.
+ -  Each task again create a list of subtasks based on the QueueTransport  (concurrent property) to retrieve message context(QueueMessageContext) (until server stops)
+ - The fetched QueueMessageContext is passed to the middleware. The first middleware layer reads the message bytes from a PipeReader and decodes it to a Message object.
+ - The Message object is put into QueueMessageContext object and passed to next layer QueueProcessMessage.
+ - The QueueProcessMessage, read the message and invokes DispatchAsync on the IServiceChannelDispatcher to call the service based on the binding and message header information.
+ - Once the service is processed, QueueMessageContext sends result to concrete transport to notify that message is being processed. It will either be a NullMessage meaning everything was successful, or  it will be a message fault which indicates an error processing the request.
+
+# Nofiying Client
```
mconnew, line 187: Did you mean notifying?

Documentation/DesignDocs/corewcfqueue-design.md
birojnayak: do you want me to add this ?

Documentation/DesignDocs/corewcfqueue-design.md
birojnayak: Agree, this will evolve as we build more and more layers..

Documentation/DesignDocs/corewcfqueue-design.md
birojnayak: QueueFetchAndDispatchMiddleware ? => As it describes the fact that it first fetches the queue by using the transport and dispatch service dispatcher to process.
QueuePollingHostedService would be better as it clearly speaks about the IHostedService

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+            return innerDispatcher;
+        }
+
+    }
+ ```
+- The concrete implementation of `IQueueTransport` will be used by CoreWCF to retrieve messages from the queue storage service. 
+- When `ReceiveQueueMessageContextAsync` method is called, it returns an instance of `QueueMessageContext` object.
+- `ServiceDispatcher`, `QueueBaseTransportBindingElement`, `QueueMessageDispatcherDelegate`, and `QueueTransportPump` are all stored in `QueueTransportContext`. 
+-`QueueTransportPump` is an abstract class with `StartPumpAsync` and `StopPumpAsync` task. 
+- Any queue transport which is using push mechanism (e:g:- RabbitMQ) needs to implement this class.
```
mconnew, line 149: Inconsistent spacing between - and text on this line. Also, StartPumpAsync and StopPumpAsync are methods, not tasks. Maybe refer to them as "Task returning methods". Technically calling them async methods isn't accurate as they could be implemented synchronously, they just need to return a Task.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+
+}
+```
+- for each service type (which is fetched from DI), service dispatcher is created.
+- for each service dispatcher, validation is made to ensure the service dispatcher binding element collection property has a transport binding element of type `QueueBaseTransportBindingElement`.
+- using the `QueueBaseTransportBindingElement` which is fetched from above step, `BuildQueueTransportPump` method is called to get the `QueueTransportPump`.
+ - In the end, for every service dispatcher, a `QueueTransportContext` object is created which stores objects of type `QueueTransportPump`, `ServiceDispatcher`, `QueueBaseTransportBindingElement`,`MessageEncoderFactory` and `QueueHandShakeDelegate`.
+- Since `QueuePollingService` implements `IHostingService`, it implements the `StartAsync` and `StopAsync` methods.
```
mconnew, line 234: Extra space at start of line

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+{
+    await queueTransport.QueuePump.StartPumpAsync(queueTransport, CancellationToken.None);
+}
+
+```
+- For StartAsync
+ - First we create a list of tasks`StartFetchingMessage` for each `QueueTransportContext` (populated during Init).
+ - `QueueTransportPump` from `QueueTransportContext` is used by each `StartFetchingMessage` to call the `StartPumpAsync` method.
+ - If the `QueueTransportPump` is based on a push message mechanism, it awaits until a message is received and invokes service as soon as the message is received.
+ - If the `QueueTransportPump` is using `DefaultQueueTransportPump`, it creates multiple concurrent tasks to invoke `FetchAndProcessAsync` method based on concurrency level.
```
mconnew, line 253: If the `QueueTransportPump` is based on a push message mechanism, the provided implementation awaits...  

For push, the implementer needs to provide a concrete implementation instead of proving an in-box implementation. Without this small change in wording, I think it might come across as there's one in box which does this already.

mconnew: I also think saying "awaits" implies an implementation which might not be the case. With the EventHandler mechanism some libraries use you aren't awaiting anything. Maybe replace awaits with waits as that's more generic, and awaiting is a type of waiting, as is an EventHandler being executed. I know that's a lot of words to say awaits -> waits, but wanted to give my rationale.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+ -The `FetchAndProcessAsync` method of QueueTransport (concrete transport) calls `ReceiveQueueMessageContextAsync` to retrieve the message and pass the handle to the middleware for processing. 
+- StopAsync cancels all tasks and, as a result, the fetching of messages is stopped.
+#### QueueFetchMessage
+- `QueueFetchMessage` is the first middleware which is invoked as soon as a message is fetched from the queue.
+- The `InvokeAsync` method reads messages from pipereader and encodes them into a concrete `Message` objet and puts them back to `QueueMessageContext` and invokes the next delegate , which is `QueueProcessMessage`.
+#### QueueProcessMessage
+- Invokes the `DispatchAsync` method of the channel dispatcher by passing `QueueMessageContext` object and ending the middleware pipeline.
+- As a part of `DispatchAsync`, the service operation is called, processing of messages is done, eventually `QueueMessageContext` (inherits `RequestContext`) `ReplyAsync` method is called.
```
mconnew, line 261: ending the queue middleware pipeline and transitions to the dispatch pipeline

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+- The `ReplyAsync` mechanism is used to inform the concrete transport that a message processing has finished with a status of success or failure by invoking an Action `DispatchResultHandler`. 
+- By calling the `DispatchResultHandler` each queue providers which support timeouts or requires notification to prevent reprocessing a message can take appropriate actions.
```
mconnew, line 263: inform the transport implementation that message processing has finished with a status of success of failure. It signals the status result by invoking a `DispatchResultHandler` delegate.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+- The `InvokeAsync` method reads messages from pipereader and encodes them into a concrete `Message` objet and puts them back to `QueueMessageContext` and invokes the next delegate , which is `QueueProcessMessage`.
+#### QueueProcessMessage
+- Invokes the `DispatchAsync` method of the channel dispatcher by passing `QueueMessageContext` object and ending the middleware pipeline.
+- As a part of `DispatchAsync`, the service operation is called, processing of messages is done, eventually `QueueMessageContext` (inherits `RequestContext`) `ReplyAsync` method is called.
+- The `ReplyAsync` mechanism is used to inform the concrete transport that a message processing has finished with a status of success or failure by invoking an Action `DispatchResultHandler`. 
+- By calling the `DispatchResultHandler` each queue providers which support timeouts or requires notification to prevent reprocessing a message can take appropriate actions. 
+- The reply message sent via `DispatchResultHandler` could be a fault message, which will signal a failure to process the message, and can be used to provide an error message back to the queue provider.
```
mconnew, line 264: Flip the sentence around:  
If a queue provider requires completion notification to support timeouts or other features, they can do so by setting the `DispatchResultHandler` delegate and taking action accordingly.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+
+# Workflow Diagram
+![CoreWCF Generic Queue](/Documentation/DesignDocs/corewcf_queue.png?raw=true)
+
+
+ - During the server registrations, the QueuePollingService initializes the service dispatcher and gets the concrete transport from each queue layer transport.
+ - When the ASP.NET Core WebHost starts, it calls IHostedService. StartAsync on the QueuePollingService, creates a list of tasks (StartFetchingMessage) for each transport and runs them in a loop.
+ - Each task again creates a list of subtasks based on the QueueTransport (concurrent property) to retrieve message context(QueueMessageContext) (until server stops)
```
mconnew, line 272: Remove space between IHostedService and StartAsync, and backtick quote as code. Same with QueuePollingService and StartFetchingMessage

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+ - The fetched QueueMessageContext is passed to the middleware. The first middleware layer reads the message bytes from a PipeReader and decodes them to a Message object.
```
mconnew, line 273: Name the concurrent property with ClassName.PropertyName and backtick quote.. Remove (until server stops) from parenthesis and instead add a comma just before that phrase. Also worth mentioning until Webhost calls QueuePollingService.StopAsync as the mechanism that it will be stopped.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+ - The Message object is put into the QueueMessageContext object and passed to the next layer, QueueProcessMessage.
```
mconnew, line 274: There are multiple types of middleware so worth disambiguating by calling it the queue middleware or queuing middleware (make sure it's consistent through document.). Might be worth specifying earlier in the document where you are talking about building this middleware that this middleware pipeline will be referred to as the queue middleware.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+ - The QueueProcessMessage reads the message and invokes DispatchAsync on the IServiceChannelDispatcher to call the service based on the binding and message header information.
+ - Once the service is processed, QueueMessageContext sends the result to concrete transport to notify that the message is being processed. It will pass status (failed/success) and a full message , so that concrete queue transport can take appropriate actions.
```
mconnew, line 276: Reads the message using the binding provided MessageEncoder.

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+
```
mconnew, line 277: Once the service request, or maybe just request. "Concrete" typically makes more sense when talking about type hierarchies. Here it sounds better to say the queue transport implementation.  
It will pass a result status (failed/success) and an optional fault Message so that the queue transport implementation... (remove comma)

Documentation/DesignDocs/corewcfqueue-design.md
```diff
@@ -0,0 +1,308 @@
...
+# Notifying Client
+- Every internal transport when constructing a `QueueMessageContext` via `ReceiveQueueMessageContextAsync` method, they can pass a delegate in the form of `DispatchResultHandler` .
+- Once a message is processed and service is invoked, the control comes to `ReplyAsync` method of `QueueMessageContext`, where if `DispatchResultHandler` is defined, it's invoked by passing a) status of service invocation and b)full `QueueMessageContext`
+- Internal transport can act based on that to notify client or put into dead letter queue or trigger any custom fuctions.
```
mconnew, line 281: ...method of `QueueMessageContext`. If `DispatchResultHandler` has been set, it's invoked...  
Defined would mean whether the property or field exists on the class, not whether it holds a value. Also don't use a) and b) to list passed parameters, just list them with commas.

mconnew:
Does the class diagram need to be updated now?

birojnayak:
> Does the class diagram need to be updated now?

I have updated

mconnew:
LGTM. Can you rebase and squash your changes into a single commit then I'll merge.

PR #737, opened 2022-06-30
Fix HttpSys HTTPS tests running in Azure DevOps

mconnew:
I discovered an undocumented command in IisExpressAdminCmd which resets the certificate and bindings for the IIS dev certificate. Adding it to the pipeline and reenabling the tests.


g7ed6e (approved):

mconnew:
I got notification that the underlying issue with the hosts has been fixed, so I'm going to modify this PR to just enable the tests only and see if that's the case.

PR #752, opened 2022-07-15
Compatibility with System.ServiceModel.Web.Attributes

arontsang:
This is a little ugly how we hook into `CustomAttributeProvider`. Any suggestions are welcome.

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using CoreWCF.Channels;
 using CoreWCF.Collections.Generic;
 using CoreWCF.Dispatcher;
@@ -145,6 +147,8 @@ namespace CoreWCF.Description
                 }
             }
 
+            WebHttpServiceModelCompat.ServiceModelAttributeFixup(endpoint);
+
             if (HelpEnabled)
             {
                 HelpUri = new UriTemplate(HelpPage.OperationListHelpPageUriTemplate).BindByPosition(endpoint.ListenUri);
@@ -262,6 +266,8 @@ namespace CoreWCF.Description
 
             AddServerErrorHandlers(endpoint, endpointDispatcher);
         }
+
+
 
         public virtual void Validate(ServiceEndpoint endpoint)
         {
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs
@@ -267,8 +267,6 @@ namespace CoreWCF.Description
             AddServerErrorHandlers(endpoint, endpointDispatcher);
         }
 
-
-
         public virtual void Validate(ServiceEndpoint endpoint)
         {
             if (endpoint == null)
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
@@ -21,7 +21,6 @@ namespace CoreWCF.Description;
 /// </summary>
 internal static class WebHttpServiceModelCompat
 {
-
     public static void ServiceModelAttributeFixup(ServiceEndpoint endpoint)
     {
         foreach (OperationDescription operationDescription in endpoint.Contract.Operations)
@@ -43,17 +42,16 @@ internal static class WebHttpServiceModelCompat
         var convertedAttributes = AttributeConverters.Convert(attributes);
         foreach(var converted in convertedAttributes)
             od.OperationBehaviors.Add(converted);
-
     }
 
     private static class AttributeConverters
     {
 
-        private static readonly IReadOnlyDictionary<string, IAttributeConverter> s_attributeConverters =
-            new Dictionary<string, IAttributeConverter>
+        private static readonly IReadOnlyDictionary<string, Func<Attribute, IOperationBehavior?>> s_attributeConverters =
+            new Dictionary<string, Func<Attribute, IOperationBehavior?>>
             {
-                { "System.ServiceModel.Web.WebGetAttribute", new AttributeConverter<WebGetAttribute>() },
-                { "System.ServiceModel.Web.WebInvokeAttribute", new AttributeConverter<WebInvokeAttribute>() }
+                { "System.ServiceModel.Web.WebGetAttribute", AttributeConverter<WebGetAttribute>.Convert },
+                { "System.ServiceModel.Web.WebInvokeAttribute", AttributeConverter<WebInvokeAttribute>.Convert }
             };
 
         public static IEnumerable<IOperationBehavior> Convert(IEnumerable<Attribute> attributes)
@@ -69,35 +67,28 @@ internal static class WebHttpServiceModelCompat
         {
             if (s_attributeConverters.TryGetValue(attribute.GetType().FullName, out var converter))
             {
-                return converter.Convert(attribute);
+                return converter.Invoke(attribute);
             }
 
             return null;
         }
     }
 
-    private interface IAttributeConverter
-    {
-        IOperationBehavior? Convert(Attribute attribute);
-    }
 
-    private class AttributeConverter<TOut> : IAttributeConverter where TOut : class, IOperationBehavior
+    private static class AttributeConverter<TOut> where TOut : class, IOperationBehavior
     {
         private static readonly ConcurrentDictionary<Type, Func<Attribute, TOut?>> s_converterCache = new();
 
         private static readonly MethodInfo s_buildDynamicMethodInfo = typeof(AttributeConverter<TOut>)
             .GetMethod(nameof(BuildDynamic), BindingFlags.Static | BindingFlags.NonPublic)!;
 
-
-        public IOperationBehavior? Convert(Attribute attribute)
+        public static IOperationBehavior? Convert(Attribute attribute)
         {
             var type = attribute.GetType();
-
             if (!s_converterCache.TryGetValue(type, out var converter))
             {
                 converter = s_converterCache[type] = Build(type);
             }
-
 
             return converter(attribute);
         }
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
@@ -1,0 +1,215 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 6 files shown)


mconnew:
This isn't the correct approach. You can't expose internal api's. I deep a quick deep dive to understand how the attributes get exposed to CoreWCF.WebHttp to work out why this was needed and to try and find a solution. I see now the difficulty, WebHttp doesn't directly reflect over the methods, instead the attribute is instantiated and added to a collection of IOperationBehavior instances, which eventually gets populated to OperationDescription.OperationBehaviors. This happens in the code here:
https://github.com/CoreWCF/CoreWCF/blob/ee8eaaaee648ac7a897e39491e8b53317c905283/src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs#L253-L254
The method call to ` GetIOperationAttributesFromType<IOperationBehavior>` does all the work. I understand now why you tried to make this work the way you did. To find the correct solution we need to examine the existing pattern.  

The current compat code lives in `CustomAttributeProvider.GetDualCustomAttributes`. This uses a technique which we can reuse, but the implementation can't live in the same place as you'll see in a bit. The way this works is it calls `ICustomAttributeProvider.GetCustomAttributes` to get all attributes and then filters the set to the requested attribute type. If this list is empty, then it goes down the compat path looking for attributes which have a type name which matches the `System.ServiceModel` equivalent attribute. If there are any, then it calls an attribute type specific method to do the conversion. For example, we might call `ConvertFromServiceModelOperationContractAttribute` if there was an attribute with the type fullname of `System.ServiceModel.OperationContractAttribute`. This only works because we can instantiate the CoreWCF target attribute type. I.e. we can call `new OperationContractAttribute()` and then use reflection to grab the named property values we know exist and set them on the CoreWCF implementation.  
The first problem we hit is that the Web attributes are populated by looking for attributes which implement `IOperationBehavior`. If there are any attributes applied which implement this, then the compat path doesn't get called. This is fine for the current compat code as it's all for concrete attribute types. If you've attributed with both `CoreWCF.OperationContractAttribute` and `System.ServiceModel.OperationContractAttribute`, we don't want to deal with any ambiguities, so we treat the CoreWCF version as the source of truth and ignore the other. Taking the same approach for the Web attributes would cause problems as an unrelated `IOperationBehavior` attribute would obscure the S.SM.Web* attributes and if added later would cause the Web attributes to suddenly disappear.  
The next problem we have is that CoreWCF.Primitives doesn't know about CoreWCF.WebHttp so it can't new the relevant attributes and can't reference the properties directly. You could use reflection, but then this gets really messy. You have to search the list of loaded assemblies for the one you want (and this breaks with some linker scenarios) and then find the type. This is why the solution can't live in the same place as the existing compat code. This new compat code needs to live in CoreWCF.WebHttp which can have knowledge about the properties we need to copy and can new up the correct type.   

To summarize, we need to populate OperationDescription.OperationBehaviors with either WebGetAttribute or WebInvokeAttribute, and the code to do this needs to be in WebHttp. The easiest way to do this is with a behavior as behaviors get called as part of wiring everything up before the service starts. The OperationBehaviors need to be populated before any code runs which depends on that behavior existing. We also only want to invoke the code once. The ideal place to do this is in `WebHttpBehavior.ApplyDispatchBehavior`. The earliest place I could find that needs these populated is in the call to `GetOperationSelector`:
https://github.com/CoreWCF/CoreWCF/blob/453d91f0b0595c71747cd962b0c916660b0580dc/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs#L159
The fixup needs to happen before this line. To me the best seems to be just after the validation code which can throw, so after the code which ends on line 146.
https://github.com/CoreWCF/CoreWCF/blob/453d91f0b0595c71747cd962b0c916660b0580dc/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs#L138-L146

The code in `WebHttpBehavior.ApplyDispatchBehavior` would look something like this:
```c#
foreach (OperationDescription od in endpoint.Contract.Operations)
{
    WebGetAttribute wga = ((KeyedByTypeCollection<IOperationBehavior>)od.OperationBehaviors).Find<WebGetAttribute>();
    WebInvokeAttribute wia = ((KeyedByTypeCollection<IOperationBehavior>)od.OperationBehaviors).Find<WebInvokeAttribute>();
    if (wga==null && wia==null)
    {
        CheckForAndConvertSMAttributes(od);
    }
}
`
Then code similar to the following also needs to be added to `WebHttpBehavior`.
```c#
internal const string SMWWebGetAttributeFullName = "System.ServiceModel.Web.WebGetAttribute";
internal const string SMWWebInvokeAttributeFullName = "System.ServiceModel.Web.WebInvokeAttribute";

private static void CheckForAndConvertSMAttributes(OperationDescription od)
{
    var opMethod = SyncMethod ?? TaskMethod ?? BeginMethod;
    var attributes = opMethod.GetCustomAttributes();
    object smWga = attributes.Where(attribute => (attribute.GetType().FullName.Equals(SMWWebGetAttributeFullName)).SingleOrDefault();
    object smWia = attributes.Where(attribute => (attribute.GetType().FullName.Equals(SMWWebInvokeAttributeFullName)).SingleOrDefault();
    if (smWga != null)
    {
        od.OperationBehaviors.Add(ConvertSMWebGet(smWga));
    }
    if (smWia != null)
    {
        od.OperationBehaviors.Add(ConvertSMWebInvoke(smWia));
    }
}

private static WebGetAttribute ConvertSMWebGet(object smWga)
{
    // Converts the S.SM.Web.WebGetAttribute to CoreWCF.Web.WebGetAttribute
}

private static WebInvokeAttribute ConvertSMWebInvoke(object smWia)
{
    // Converts the S.SM.Web.WebInvokeAttribute to CoreWCF.Web.WebInvokeAttribute
}
```
There is a little bit of missing validation here, but I don't think it's something that needs to be worried about. Basically if you have a Sync method and a Task method defined with the same operation name (e.g. Foo and FooAsync), all the attributes need to be declared on the sync method. The same is true for Sync and APM as well as Task and APM. This code won't throw an exception if this constraint has been broken using the S.SM attributes. Instead, it will ignore anything on the lower-priority method. I don't see this as being an issue as it's intended to keep using the same code on .NET Framework which would fail if you do this.

arontsang:
@mconnew Updated PR using suggested changes.

Also switched to using dynamic compilation instead of reflection so that we can avoid using `CustomAttributeProvider.GetProperty`.

arontsang:
@mconnew Bump

arontsang:
@tibor19 @mconnew
Any chance of a review/merge?

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpBehavior.cs
```diff
@@ -263,6 +267,8 @@ public virtual void ApplyDispatchBehavior(ServiceEndpoint endpoint, EndpointDisp
             AddServerErrorHandlers(endpoint, endpointDispatcher);
         }
 
+
+
```
mconnew, line 271: Nit: you added some extra blank lines

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -0,0 +1,166 @@
...
+/// <summary>
+/// This class contains the logic to convert System.ServiceModel.Web attributes into
+/// CoreWCF.WebHttp attributes.
+/// This allows endpoints to be defined using WCF classic attributes as well as CoreWCF
+/// attributes, for a simpler migration path.
+/// </summary>
+internal static class WebHttpServiceModelCompat
+{
+
+    public static void ServiceModelAttributeFixup(ServiceEndpoint endpoint)
```
mconnew, line 25: Nit, there's an extra blank line at the start of the class

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -0,0 +1,166 @@
...
+                { "System.ServiceModel.Web.WebGetAttribute", new AttributeConverter<WebGetAttribute>() },
+                { "System.ServiceModel.Web.WebInvokeAttribute", new AttributeConverter<WebInvokeAttribute>() }
+            };
+
+        public static IEnumerable<IOperationBehavior> Convert(IEnumerable<Attribute> attributes)
+        {
+            foreach (Attribute attribute in attributes)
+            {
+                if (Convert(attribute) is { } convertedAttribute)
+                    yield return convertedAttribute;
```
mconnew, line 64: I've never seen this syntax before. What does `is { }` do?

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private class AttributeConverter<TOut> : IAttributeConverter where TOut : class, IOperationBehavior
+    {
+        private static readonly ConcurrentDictionary<Type, Func<Attribute, TOut?>> s_converterCache = new();
+
+        private static readonly MethodInfo s_buildDynamicMethodInfo = typeof(AttributeConverter<TOut>)
+            .GetMethod(nameof(BuildDynamic), BindingFlags.Static | BindingFlags.NonPublic)!;
+
+
```
mconnew, line 91: Nit: extra blank line

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -0,0 +1,166 @@
...
+        public IOperationBehavior? Convert(Attribute attribute)
+        {
+            var type = attribute.GetType();
+
+            if (!s_converterCache.TryGetValue(type, out var converter))
+            {
+                converter = s_converterCache[type] = Build(type);
+            }
+
+
```
mconnew, line 101: Nit: extra blank line

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <typeparam name="TOut">The converted attribute type</typeparam>
+        private static Func<Attribute, TOut?> Build(Type inputType)
+        {
+            return (Func<Attribute, TOut?>) s_buildDynamicMethodInfo
+                .MakeGenericMethod(inputType)
+                .Invoke(null, Array.Empty<object>());
+        }
+
+        private static Func<Attribute, TOut?> BuildDynamic<TInput>()
+        {
```
mconnew, line 124: Just want to say this is some very nice clean code.

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
arontsang: That is the new(ish) C#9 conditional pattern matching.

The full form is this.

`if (rhs is SomeType { patternMatch } lhs) { }`

We can omit the `SomeType` because we are casting from `IOperationBehavior?` to `IOperationBehavior`. 

The pattern we are matching is empty (ie, match on any).

Overall, this is just a nice way to do a null check with the new nullable types, whilst putting the value into a variable that is explicitly not null.

arontsang:
@mconnew removed the extra whitespace

arontsang:
@mconnew Any chance we can get this merged?

mconnew:
I plan to release a new build in the next day or two which will include this.

arontsang:
Thanks

PR #785, opened 2022-08-06
Adding support generic queue transport

Ximik87:
Adding support of transports for message brokers. Issue #164
Support for NetMsmqBinding and RabbitMqBinding
Examples of integrations indicated [WcfMsmqIntegration](https://github.com/Ximik87/WcfMsmqIntegration) and [WcfRabbitMqIntegration](https://github.com/Ximik87/WcfRabbitMqIntegration) were used
I conducted load testing (10 hours under the specified number of requests)
NetMsmqBinding - 2000 RPS
RabbitMqBinding - 7000 RPS

Integration tests are also excluded, since the necessary CI infrastructure is missing.
So I will be grateful if you indicate the case that needs to be tested.


Ximik87:
@mconnew Hi, the test NetTCPRequestReplyWithTransportMessageEchoStringDemuxFailure  fail in the PR check, although I did not change this code. I checked the test locally and it passed. Can we do something about it?

birojnayak:
> @mconnew Hi, the test NetTCPRequestReplyWithTransportMessageEchoStringDemuxFailure fail in the PR check, although I did not change this code. I checked the test locally and it passed. Can we do something about it?

@Ximik87 thank you for posting this CR. I have worked on your codebase with @mconnew we have made some changes to Common component to make it more extensible multiple queuing mechanism. Please clone this branch https://github.com/birojnayak/CoreWCF/tree/DmitryModifiedCommit and apply changes on top of your changes. Few things I have left was a)Commented some file, which you can remove (left so that you can see why we made those changes) b)I have ensured the integration tests work, if you want to add other tests, feel free to do that. c

birojnayak (changes requested):

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqInputChannel.cs
```diff
@@ -0,0 +1,45 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace CoreWCF.Channels
+{
+    internal class MsmqInputChannel : ServiceChannelBase, IInputChannel
```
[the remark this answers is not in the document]
birojnayak: checked with @mconnew , we should get rid of this class and it's dependency

birojnayak:
@Ximik87 are you working on the feedback or awaiting on something which I can help with ..

Ximik87:
@birojnayak I will do changes from feedback, but now I can't make them, because I relocate to cyprus. I plan to do it by the end of october

src/CoreWCF.MSMQ/src/CoreWCF.MSMQ.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>CoreWCF.MSMQ</AssemblyName>
+    <PackageId>CoreWCF.MSMQ</PackageId>
+    <RootNamespace />
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <RootNamespace />
+  </PropertyGroup>
```
mconnew, line 10: RootNamespace appears twice

src/CoreWCF.MSMQ/src/CoreWCF/Channels/DeadLetterQueueSender.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace CoreWCF.Channels
+{
+    internal class DeadLetterQueueSender : IDeadLetterQueueMsmqSender
+    {
+        public async Task Send(PipeReader message, Uri endpoint)
+        {
+            string nativeQueueName = MsmqQueueNameConverter.GetMsmqFormatQueueName(endpoint);
+            MemoryStream memStream = await ConvertToStream(message);
+            var queue = new MessageQueue(nativeQueueName);
+            var messageForQueue = new MSMQ.Messaging.Message { BodyStream = memStream };
```
mconnew, line 20: Do we validate the queue exists before we get to executing this code?

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+{ 
+    public abstract class MsmqBindingBase : Binding
+    {
+        internal MsmqBindingElementBase _transport;
+
+        protected MsmqBindingBase()
+        { }
+
+        // [DefaultValue(typeof(System.TimeSpan), MsmqDefaults.ValidityDurationString)]
+        //public TimeSpan ValidityDuration
```
mconnew, line 17: Can you add a comment that this will need to be enabled when we add System.Transactions support?

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        [DefaultValue(MsmqDefaults.ExactlyOnce)]
+        public bool ExactlyOnce
+        {
+            get { return _transport.ExactlyOnce; }
+            set { _transport.ExactlyOnce = value; }
+        }
+
+        //[DefaultValue(TransportDefaults.MaxReceivedMessageSize)]
+        public long MaxReceivedMessageSize
```
mconnew, line 52: Can you add the relevant constant so this attribute can be uncommented? This is used when working with Configuration.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        [DefaultValue(MsmqDefaults.ReceiveContextEnabled)]
+        public bool ReceiveContextEnabled
+        {
+            get { return _transport.ReceiveContextEnabled; }
+            set { _transport.ReceiveContextEnabled = value; }
+        }
+
+        //[DefaultValue(MsmqDefaults.ReceiveErrorHandling)]
+        //public ReceiveErrorHandling ReceiveErrorHandling
```
mconnew, line 80: I see this is related to System.Transactions support. Can you add a similar comment here.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+        //{
+        //    get { return transport.ReceiveErrorHandling; }
+        //    set { transport.ReceiveErrorHandling = value; }
+        //}
+
+        [DefaultValue(typeof(TimeSpan), MsmqDefaults.RetryCycleDelayString)]
+        public TimeSpan RetryCycleDelay
```
mconnew, line 87: The only usages I could find of this property is with System.Transactions support In the Msmq4PoisonHandler on .NET Framework. I think this can be left out for now unless it's used somewhere I'm not aware of.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            get { return _transport.RetryCycleDelay; }
+            set { _transport.RetryCycleDelay = value; }
+        }
+
+        public override string Scheme { get { return _transport.Scheme; } }
+
+        [DefaultValue(typeof(TimeSpan), MsmqDefaults.TimeToLiveString)]
+        public TimeSpan TimeToLive
+        {
```
mconnew, line 97: This is a client only api so remove.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+            get { return _transport.TimeToLive; }
+            set { _transport.TimeToLive = value; }
+        }
+
+        [DefaultValue(MsmqDefaults.UseSourceJournal)]
+        public bool UseSourceJournal
+        {
```
mconnew, line 104: I could only find client usage of this property

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingBase.cs
```diff
@@ -0,0 +1,116 @@
...
+            get { return _transport.UseSourceJournal; }
+            set { _transport.UseSourceJournal = value; }
+        }
+
+        [DefaultValue(MsmqDefaults.UseMsmqTracing)]
+        public bool UseMsmqTracing
```
mconnew, line 110: I can only find usage of this on the client side.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public DeadLetterQueue DeadLetterQueue
+        {
+            get { return _deadLetterQueue; }
+            set
+            {
+                if (!DeadLetterQueueHelper.IsDefined(value))
+                {
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("value"));
+                }
```
mconnew, line 30: Can you replace string quoted parameter names with nameof. So here it would be `nameof(value)`. There's another one on line 52. These don't get discovered easily by the csharp source analyzers. Here is a useful list of find/replace pairs that I use in VS for doing these more easily. You need to turn on the use regex option on the find, but this makes life a lot easier.
```
ThrowHelperError\(new ArgumentNullException\(\"([a-zA-Z0-9]+)\"\)
ThrowHelperError(new ArgumentNullException(nameof($1))

ThrowHelperError\(new ArgumentOutOfRangeException\(\"([a-zA-Z0-9]+)\"\)
ThrowHelperError(new ArgumentOutOfRangeException(nameof($1))

ThrowHelperError\(new ArgumentOutOfRangeException\(\"([a-zA-Z0-9]+)\",
ThrowHelperError(new ArgumentOutOfRangeException(nameof($1),

new ArgumentOutOfRangeException\(\"([a-zA-Z0-9]+)\",
new ArgumentOutOfRangeException(nameof($1),

ThrowHelperArgumentNull\(\"([a-zA-Z0-9]+)\"\)
ThrowHelperArgumentNull(nameof($1))
```

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                _deadLetterQueue = value;
+            }
+        }
+
+        public bool Durable { get; set; }
+
```
mconnew, line 37: Only used on client

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+        public bool TransactedReceiveEnabled => ExactlyOnce;
+
```
mconnew, line 39: This is an implementation of ITransactedReceiveBindingElement which we don't have yet so you can remove this property.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public bool ExactlyOnce { get; set; }
+
+
+        public int ReceiveRetryCount
+        {
```
mconnew, line 45: Only applicable when using transactions

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
+                        new ArgumentOutOfRangeException("value", value, "MsmqNonNegativeArgumentExpected"));
+                }
+
+                _receiveRetryCount = value;
+            }
+        }
+
+        public int MaxRetryCycles
+        {
```
mconnew, line 60: Only applicable when using transactions

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+
+        public bool Durable { get; set; }
+
+        public bool TransactedReceiveEnabled => ExactlyOnce;
+
+
+        public bool ExactlyOnce { get; set; }
+
```
mconnew, line 42: Nit, extra blank lines around property

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                        new ArgumentOutOfRangeException("value", value, "MsmqNonNegativeArgumentExpected"));
+                }
+
+                _maxRetryCycles = value;
+            }
+        }
+
+
+        public MsmqTransportSecurity MsmqTransportSecurity { get; internal set; }
+
```
mconnew, line 76: Nit, extra blank lines

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException());
+                }
+
+                _retryCycleDelay = value;
+            }
+        }
+
+        public TimeSpan TimeToLive
+        {
```
mconnew, line 120: Client only

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqBindingElementBase.cs
```diff
@@ -0,0 +1,199 @@
...
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException());
+                }
+
+                _timeToLive = value;
+            }
+        }
+
+
+        public bool UseMsmqTracing { get; set; }
+
```
mconnew, line 140: Client only

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqInputChannel.cs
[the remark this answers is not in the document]
mconnew: Ping?

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqNetcoreTransport.cs
```diff
@@ -0,0 +1,114 @@
...
+using CoreWCF.Queue.Common;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using MSMQ.Messaging;
+using MSMQM = MSMQ.Messaging;
+
+namespace CoreWCF.Channels
+{
+    public class MsmqNetcoreTransport : IQueueTransport, IDisposable
+    {
```
mconnew, line 19: No need to have Netcore in the name. Maybe MsmqQueueTransport, basically strip the I and prefix the purpose.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqNetcoreTransport.cs
```diff
@@ -0,0 +1,114 @@
...
+        private readonly TimeSpan _queueReceiveTimeOut;
+        private readonly IDeadLetterQueueMsmqSender _deadLetterQueueSender;
+        private readonly ILogger<MsmqNetcoreTransport> _logger;
+
+        public MsmqNetcoreTransport(QueueOptions options, IServiceDispatcher serviceDispatcher, IServiceProvider serviceProvider)
+        {
+            _deadLetterQueueSender = serviceProvider.GetRequiredService<IDeadLetterQueueMsmqSender>();
+            _baseAddress = serviceDispatcher.BaseAddress;
+            string nativeQueueName = MsmqQueueNameConverter.GetMsmqFormatQueueName(options.QueueName);
+            _messageQueue = new MessageQueue(nativeQueueName);
```
mconnew, line 31: The queue name shouldn't come from QueueOptions, it should be derived from the URL (which is in BaseAddress). There's only a single QueueOptions as well, which means you could only pull from a single queue which might not be the case.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqPoisonMessageException.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace CoreWCF.Channels
+{
+    public class MsmqPoisonMessageException : Exception
+    {
```
mconnew, line 9: This should be derivied from PoisonMessageException which derives from CommunicationException. All CoreWCF types ultimately derive from CommunicationException as it makes exception handling easier.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqQueueNameConverter.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace CoreWCF.Channels
+{
+    public class MsmqQueueNameConverter
+    {
```
mconnew, line 9: This should be internal as it's a helper class. Also there's some formatting issues that need tidying up.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqTransportBindingElement.cs
```diff
@@ -0,0 +1,83 @@
...
+        private MsmqTransportBindingElement(MsmqTransportBindingElement elementToBeCloned)
+            : base(elementToBeCloned)
+        {
+            _useActiveDirectory = elementToBeCloned._useActiveDirectory;
+            _maxPoolSize = elementToBeCloned._maxPoolSize;
+        }
+
+
+        public int MaxPoolSize
+        {
```
mconnew, line 27: It doesn't look like you use this property anywhere. This value is used to configure a pool of reusable objects. It's not needed with the current implementation.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqTransportBindingElement.cs
```diff
@@ -0,0 +1,83 @@
...
+        public override string Scheme
+        {
+            get
+            {
+                return "net.msmq";
+            }
+        }
+
+        public bool UseActiveDirectory
+        {
```
mconnew, line 52: I'm not finding this used anywhere.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqTransportSecurity.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [DefaultValue(MsmqDefaults.MsmqAuthenticationMode1)]
+        public MsmqAuthenticationMode MsmqAuthenticationMode
+        {
+            get { return _msmqAuthenticationMode; }
+            set
+            {
+                //if (!MsmqAuthenticationModeHelper.IsDefined(value))
+                //    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("value"));
```
mconnew, line 50: The value does need to be validated as you can cast any integer to an enum type.

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqTransportSecurity.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [DefaultValue(MsmqDefaults.DefaultMsmqSecureHashAlgorithm)]
+        public MsmqSecureHashAlgorithm MsmqSecureHashAlgorithm
+        {
+            get { return _msmqHashAlgorithm; }
+            set
+            {
+                //if (!MsmqSecureHashAlgorithmHelper.IsDefined(value))
+                //    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("value"));
```
mconnew, line 74: This needs to be validated

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqTransportSecurity.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [DefaultValue(MsmqDefaults.MsmqProtectionLevel)]
+        public ProtectionLevel MsmqProtectionLevel
+        {
+            get { return _msmqProtectionLevel; }
+            set
+            {
+                //if (!ProtectionLevelHelper.IsDefined(value))
+                //    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("value"));
```
mconnew, line 86: Validation needs to be enabled

src/CoreWCF.MSMQ/tests/Helpers/DependencyResolverHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private readonly IWebHost _webHost;
+
+        public DependencyResolverHelper(IWebHost webHost)
+        {
+            _webHost = webHost;
+        }
+
+        public T GetService<T>()
+        {
```
mconnew, line 20: Potentially fragile. I don't know if a scoped ServiceProvider has a finalizer registered, but if it does, it could dispose anything it created.

src/CoreWCF.MSMQ/tests/Helpers/MessageContainer.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.IO;
+
+namespace CoreWCF.MSMQ.Tests.Helpers
+{
+    internal class MessageContainer
+    {
+        public static Stream GetTestMessage()
+        {
+             string currentDirectory = AppDomain.CurrentDomain.BaseDirectory ?? throw new Exception("AppDomain current directroy is empty");
+              string path = Path.Combine(currentDirectory, "Resources/msmqTestMessage.bin");
```
mconnew, line 14: Please fix indentation. Also typo in message, directroy -> directory

src/CoreWCF.Queue/src/CoreWCF.Queue.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>CoreWCF.Queue</AssemblyName>
+    <PackageId>CoreWCF.Queue</PackageId>
+    <RootNamespace />
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <RootNamespace />
+  </PropertyGroup>
```
mconnew, line 10: RootNamespace is added twice

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueFetchMessage.cs
```diff
@@ -0,0 +1,34 @@
...
+        public async Task InvokeAsync(QueueMessageContext queueMessageContext)
+        {
+            var pipeReader = queueMessageContext.QueueMessageReader;
+            var readResult = await pipeReader.ReadAsync();
+            var memStream = new MemoryStream(readResult.Buffer.ToArray());
+            var encoder = queueMessageContext.QueueTransportContext.MessageEncoderFactory.Encoder;
+            var maxReceivedMessageSize = queueMessageContext.QueueTransportContext.QueueBindingElement.MaxReceivedMessageSize;
+            var message = await encoder.ReadMessageAsync(memStream, (int)maxReceivedMessageSize); 
+            message.Headers.To = queueMessageContext.LocalAddress.Uri;
+            queueMessageContext.SetRequestMessage(message);
```
mconnew, line 29: Should only set To is it's currently null.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueHandshakeMiddleware.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using CoreWCF.Queue.Common.Configuration;
+
+namespace CoreWCF.Queue.Common
+{
+    public class QueueHandShakeMiddleWare
+    {
```
mconnew, line 9: I don't like the use of the word HandShake here. This comes from NetTcp, but that has an actual connection handshake.Maybe just call is QueueMiddleware. I think this can be internal for now.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+            get { return _concurrencyLevel; }
+            set
+            {
+                if (value < 1) throw new ArgumentOutOfRangeException(nameof(ConcurrencyLevel), SR.ValueMustBeGreaterThanZero);
+                _concurrencyLevel = value;
+
+            }
+        }
+        public string QueueName { get; set; }
+    }
```
mconnew, line 22: We don't need the QueueName here, it should be derived from the endpoint url

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueuePollingService.cs
```diff
@@ -0,0 +1,120 @@
...
+            Task.Run(() => WaitAndStart(cancellationToken), cancellationToken);
+            return Task.CompletedTask;
+        }
+
+        private async Task WaitAndStart(CancellationToken cancellationToken)
+        {
+            while (!_isServiceBuilderOpened)
+            {
+                await Task.Delay(100, cancellationToken);
+            }
```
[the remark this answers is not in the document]
mconnew: Yes, create a TaskCompletionSource in the constructor and have the Opened event handler complete it. In this method first check the state of the service builder and if not Opened, await the TaskCompletionSource.Task. You don't want wait unconditionally as there is a potential race condition when the service builder is being opened where it might get past the point of calling the event handler before you set it. So only await the Task if it's not opened. If you want to honor the cancellationToken, register a cancellation callback with the token which cancels the TaskCompletionSource (don't forget to dispose the registration).

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueProcessMessage.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Threading.Tasks;
+using CoreWCF.Queue.Common.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace CoreWCF.Queue.Common
+{
+    public class QueueProcessMessage
+    {
```
mconnew, line 12: Make this internal. Anything we make public has it's api locked down which doesn't allow for evolving it as easily if we need to make changes.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueTransportContext.cs
```diff
@@ -0,0 +1,18 @@
...
+using CoreWCF.Channels;
+using CoreWCF.Configuration;
+using CoreWCF.Queue.Common.Configuration;
+
+namespace CoreWCF.Queue.Common
+{
+    public class QueueTransportContext
+    {
+        public IServiceDispatcher ServiceDispatcher { get;  set; }
+        public MessageEncoderFactory MessageEncoderFactory { get; set; }
```
mconnew, line 13: Nit: there's an extra space between the get and set

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/DefaultQueueTransportPump.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CoreWCF.Queue.Common
+{
+    /// <summary>
+    /// DefaultQueueTransportPump model serves the purpose of pull model
+    /// </summary>
+    public class DefaultQueueTransportPump : QueueTransportPump
+    {
```
mconnew, line 15: I think this can be internal as we have a factory method on QueueTransportPump to create an instance.

src/CoreWCF.RabbitMQ/src/CoreWCF.RabbitMQ.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>CoreWCF.RabbitMQ</AssemblyName>
+    <PackageId>CoreWCF.RabbitMQ</PackageId>
+    <RootNamespace />
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+    <RootNamespace />
+  </PropertyGroup>
```
mconnew, line 10: Duplicate property

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/CurrentVersion.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Channels
+{
+    /// <summary>
+    /// Properties of the current RabbitMQ Service Model Version
+    /// </summary>
+    public static class CurrentVersion
+    {
```
mconnew, line 10: Make class internal ass it only has an internal member

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        { }
+
+        /// <summary>
+        /// Uses the broker specified by the given hostname and port with
+        /// Protocols.DefaultProtocol.
+        /// </summary>
+        /// <param name="hostname">The hostname of the broker to connect to</param>
+        /// <param name="port">The port of the broker to connect to</param>
+        public RabbitMqBinding(string hostname, int port)
+            : this(hostname, port, Protocols.DefaultProtocol)
```
mconnew, line 30: Hostname and port should be encoded in the endpoint url and not set on the binding.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// </summary>
+        /// <param name="hostname">The hostname of the broker to connect to</param>
+        /// <param name="port">The port of the broker to connect to</param>
+        /// <param name="username">The broker username to connect with</param>
+        /// <param name="password">The broker password to connect with</param>
+        /// <param name="virtualHost">The broker virtual host</param>
+        /// <param name="maxMessageSize">The largest allowable encoded message size</param>
+        /// <param name="protocol">The protocol version to use</param>
+        public RabbitMqBinding(string hostname, int port,
+                               string username, string password, string virtualHost,
```
mconnew, line 57: It's not typical to set properties via binding constructors parameters except security mode.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Uses the specified protocol. The broker must be set before use.
+        /// </summary>
+        /// <param name="protocol">The protocol version to use</param>
+        public RabbitMqBinding(IProtocol protocol)
+        {
+            BrokerProtocol = protocol;
+            Name = "RabbitMQBinding";
+            Namespace = "http://schemas.rabbitmq.com/2007/RabbitMQ/";
+
```
mconnew, line 79: Where did this namespace come from? We need to be careful to not apply meaning to namespaces we don't own.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+            Initialize();
+        }
+
+        public override BindingElementCollection CreateBindingElements()
+        {
+            //Transport.HostName = HostName;
+            //Transport.Port = Port;
+            Transport.BrokerProtocol = BrokerProtocol;
+            if (MaxMessageSize != DefaultMaxMessageSize)
+            {
```
mconnew, line 89: It's better to have the MaxMessageSize property be a passthrough to setting Transport.MaxReceivedMessageSize rather than needing to synchronize them.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+            BindingElementCollection elements = new BindingElementCollection { _encoding, Transport, };
+
+            return elements;
+        }
+
+        private void Initialize()
+        {
+            lock (this)
+            {
```
mconnew, line 100: A lock shouldn't be necessary as this is only called from a constructor. Also it's a bad idea to lock on `this` or anything that external code can have a reference to as you can cause unexpected problems if outside code takes a lock on this instance. It's only safe in this case because this is called from a constructor so the instantiating code doesn't have a reference yet, which then invalidates the need for the lock.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// Gets the AMQP transport binding element
+        /// </summary>
+        public RabbitMqTransportBindingElement Transport { get; private set; }
+
+        /// <summary>
+        /// Determines whether or not the TransactionFlowBindingElement will
+        /// be added to the channel stack
+        /// </summary>
+        public bool TransactionFlow { get; set; }
+
```
mconnew, line 149: Remove or comment for now.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Specifies whether or not the CompositeDuplex and ReliableSession
+        /// binding elements are added to the channel stack.
+        /// </summary>
+        public bool OneWayOnly { get; set; }
+    }
```
mconnew, line 155: We don't have either CompositeDuplex or ReliableSession yet so remove or comment.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,156 @@
...
+        public override string Scheme
+        {
+            get { return CurrentVersion.Scheme; }
+        }
+
+        /// <summary>
+        /// Specifies the hostname of the RabbitMQ Server
+        /// </summary>
+        public string HostName { get; set; }
+
```
mconnew, line 123: As mentioned earlier, HostName and Port shouldn't exist on the binding.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// The largest receivable encoded message
+        /// </summary>
+        public override long MaxReceivedMessageSize { get; set; }
+
+        /// <summary>
+        /// The username  to use when authenticating with the broker
+        /// </summary>
+        internal string UsernameConfigKey { get; set; }
+
```
mconnew, line 78: I think an ICredentials would be a better property type to use. NetworkCredential implements this interface so at its simplest a NetworkCredential can be passed, but if you need something more flexible, a custom ICredentials can be provided to dynamically lookup the credentials to use, eg from some kind of Key Vault.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitmqTransportPump.cs
```diff
@@ -0,0 +1,78 @@
...
+            _channel.Dispose();
+            _connection.Dispose();
+            return Task.CompletedTask;
+        }
+
+        private async void ConsumeMessage(BasicDeliverEventArgs eventArgs, QueueTransportContext queueTransportContext)
+        {
+            _logger.LogInformation("Receiving message from rabbitmq");
+            var reader = PipeReader.Create(new MemoryStream(eventArgs.Body.ToArray()));
+
```
mconnew, line 62: There's a lot of allocation and copying here. eventArgs.Body is a `ReadOnlyMemory<byte>` so calling ToArray allocates a new array and copies. Then MemoryStream is another allocation. There's a newer overload of PipeReader.Create which takes a `ReadOnlySequence<byte>` instead which avoids all this allocation. So can replace the code with tihs:
```c#
var reader = PipeReader.Create(new ReadOnlySequence<byte>(eventArgs.Body));
```
You might need to reference the 6.0 version of the nuget package System.IO.Pipelines, but that targets netstandard2.0 so should work before we require .net 6.0.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueMessageContext.cs
```diff
@@ -0,0 +1,93 @@
...
+using CoreWCF.Channels;
+using CoreWCF.Runtime;
+
+namespace CoreWCF.Queue.Common
+{
+    public class QueueMessageContext : RequestContext
+    {
+        public PipeReader QueueMessageReader { get; set; }
+        public IDictionary<string, object> Properties { get; set; }
+        private Message _requestMessage;
```
mconnew, line 18: I'm a little concerned about extra unnecessary allocation here. I think we should make this virtual, get only, and default to lazily creating a Dictionary here instead of the code which constructs an instance.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueTransportContext.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using CoreWCF.Channels;
+using CoreWCF.Configuration;
+using CoreWCF.Queue.Common.Configuration;
+
+namespace CoreWCF.Queue.Common
+{
+    public class QueueTransportContext
+    {
```
mconnew, line 11: All these properties should be internal set.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueTransportContext.cs
```diff
@@ -0,0 +1,18 @@
...
+        public IServiceDispatcher ServiceDispatcher { get;  set; }
+        public MessageEncoderFactory MessageEncoderFactory { get; set; }
+        public QueueBaseTransportBindingElement QueueBindingElement { get; set; }
+        public QueueMessageDispatcherDelegate QueueHandShakeDelegate { get; set; }
+        public QueueTransportPump QueuePump { get; internal set; }
```
mconnew, line 16: Rename property to QueueMessageDispatcher

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/Configuration/QueueMiddlewareBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+            return app;
+        }
+
+        private T GetProperty<T>(string key)
+        {
+            return Properties.TryGetValue(key, out object value) ? (T)value : default;
+        }
+
+        private void SetProperty<T>(string key, T value)
+        {
```
mconnew, line 69: No need to have generic on method as Properties value type is object

src/CoreWCF.RabbitMQ/tests/Fakes/Interceptor.cs
```diff
@@ -0,0 +1,43 @@
...
+        public string Name { get; private set; }
+
+        public void SetName(string name)
+        {
+            Name = name;
+        }
+    }
+    /*
+    internal class TestConnectionHandler : IQueueConnectionHandler
+    {
```
mconnew, line 27: Remove commented out code

src/CoreWCF.RabbitMQ/tests/Helpers/MessageContainer.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.IO;
+
+namespace CoreWCF.RabbitMQ.Tests.Helpers
+{
+    internal class MessageContainer
+    {
+        public static Stream GetTestMessage()
+        {
+            string currentDirectory = AppDomain.CurrentDomain.BaseDirectory ?? throw new Exception("AppDomain current directroy is empty");
+            string path = Path.Combine(currentDirectory, "Resources/rabbitmqTestMessage.bin");
```
mconnew, line 14: Does Environment.CurrentDirectory do the same thing? AppDomain is only there for code compat reasons and a few methods which need to be available somewhere (like type loaders etc) so I like to avoid it's usage.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/Configuration/QueueMiddlewareBuilderExtension.cs
```diff
@@ -0,0 +1,185 @@
...
+                if (invokeMethods.Length > 1)
+                {
+                    // TODO: String resources
+                    throw new InvalidOperationException(
+                        $"Resources.FormatException_UseMiddleMultipleInvokes({InvokeAsyncMethodName}");
+                }
+
+                if (invokeMethods.Length == 0)
+                {
+                    // TODO: String resources
```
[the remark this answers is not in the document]
Ximik87: I don't know, it copy paste from NetTcp project, I want remove it, ok?

src/CoreWCF.MSMQ/src/CoreWCF/Channels/DeadLetterQueueSender.cs
Ximik87: Yes, sure

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueTransportContext.cs
Ximik87: Property ServiceDispatcher required [here](https://github.com/Ximik87/CoreWCF/blob/main/src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqQueueTransport.cs#L97), can we do it differently?

src/CoreWCF.Queue/tests/DefaultQueueTransportPumpTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using CoreWCF.Queue.Common;
+using CoreWCF.Queue.Tests.Fakes;
+using Xunit;
+
+namespace CoreWCF.Queue.Tests
+{
+    public class DefaultQueueTransportPumpTests
+    {
+        [Fact]
+        public async Task DefaultQueueTransportPump_WhenStated_Success()
```
birojnayak, line 14: thanks for these tests

src/CoreWCF.Queue/src/CoreWCF/Runtime/Diagnostics/EtwDiagnosticTrace.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+            set
+            {
+                s_defaultEtwProviderId = value;
+            }
+        }
+
+        public EtwDiagnosticTrace(string traceSourceName, Guid etwProviderId)
+        //: base(traceSourceName)
+        {
```
birojnayak, line 27: clean up if not needed

src/CoreWCF.MSMQ/src/CoreWCF/Channels/NetMsmqBinding.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        internal int MaxPoolSize
+        {
+            get { return (_transport as MsmqTransportBindingElement).MaxPoolSize; }
+            set { (_transport as MsmqTransportBindingElement).MaxPoolSize = value; }
+        }
+
+        [DefaultValue(MsmqDefaults.UseActiveDirectory)]
+        public bool UseActiveDirectory
```
birojnayak, line 77: does this work ? if not should we throw not implemented exception ?

src/CoreWCF.MSMQ/tests/IntegrationTests.cs
```diff
@@ -59,21 +59,6 @@ public async Task ReceiveMessage_ServiceCall_Fail()
                 Assert.True(result);
             }
         }
-
```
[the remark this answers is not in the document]
Ximik87: Yes, sorry, I rollback this change

src/CoreWCF.MSMQ/src/CoreWCF/Channels/NetMsmqBinding.cs
Ximik87: Yes, I added throw exception in MsmqTransportBindingElement

src/CoreWCF.MSMQ/src/CoreWCF/Channels/MsmqQueueNameConverter.cs
Ximik87: Ok, I will fixed, but how to test this code for example? I have tried applying the attribute internalVisibleTo, but not work, because need assembly strong signing

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueTransportContext.cs
Ximik87: thanks, I misunderstood

birojnayak:
@Ximik87 could you please take a look why tests are failing ? I can take a final look before approving the PR...

birojnayak:
looks good to me.. @mconnew can you take a final  look ?

mconnew:
Merged into feature/queue branch and will merge to main when ready to release.

PR #796, opened 2022-08-11
Add corewcf perf testing

JiakaiDou:
1. Pull code from https://github.com/CoreWCF/CoreWCF/pull/273
2. Update projects to support net6.0 and update the version of referenced packages
3. Update the jobs and profiles in the CoreWCF.Perf.yml file
@HongGit @mconnew


src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,168 @@
...
+                var stopwatchFirstReq = new Stopwatch();
+                stopwatchFirstReq.Start();
+                var result = client.HelloAsync("hello world").Result;
+
+                BenchmarksEventSource.Measure("firstrequest", stopwatchFirstReq.ElapsedMilliseconds);
+
+                while (DateTime.Now <= startTime.Add(measurementDurationPerTime))
+                {
+                    var rtnResult = client.HelloAsync("hello world").Result;
+                    Console.WriteLine(rtnResult);
```
mconnew, line 75: Do the tests need to be in a synchronous method? .NET now supports an async Main method.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,168 @@
...
+                    request++;
+                    requestTime = request;
+                    if (requestTime >= test._paramRequests & test._paramCloseAndOpen)
+                    {
+                        try
+                        {
+                            ((IClientChannel)client).Close();
+
```
mconnew, line 83: If you switch to async, you will replace this with `await Task.Factory.FromAsync(((IClientChannel)client).BeginClose, ((IClientChannel)client).EndClose, null);` and the equivalent for the Open call below.

src/CoreWCF.CrankTests/CoreWCFPerfService/CoreWCF.PerfService.csproj
```diff
@@ -0,0 +1,16 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <PackageReference Include="CoreWCF.Http" Version="1.1.0" />
```
mconnew, line 11: This is only needed for net472. For all other tfm's, you need to use `<FrameworkReference Include="Microsoft.AspNetCore.App" />`. Also this should be 2.1.x. I think the current version is 2.1.7. AspNetCore 2.2 is not in support, but 2.1 is still in support on net472 only. See [https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj](https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj) for reference.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
JiakaiDou: Already changed to an async Main method.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
JiakaiDou: Have updated

src/CoreWCF.CrankTests/CoreWCFPerfService/CoreWCF.PerfService.csproj
JiakaiDou: Have update

src/CoreWCF.CrankTests/CoreWCFPerfClient/CoreWCF.PerfClient.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Crank.EventSources" Version="0.2.0-alpha.22228.2" />
+    <PackageReference Include="System.Private.ServiceModel" Version="4.9.0" />
+    <PackageReference Include="System.ServiceModel.Http" Version="4.9.0" />
```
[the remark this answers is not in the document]
JiakaiDou: Removed the reference to System.Private.ServiceModel

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                stopwatchFirstReq.Start();
+                var result = client.HelloAsync("hello world").Result;
+
+                BenchmarksEventSource.Measure("firstrequest", stopwatchFirstReq.ElapsedMilliseconds);
+
+                while (DateTime.Now <= startTime.Add(measurementDurationPerTime))
+                {
+                    var rtnResult = await client.HelloAsync("hello world");
+                    Console.WriteLine(rtnResult);
+                    request++;
```
mconnew, line 77: Console.WriteLine is a relatively expensive operation and will affect the results.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                    requestTime = request;
+                    if (requestTime >= test._paramRequests & test._paramCloseAndOpen)
+                    {
```
mconnew, line 80: Once request reaches test._paramRequests, it will create a new client every time as you are just copying request to requestTime. The resetting of requestTime to 0 at the end of the if block suggests you meant to simply increment requestTime instead of copying request across.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                        catch (Exception ex)
+                        {
+                            Console.WriteLine($"Client Close failed : {ex}.");
+                            ((IClientChannel)client).Abort();
+                        }
+                        requestTime = 0;
+                    }
+
+                    if (((IClientChannel)client).State != CommunicationState.Opened)
+                    {
```
mconnew, line 95: This code should probably live inside the if block just above. Otherwise you are doing a cast a virtual fetch of the State property on every loop

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+        private bool ProcessRunOptions(string[] args)
+        {
+            foreach (string s in args)
+            {
+                Console.WriteLine(s);
+                string[] p = s.Split(new char[] { ':' }, count: 2);
+                if (p.Length != 2)
+                {
+                    continue;
+                }
```
mconnew, line 117: An argument which can't be parsed was passed on the command line. I don't think we should silently ignore it. I think passing an invalid command line argument should result in failing to run the test. Otherwise you risk a typo resulting in testing something different than intended.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                BenchmarksEventSource.Measure("corewcfperf/rps/max", request / test._paramPerfMeasurementDuration.TotalSeconds);
+            }
+        }
+
+        private bool ProcessRunOptions(string[] args)
+        {
+            foreach (string s in args)
+            {
+                Console.WriteLine(s);
+                string[] p = s.Split(new char[] { ':' }, count: 2);
```
mconnew, line 113: I think this was left in by mistake.

src/CoreWCF.CrankTests/CoreWCFPerfClient/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                    case Parameters.CloseAndOpen:
+                        if (!bool.TryParse(p[1], out _paramCloseAndOpen))
+                        {
+                            return ReportWrongArgument(s);
+                        }
+                        break;
+
+                    default:
+                        Console.WriteLine("unknown argument: " + s);
+                        continue;
```
mconnew, line 157: You already have a ReportWrongArguments method. This will again ignore a bad argument and keep going which isn't what's needed.

src/CoreWCF.CrankTests/CoreWCFPerfService/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            string filePath = Path.Combine(Environment.CurrentDirectory, "CoreWCF.PerfService.exe");
+            Console.WriteLine(filePath);
+            string command = $" advfirewall firewall add rule name=\"CoreWCFPerfService\" dir=in protocol=any action=allow program=\"{filePath}\" enable=yes";
+            ExecuteCommand(command, Environment.CurrentDirectory, TimeSpan.FromSeconds(20));
```
mconnew, line 26: See my comment on the WCF client crank test PR about modifying the firewall rules programmatically.

src/CoreWCF.CrankTests/CoreWCFPerfService/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            command = $" advfirewall firewall delete rule name=\"CoreWCFPerfService\" program=\"{filePath}\"";
+            ExecuteCommand(command, Environment.CurrentDirectory, TimeSpan.FromSeconds(20));
+            Console.WriteLine("Clean up the firewall rule.");
+        }
+
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+            .UseKestrel(options => { options.ListenLocalhost(8080); })
+            .UseNetTcp(8808)
```
mconnew, line 46: Probably don't want to use ListenLocalhost as that won't accept requests from other hosts. You probably want to listen on IPv6Any

src/CoreWCF.CrankTests/CoreWCFPerfService/Program.cs
JiakaiDou: I used the firewall rules you recommended, but when I build the project with dotnet build command, I get the error "error MSB4803: The task "ResolveComReference" is not supported on the .NET Core version of MSBuild."

PR #801, opened 2022-08-13
Introduce MessageEncoder.ReadMessageAsync(ReadOnlySequence<byte> buffer)

g7ed6e:


src/CoreWCF.Primitives/src/CoreWCF/Runtime/ReadOnlySequenceExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            : stackalloc char[MaxStackAllocSize];
+        try
+        {
+            foreach (var memory in slice)
+            {
+                preProcessedBytes += memory.Length;
+                bool isLast = preProcessedBytes == slice.Length;
+                var emptyCharSlice =
+                    characterSpan.Slice(processedCharacters, stringLength - processedCharacters);
+
```
mconnew, line 32: Both stackalloc and the rented allocation could be (probably are) longer than stringLength. If you pre-slice characerSpan with Slice(0, stringLength), then here you can just use characterSpan.Slice(preprocessedCharacters) and not provide the length.

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
```diff
@@ -538,20 +536,20 @@ private ArraySegment<byte> ExtractSessionInformationFromMessage(ArraySegment<byt
...
                             int utf8ValueSize = decoder.Value;
                             offset += bytesDecoded;
                             size -= bytesDecoded;
                             if (utf8ValueSize > size)
                             {
                                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidDataException(SR.BinaryEncoderSessionMalformed));
                             }
-                            string value = Encoding.UTF8.GetString(buffer, offset, utf8ValueSize);
+                            string value = messageData.Slice(offset, utf8ValueSize).ParseAsUTF8String(utf8ValueSize);
                             offset += utf8ValueSize;
```
mconnew, line 553: We are told the size of the string in bytes (that's what the IntDecoder is returning) which is why you call `messageData.Slice(offset, utf8ValueSize)`, you end up with a ReadOnlySequence which represents the string and only the string. As this work is targeting CoreWCF 2.0, we can rely on newer api's. There's an extension method `EncodingExtensions.GetString` which takes a `ReadOnlySequence<byte>` and returns the string that was introduced in .NET 5. You can find the docs [here](https://docs.microsoft.com/en-us/dotnet/api/system.text.encodingextensions.getstring?view=net-5.0). That way you don't need your own ParseAsUTF8String implementation.

src/CoreWCF.Primitives/src/CoreWCF/Runtime/ReadOnlySequenceExtensions.cs
g7ed6e: good catch thanks i fixed it

g7ed6e:
``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19044.1889 (21H2)
Intel Xeon W-11955M CPU 2.60GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.400
  [Host]                                   : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT
  VeryLongRun                              : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT
  refactor/use-pipereader-to-read-messages : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT

Server=True  IterationCount=500  LaunchCount=4  
WarmupCount=30  

```
|   Method |                                      Job | EnvironmentVariables |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |  Gen 0 | Allocated |
|--------- |----------------------------------------- |--------------------- |---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|
| Buffered |                              VeryLongRun |     CoreWCFBasePath= | 74.82 μs | 0.184 μs | 2.487 μs | 74.97 μs |  1.00 |    0.00 | 0.1221 |     22 KB |
| Buffered | refactor/use-pipereader-to-read-messages |                Empty | 74.96 μs | 0.163 μs | 2.206 μs | 75.07 μs |  1.00 |    0.05 |      - |     22 KB |

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,74 +123,81 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
-                if (bytesRead == 0) // EOF
+                ReadOnlySequence<byte> buffer;
+                while (true)
                 {
-                    if (ContentLength != -1)
+                    ReadResult result = await reader.ReadAsync();
+                    buffer = result.Buffer;
+
+                    reader.AdvanceTo(buffer.Start, buffer.End);
+
```
mconnew, line 157: I presume the idea here is to force it to keep on reading without consuming anything until the entire request has been read. The [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.pipelines.pipereader.advanceto?view=dotnet-plat-ext-8.0#system-io-pipelines-pipereader-advanceto(system-sequenceposition-system-sequenceposition)) for PipeReader.AdvanceTo explicitly states "The Buffer previously returned from ReadAsync(CancellationToken) must not be accessed after this call.". While currently you're getting away with it due to using PipeReader.Create, when we move on to .NET 6 or later, we won't be getting the body as a Stream, we'll get it as a PipeReader and that might not work. You should move the AdvanceTo to be after the check for IsCompleted and then call AdvanceTo in the finally block before calling CompleteAsync. This way the returned buffer is guaranteed to be valid when calling DecodeBufferedMessageAsync.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,74 +123,81 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
 
-        private async Task<Message> DecodeBufferedMessageAsync(ArraySegment<byte> buffer, Stream inputStream)
+        private async ValueTask<Message> DecodeBufferedMessageAsync(ReadOnlySequence<byte> buffer)
         {
             try
             {
-                // if we're chunked, make sure we've consumed the whole body
-                if (ContentLength == -1 && buffer.Count == _settings.MaxReceivedMessageSize)
+                // if we're chunked, ensure message size is lower than or equal to MaxReceivedMessageSize
+                if (ContentLength == -1 && buffer.Length > _settings.MaxReceivedMessageSize)
```
mconnew, line 131: This check needs to move to `ReadBufferedMessageAsync(Stream inputStream)` and check that buffer.Length isn't more the MaxReceivedMessageSize after each call to ReadAsync. The purpose of this is if you only expect small messages to be received, eg less than 1MB, and a malicious client sends you a 1GB request, you want to stop allocating memory reading the request after 1MB. This current implementation would read the entire 1GB and then decide it's over 1MB.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,74 +123,81 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
             {
-                return _messageEncoder.ReadMessage(await BufferMessageStreamAsync(inputStream, _bufferManager, _settings.MaxBufferSize), _bufferManager, ContentType);
+                ReadOnlySequence<byte> buffer;
+                while (true)
+                {
+                    ReadResult result = await reader.ReadAsync();
+                    buffer = result.Buffer;
+
+                    if (ContentLength == -1 && buffer.Length >= _settings.MaxBufferSize)
+                    {
```
mconnew, line 184: This method is only called when ContentLength == -1, no need to double check. I see you are doing the correct length check here. It looks like we don't need separate methods for ReadBufferedMessageAsync and ReadChunkedBufferedMessageAsync as the implementation is looking like it should be the same now, unless I missed something.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,74 +123,81 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
+                        throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(MaxMessageSizeStream.CreateMaxReceivedMessageSizeExceededException(_settings.MaxBufferSize));
+                    }
+
+                    reader.AdvanceTo(buffer.Start, buffer.End);
+
```
mconnew, line 189: Same issue as above, AdvanceTo technically invalidates the buffer.

src/CoreWCF.Http/src/CoreWCF/Channels/WebSocketTransportDuplexSessionChannel.cs
```diff
@@ -753,8 +754,8 @@ private async Task<Message> PrepareMessageAsync(WebSocketReceiveResult result, b
                     }
                     else
                     {
-                        ArraySegment<byte> bytes = new ArraySegment<byte>(buffer, 0, count);
-                        message = _encoder.ReadMessage(bytes, _bufferManager);
+                        ReadOnlySequence<byte> bytes = new(buffer, 0, count);
+                        message = await _encoder.ReadMessageAsync(bytes, _bufferManager);
```
mconnew, line 758: I don't know if you want to do this here, but there's an opportunity for a perf improvement here. If you look at the code around line 520, the current code can make multiple calls to WebSocket.ReceiveAsync and if the EndOfMessage flag isn't set in the result, it calls it again. ReceiveAsync is passed an ArraySegment so if there's space in the buffer, it just passes the same buffer with the offset starting where the previous ReadAsync completed. But if the buffer is full, it allocates a new one twice the size and copies the previous once across. But when dealing with a ReadOnlySequence, you don't need a single buffer. That code around line 520 could allocate a new buffer the same size as the first one and fill that up. PrepareMessageAsync could then be passed the ReadOnlySequence containing all the buffers without any copying having occurred.

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
```diff
@@ -538,20 +536,20 @@ private ArraySegment<byte> ExtractSessionInformationFromMessage(ArraySegment<byt
...
                         int size = dictionarySize;
-                        int offset = messageData.Offset + headerSize;
+                        int offset = headerSize;
 
                         while (size > 0)
                         {
                             decoder.Reset();
-                            int bytesDecoded = decoder.Decode(buffer, offset, size);
+                            int bytesDecoded = decoder.Decode(messageData.Slice(offset, size));
                             int utf8ValueSize = decoder.Value;
```
mconnew, line 545: Inside of the `if (dictionarySize > 0) block, you could initially slice messageData to (0, headerSize) so it only contains the dictionary header. Then you don't need to do any booking of size  as you read each dictionary item as you can just do `Slice(bytesDecoded)`.

mconnew: I realized headerSize would be the wrong thing, I think it would be `Slice(headerSize, headerSize + DictionarySize) so just give the dictionary. I hope you get what I'm saying,

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
```diff
@@ -568,21 +566,22 @@ private ArraySegment<byte> ExtractSessionInformationFromMessage(ArraySegment<byt
                     }
                 }
 
-                return new ArraySegment<byte>(buffer, newOffset, newSize);
+                return messageData.Slice(headerSize + dictionarySize, messageData
+                    .Length - headerSize - dictionarySize);
```
mconnew, line 570: I think you can skip the second parameter here as it will default to the rest of the sequence after the offset.

src/CoreWCF.Primitives/src/CoreWCF/Channels/BufferedMessageData.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Buffers;
 using System.Xml;
```
mconnew, line 6: You've added this using but there's not code change.

src/CoreWCF.Primitives/src/CoreWCF/Channels/BufferManager.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Buffers;
 using CoreWCF.Runtime;
```
mconnew, line 6: You've added this using but no code change

src/CoreWCF.Primitives/src/CoreWCF/Channels/ByteStreamMessageEncoder.cs
```diff
@@ -79,7 +75,11 @@ public override Message ReadMessage(ArraySegment<byte> buffer, BufferManager buf
             //    TD.ByteStreamMessageDecodingStart();
             //}
 
-            var messageData = new ByteStreamBufferedMessageData(buffer, bufferManager);
+            int bufferLength = (int)buffer.Length;
+            byte[] bytes = bufferManager.TakeBuffer(bufferLength);
+            buffer.CopyTo(bytes.AsSpan(0, bufferLength));
+
```
mconnew, line 81: This is going to reduce performance as we're copying the ReadOnlySequence into an allocated buffer. Can we do anything in ByteStreamBufferedMessageData so use the ReadOnlySequence instead?

mconnew: I looked at the code and I believe we can just keep the ReadOnlySequence. There's some changes needed in ByteStreamMessage. Where it calls writer.WriteBase64, we should check if writer is an XmlByteStreamWriter (which we own) and case to that. Have a new overload on XmlByteStreamWriter which takes a ReadOnlySequence and have a ReadOnlySequence specific implementation. All it does is call Stream.Write. ReadOnlySequence has an enumerator which emits ReadOnlyMemory. The problem here is netstandard2.0 doesn't have a Stream.Write overload which can rake a ReadOnlyMemory. There is a MemoryMarshal helper method which can try to get you the underlying buffer which you can pass to Stream.Write, but if it can't (eg it's backed by native memory), then you need to allocate a buffer and copy it. I have code to do this in the NetNamedPipe implementation so if you are interested in getting all this working, I can point you at the code.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MtomMessageEncoder.cs
```diff
@@ -374,7 +377,13 @@ public override Message ReadMessage(ArraySegment<byte> buffer, BufferManager buf
 
             MtomBufferedMessageData messageData = _factory.TakeBufferedReader(this);
             messageData._contentType = contentType;
-            messageData.Open(buffer, bufferManager);
+
+            int bufferLength = (int)buffer.Length;
+            byte[] bytes = bufferManager.TakeBuffer(bufferLength);
+            buffer.CopyTo(bytes.AsSpan(0, bufferLength));
+
```
mconnew, line 384: We should avoid copying where necessary. Ideally MtomBufferedMessageData should handle a ReadOnlySequence natively

src/CoreWCF.Primitives/src/CoreWCF/Channels/TextMessageEncoderFactory.cs
```diff
@@ -395,18 +396,25 @@ public override bool IsContentTypeSupported(string contentType)
...
                 }
 
                 Message message;
 
                 UTF8BufferedMessageData messageData = TakeBufferedReader();
                 messageData.Encoding = GetEncodingFromContentType(contentType, _contentEncodingMap);
-                messageData.Open(buffer, bufferManager);
+
+                int bufferLength = (int)buffer.Length;
+                byte[] bytes = bufferManager.TakeBuffer(bufferLength);
```
mconnew, line 412: UTF8BufferedMessageData should handle ReadOnlySequence natively. If the ReadOnlySequence can be wrapped as a stream (and I think I saw some code elsewhere where you do that via a Sequence), the TakeReader which calls XmlDictionaryReader.CreateTextReader can call the overload of CreateTextReader which takes a stream.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MtomMessageEncoder.cs
g7ed6e: @mconnew I agree. 

The issue i have is that `IBufferedMessageData` is a public type which exposes `Buffer` through a `ArraySegment<byte>`.  Would you be open to the following update of the IBufferedMessageData

```c#
public interface IBufferedMessageData
    {
        MessageEncoder MessageEncoder { get; }
        [Obsolete("Use ReadOnlyBuffer property", true)]
        ArraySegment<byte> Buffer { get; }
        ReadOnlySequence<byte> ReadOnlyBuffer { get; }
        XmlDictionaryReaderQuotas Quotas { get; }
        void Close();
        void EnableMultipleUsers();
        XmlDictionaryReader GetMessageReader();
        void Open();
        void ReturnMessageState(RecycledMessageState messageState);
        RecycledMessageState TakeMessageState();
    }
```
Introducing this property would allow to avoid this useless buffer allocation. 
Other buffer recopy you noticed during the review should also be solved and we should be able to drop `BufferManager` usages from the encoder implementation.
The missing piece now is a `Stream.Write(ReadOnlyMemory<byte> memory)`. i would be happy if you can link the code you have in NetNamedPipe implementation.

src/CoreWCF.Primitives/src/CoreWCF/Channels/TextMessageEncoderFactory.cs
g7ed6e: @mconnew fixed in https://github.com/CoreWCF/CoreWCF/pull/801/commits/5106720621414b5d49047297eec45cc8a30aec1e

src/CoreWCF.Primitives/src/CoreWCF/Channels/MtomMessageEncoder.cs
g7ed6e: @mconnew I push this [commit](https://github.com/CoreWCF/CoreWCF/commit/5106720621414b5d49047297eec45cc8a30aec1e) that implement the intention described above

src/CoreWCF.Primitives/src/CoreWCF/Channels/BufferManager.cs
g7ed6e: done thanks

src/CoreWCF.Primitives/src/CoreWCF/Channels/BufferedMessageData.cs
g7ed6e: done thanks

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
g7ed6e: Added a TODO so we can ctach thos once moved toCoreWCF2.0 https://github.com/CoreWCF/CoreWCF/pull/801/commits/d76d9b56a51f1ab211175beeec2004feac504ad3

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: Thanks. done [here](https://github.com/CoreWCF/CoreWCF/pull/801/commits/47b8404eb568804204b09fa1a3d5a5de9beec845)
I had to add an additional commit to avoid the buffers from PipeReader being freed before DispatchAsync is called . please see https://github.com/CoreWCF/CoreWCF/pull/801/commits/58f962a8249494517772a305c3fbb18a3b851228

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
g7ed6e: done https://github.com/CoreWCF/CoreWCF/pull/801/commits/652e27f4d056acbe8836bfa2eb4ebb7635c0dd76

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
g7ed6e: I see it. implemented https://github.com/CoreWCF/CoreWCF/pull/801/commits/652e27f4d056acbe8836bfa2eb4ebb7635c0dd76

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
```diff
@@ -13,45 +17,30 @@ internal static class MessageEncoderCompressionHandler
...
-                byte[] decompressedBytes = bufferedOutStream.ToArray(out int length);
-                bufferManager.ReturnBuffer(buffer.Array);
-                buffer = new ArraySegment<byte>(decompressedBytes, buffer.Offset, length - buffer.Offset);
-            }
+            using Stream gZipStream =
+                compressionFormat == CompressionFormat.GZip
+                    ? new GZipStream(pipe.Reader.AsStream(), CompressionMode.Decompress)
+                    : new DeflateStream(pipe.Reader.AsStream(), CompressionMode.Decompress);
+
+            using var ms = new MemoryStream();
```
g7ed6e, line 41: do something else

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,80 +123,62 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
-            }
+                    if (result.IsCompleted)
+                    {
+                        break;
+                    }
 
-            return await DecodeBufferedMessageAsync(new ArraySegment<byte>(buffer, 0, offset), inputStream);
-        }
+                    reader.AdvanceTo(buffer.GetPosition(0), buffer.End);
+                }
```
mconnew, line 167: Did buffer.Start not work?

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,80 +123,62 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
-                return _messageEncoder.ReadMessage(await BufferMessageStreamAsync(inputStream, _bufferManager, _settings.MaxBufferSize), _bufferManager, ContentType);
+                message = await DecodeBufferedMessageAsync(buffer);
             }
-            catch (XmlException xmlException)
+            finally
             {
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
-                    new ProtocolException(SR.MessageXmlProtocolError, xmlException));
+                reader.AdvanceTo(buffer.Start, buffer.End);
             }
```
mconnew, line 174: This line doesn't tell the reader it can release the buffer. The purpose of reader.AdvanceTo is to tell the reader what you have consumer so it can release the buffers that were allocated. This should be reader.AdvanceTo(buffer.End) to tell the reader you have consumed everything and it can return the buffers.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -121,80 +123,62 @@ private void ThrowMaxReceivedMessageSizeExceeded()
...
+
+            return message;
         }
 
-        private async Task<Message> ReadStreamedMessageAsync(Stream inputStream)
+        private async Task<Message> ReadStreamedMessageAsync(PipeReader reader)
         {
-            MaxMessageSizeStream maxMessageSizeStream = new MaxMessageSizeStream(inputStream, _settings.MaxReceivedMessageSize);
+            MaxMessageSizeStream maxMessageSizeStream = new(reader.AsStream(), _settings.MaxReceivedMessageSize);
```
mconnew, line 182: Not wanting to feature creep here so this can happen as a follow up issue as we need to get code completed and merged at some point, but we could create a MaxMessageSizePipeReader which wraps the PipeReader. It would be a small performance improvement, but cleaner.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
```diff
@@ -401,30 +352,27 @@ public async Task<(Message message, Exception requestException)> ParseIncomingMe
...
                     {
-                        return (null, requestException);
+                        return (null, null, null);
                     }
                 }
                 else
                 {
                     Stream stream = GetInputStream(true);
+                    reader = PipeReader.Create(stream, new StreamPipeReaderOptions(bufferSize: ConnectionOrientedTransportDefaults.ConnectionBufferSize));
                     if (_streamed)
```
mconnew, line 362: The buffer size should be ultimately coming from HttpTransportBindingElement.MaxBufferSize. I can't remember if that's on the settings class already, I think it might be.

src/CoreWCF.NetTcp/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
```diff
@@ -215,8 +215,8 @@ private async ValueTask<(Message, ReadOnlySequence<byte>)> DecodeMessageAsync(Re
                                 Message message;
                                 try
                                 {
-                                    message = _connection.MessageEncoder.ReadMessage(
-                                        new ArraySegment<byte>(_connection.EnvelopeBuffer.ToArray(), 0, _connection.EnvelopeSize),
+                                    message = await _connection.MessageEncoder.ReadMessageAsync(
+                                        new ReadOnlySequence<byte>(_connection.EnvelopeBuffer.ToArray(), 0, _connection.EnvelopeSize),
                                         _connection.BufferManager,
```
mconnew, line 220: Calling ToArray creates a new array and copies the contents of the EnvelopeBuffer to that array. EnvelopeBuffer is a Memory\<byte\> which can be passed to the constructor of ReadOnlySequence as-is

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
```diff
@@ -606,15 +601,19 @@ public override Message ReadMessage(ArraySegment<byte> buffer, BufferManager buf
                 Message message;
                 if (_messagePatterns != null)
                 {
-                    message = _messagePatterns.TryCreateMessage(buffer.Array, buffer.Offset, buffer.Count, bufferManager, messageData);
+                    int bufferLength = (int)buffer.Length;
+                    byte[] bytes = bufferManager.TakeBuffer(bufferLength);
+                    buffer.CopyTo(bytes.AsSpan(0, bufferLength));
+                    message = _messagePatterns.TryCreateMessage(bytes, 0, bufferLength, bufferManager, messageData);
                 }
```
mconnew, line 608: We should have a follow up task to convert MessagePatterns to use ReadOnlySequence

src/CoreWCF.Primitives/src/CoreWCF/Channels/ByteStreamMessageEncoder.cs
mconnew, line 81: This is still extra allocation. ByteSTreamBufferedMessageData needs to be converted to use a ReadOnlySequence on its constructor.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
```diff
@@ -13,45 +17,30 @@ internal static class MessageEncoderCompressionHandler
...
-            MemoryStream memoryStream = new MemoryStream(buffer.Array, buffer.Offset, buffer.Count);
-            int maxDecompressedSize = (int)Math.Min(maxReceivedMessageSize, int.MaxValue);
+            if (buffer.Length > maxReceivedMessageSize)
+            {
+                throw Fx.Exception.AsError(new QuotaExceededException(SR.Format(SRCommon.MaxReceivedMessageSizeExceeded, maxReceivedMessageSize)));
+            }
 
-            using (BufferManagerOutputStream bufferedOutStream = new BufferManagerOutputStream(SRCommon.MaxReceivedMessageSizeExceeded, 1024, maxDecompressedSize, bufferManager))
+            Pipe pipe = new();
+            foreach (ReadOnlyMemory<byte> memory in buffer)
```
mconnew, line 28: Using a pipe creates an additional buffer that you're effectively copying everything into just to create a PipeReader which you call AsStream against to pass to the compression streams. Just use PipeReader.Create(ReadOnlySequence).AsStream() and skip creating a Pipe and writing into it.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
```diff
@@ -13,45 +17,30 @@ internal static class MessageEncoderCompressionHandler
...
-                bufferManager.ReturnBuffer(buffer.Array);
-                buffer = new ArraySegment<byte>(decompressedBytes, buffer.Offset, length - buffer.Offset);
-            }
+            using Stream gZipStream =
+                compressionFormat == CompressionFormat.GZip
+                    ? new GZipStream(pipe.Reader.AsStream(), CompressionMode.Decompress)
+                    : new DeflateStream(pipe.Reader.AsStream(), CompressionMode.Decompress);
+
+            using var ms = new MemoryStream();
+            await gZipStream.CopyToAsync(ms);
```
mconnew, line 42: BufferManagerOutputStream is better performing. MemoryStream initially creates a byte[] with a size which is the power of 2 immediately larger than the first write. Then with each write, if it needs more space, it allocates a new array at least double the size of the current array (or if that's not big enough, the next power of 2 size) and copies the data over. This results in a lot of allocations and a lot of copying. BufferManagerOutputStream creates an initial buffer size that you specify, then when that's full, creates an additional buffer twice the size and starts filling that up. It keeps on allocating new buffers with no copying, but it uses the BuffferManager to do the allocation so hopefully no GC needed. When it's time to convert it to an array, it know how big of a buffer it needs so allocates the correct size byte[] and copies its contents over. A lot less copying, and a better allocation pattern.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
```diff
@@ -13,45 +17,30 @@ internal static class MessageEncoderCompressionHandler
         internal const string DeflateContentEncoding = "deflate";
         private const int DecompressBlockSize = 1024;
 
-        internal static void DecompressBuffer(ref ArraySegment<byte> buffer, BufferManager bufferManager, CompressionFormat compressionFormat, long maxReceivedMessageSize)
+        internal static async ValueTask<ReadOnlySequence<byte>> DecompressBufferAsync(ReadOnlySequence<byte> buffer, CompressionFormat compressionFormat, long maxReceivedMessageSize)
         {
```
mconnew, line 21: No need to make the method async as the source data is coming from a buffer. You can call the synchronous Stream.CopyTo() and not need anything async.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: @mconnew i think we might still have an issue as we do need the buffer in the DispatchAsync call which is made a bit later in AspNetCoreReplyChannel.HandleRequest. For the time being i'm delaying the buffer being freed by returning the PipeReader created by HttpInput.ParseIncomingMessageAsync to AspNetCoreReplyChannel.HandleRequest and call PipeReader.CompleteAsync there. 
I'm wondering if this is correct or not.. and i'm thinking to move the call to AdvanceTo(buffer.Start, buffer.End) after the DispatchAsync.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: @mconnew the PipeReaderOptions allows to configure a min buffer size but not a max buffer size. https://learn.microsoft.com/en-us/dotnet/api/system.io.pipelines.streampipereaderoptions.buffersize?view=dotnet-plat-ext-7.0#system-io-pipelines-streampipereaderoptions-buffersize

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
g7ed6e: updated to use BufferManagerOutputStream

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Channels/ByteStreamMessageEncoder.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Channels/MessageEncoderCompressionHandler.cs
g7ed6e: used BufferManagerOutputStream

src/CoreWCF.Primitives/src/CoreWCF/Channels/BinaryMessageEncoderFactory.cs
g7ed6e: done

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: it works indeed. done

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: MemoryPool<T> does allow to configure a max buffer size and it is possible to specify a custom pool in PipeReaderOptions. https://learn.microsoft.com/en-us/dotnet/api/system.buffers.memorypool-1?view=net-7.0&viewFallbackFrom=dotnet-plat-ext-7.0. 
But once we move to .NET and use BodyReader property I d'ont think we will be able to specify the pool.

src/CoreWCF.Primitives/src/CoreWCF/Channels/ByteStreamMessageEncoder.cs
g7ed6e: @mconnew this is now done

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: I think this is now fixed.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
g7ed6e: this shoudl be ok too now. I restored the ReadBufferedMessageAsync name

src/CoreWCF.Http/src/CoreWCF/Channels/WebSocketTransportDuplexSessionChannel.cs
g7ed6e: @mconnew would you accept a follow up PR dedicated to the perf imrpovement ?

PR #811, opened 2022-08-25
Remove superfluous DBNull. Just use System.DBNull instead

Sologix:
We don't need our own DBNull. Let's just use System.DBNull instead.


mconnew:
You're correct in that this can now go away. Unfortunately I made the mistake of making it public so it's technically part of the public API surface. This is going to need to wait until we start checking in 2.0 specific work. Right now I'm cleaning up and merging things which will go into a 1.x release. It should be soon that we get to 2.0 specific work.

PR #812, opened 2022-08-26
Fix multiple WSDL issues

### LINKED ISSUE #654: WSDL missing xml header
An XML header like this:
```xml
<?xml version="1.0" encoding="utf-8"?>
```
Is missing from the generated WSDL

mconnew:
Fixes #654 (Add xml declaration to WSDL document)
Also changed default behavior to leave KeyValuePairOfXXXX in the wsdl as some people actually use KeyValuePair\<K,V\> in their contracts and removing this broke their WSDL. It should be harmless in the output if it isn't needed, but in case it does cause problems, I've added an AppContext switch to turn the behavior back on. This won't be a problem for .NET 7 as they fixed the issue there.

### DIFF
--- a/src/CoreWCF.Metadata/tests/DataTypesTest.cs
+++ b/src/CoreWCF.Metadata/tests/DataTypesTest.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


mconnew:
This didn't get caught in the original tests as XML canonicalization strips out the XML declaration.

g7ed6e (approved):

mconnew:
@g7ed6e, can you take another look as I've added another fix.

g7ed6e:
it looks good to me

PR #819, opened 2022-09-03
Fixup the incorrect KeyValuePairOf schema

mconnew:
I still need to create tests for this.

### DIFF
--- a/src/CoreWCF.Metadata/tests/Contracts/ICollectionOfKeyValuePairDataService.cs
+++ b/src/CoreWCF.Metadata/tests/Contracts/ICollectionOfKeyValuePairDataService.cs
@@ -1,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using CoreWCF;
+
+namespace ServiceContract
+{
+    [ServiceContract(Namespace = Constants.NS, Name = "CollectionOfKeyValuePairDataService")]
+    public interface ICollectionOfKeyValuePairDataService
+    {
+        [OperationContract]
+        KeyValueContainingArray EchoKeyValueWithArray(KeyValueContainingArray echo);
+
+        [OperationContract]
+        KeyValueContainingList EchoKeyValueWithList(KeyValueContainingList echo);
+    }
+
+    [DataContract]
+    public class KeyValueContainingArray
+    {
+        [DataMember]
+        public KeyValuePair<string,int>[] KeyValueArray { get; set; }
+    }
+
+    [DataContract]
+    public class KeyValueContainingList
+    {
+        [DataMember]
+        public List<KeyValuePair<string, int>> KeyValueList { get; set; }
+    }
+}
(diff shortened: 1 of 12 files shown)


AAATechGuy:
thanks @mconnew for the quick fixes.

mconnew:
Also added fix for the schema provided by a class not getting used as XmlDataContract doesn't save the XsdType.

g7ed6e (approved):
There's a missing private access modifier in `DataContractSerializerOperationFormatter`

mconnew:
Good catch, I'll fix the private accessor then merge.

mconnew:
Tests were failing on NetFx because DataSet produces a different schema on NetFx and NetCore/Net. On NetFx, it by default produces a schema aimed at the WSDL v1.0 way of expressing things with it switchable to v2.0 via config (which nobody really ever did). On NetCore/Net, because config isn't there, it defaults to outputting the WSDL v2.0 style schema. The tests are written to expect the NetCore behavior, so the test fails on NetFx. Disabling the test for NetFx as we're not specifically testing that the schema is correct, but that the schema output from a class providing its schema is included in the WSDL.

Aleksanderis:
@mconnew is it available in a development build? I was trying it with 1.2.0-alpha.33 and have a few questions. Not sure if it's an issue, or a feature, or just something still in progress.
I was trying to use service using WCF Test Client, now KeyValuePairOf seems is not causing issues anymore - metadata loads and the service can be called. However there are two issues:

- it loads, but there is an error logged (in a stable version there is no such error)
```fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
      CoreWCF.ProtocolException: Content Type application/soap+xml; charset=utf-8 was sent to a service expecting text/xml; charset=utf-8.  The client and service bindings may be mismatched.
         at CoreWCF.Channels.HttpInput.ThrowHttpProtocolException(String message, HttpStatusCode statusCode, String statusDescription)
         at CoreWCF.Channels.HttpInput.ValidateContentType()
         at CoreWCF.Channels.HttpInput.ParseIncomingMessageAsync()
         at CoreWCF.Channels.AspNetCoreReplyChannel.HandleRequest(HttpContext context)
         at CoreWCF.Channels.RequestDelegateHandler.HandleRequest(HttpContext context)
         at CoreWCF.Channels.ServiceModelHttpMiddleware.InvokeAsync(HttpContext context)
         at CoreWCF.Channels.MetadataMiddleware.InvokeAsync(HttpContext context)
         at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
         at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

- it only works, when I use an URL with "?wsdl" query string - like https://localhost:5001/Myservice.svc?wsdl. Previously it was working without it - like https://localhost:5001/Myservice.svc. I guess it was treating "?wsdl" as default behaviour. It's a minore thing I think, but still a bit breaking.

mconnew:
We don't support WS-Metadata yet, which is a Soap contract to basically get the WSDL. For now you need to use either ?wsdl or ?singlewsdl. I know svcutil can get the disco file first to find the wsdl so I'm thinking about implementing that to make that scenario cleaner. I don't know whether the test client can do the same.

Aleksanderis:
What about content-type issue? I didn't mention it specifically - it's thrown, even if I'm using "?wsdl" URL. In fact everything works, and that errors looks more like warning, but I'm not sure.

PR #820, opened 2022-09-03
Fix OperationParameterInjectionGenerator performance issues

### LINKED ISSUE #779: when use scoped dependency injection,visual studio 2022 will consume a lot of CPU
when I use scoped dependency injection follw the [sample](https://github.com/CoreWCF/samples/tree/main/Scenarios/Logging-and-Dependency-Injection/LoggingSample) ,visual studio 2022 will consume a lot of CPU
image
I found it was caused by corewcf.buildtools.roslyn4.0，And the more and more complex methods in the service class, the more CPU the analyzer consumes

g7ed6e:
This should help with #779 

I completely rewrote the parser logic in LINQ as the original logic had multiple issues. (multiple iteration on referenced assemblies, acquiring semantic model multiples times, trying to manually do builtin things..).
Thanks to existing unit tests i'm really confident with this new version.
Also the source generator is now more reactive and detects instantly the missing partial keyword to report an error and as soon as partial is applied the build error disappear even without saving the source code file.

Resolves #779


src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -51,12 +53,14 @@ public void Decrement()
...
+            private readonly SourceGenerationSpec _generationSpec;
 
-            public Emitter(in OperationParameterInjectionSourceGenerationContext sourceGenerationContext, SourceGenerationSpec generationSpec)
+            public Emitter(in OperationParameterInjectionSourceGenerationContext sourceGenerationContext, in SourceGenerationSpec generationSpec)
             {
                 _sourceGenerationContext = sourceGenerationContext;
                 _generationSpec = generationSpec;
+                _builder = new StringBuilder();
+                _builder.Clear();
             }
```
mconnew, line 64: Why are you calling Clear() on a new StringBuilder instance?

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -51,12 +53,14 @@ public void Decrement()
...
-            private SourceGenerationSpec _generationSpec;
+            private readonly SourceGenerationSpec _generationSpec;
 
-            public Emitter(in OperationParameterInjectionSourceGenerationContext sourceGenerationContext, SourceGenerationSpec generationSpec)
+            public Emitter(in OperationParameterInjectionSourceGenerationContext sourceGenerationContext, in SourceGenerationSpec generationSpec)
             {
                 _sourceGenerationContext = sourceGenerationContext;
                 _generationSpec = generationSpec;
+                _builder = new StringBuilder();
+                _builder.Clear();
```
mconnew, line 63: I'm trying to work out what benefit moving this to an instance field has? I think there might be a bug introduced here. The string builder isn't cleared before each call to EmitOperationContract. If _generationSpec.OperationContractSpecs contains more than one operationContractSpec, the second call to EmitOperationContract will have both sets of code accumulated in the string builder. The second and subsequent calls to _sourceGenerationContext.AddSource will add a file which contains all the prior generated code too.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
g7ed6e: Thanks! I moved the `Clear` in the foreach loop.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
g7ed6e: Thanks, there was indeed a bug - i added a unit test to confirm and fix in the latest commit i pushed.
The benefit is to tie the StringBuilder instance lifetime per generation pass instead of per file to generate.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
```diff
@@ -51,26 +53,28 @@ public void Decrement()
...
                 _generationSpec = generationSpec;
+                _builder = new StringBuilder();
             }
 
             public void Emit()
             {
                 foreach (var operationContractSpec in _generationSpec.OperationContractSpecs)
                 {
+
                     EmitOperationContract(operationContractSpec);
```
mconnew, line 70: You've added an extra blank line by mistake.

mconnew (approved):
Once you've removed the extra blank line, feel free to merge.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.Emitter.cs
g7ed6e: I just dropped it.

Issue #700, opened 2022-06-02
BasiHttpBinding + https stripping in k8s = wrong soap address location in wsdl


betsson-vyza:
Hi, could you please advise if there is a solution to the problem.
GIVEN:
- CoreWCF service hosted in k8s;
- communication to cluster only by https, which is stripped by proxy;
- docker container has:
     EXPOSE 5000
     ENV ASPNETCORE_URLS=http://+:5000
- [Program.cs](https://github.com/CoreWCF/CoreWCF/files/8822253/Program.txt) attached
WHEN:
- I open https://my-test.tech/EchoService?singleWsdl
THEN:
- wsdl contains http address location, meaning incorrect address, wcf test client can't reach it, nobody can without knowledge that address must be https://
![image](https://user-images.githubusercontent.com/68384224/171594498-d9ecad1c-17a8-4570-b068-75ecfe611d14.png)

So the question, is there an easy way to have proper link with https in wsdl for basichttp binding?

betsson-vyza:
So the question is how to override endpoint address and binding's security just for wsdl.

mconnew:
I've made a special label to group all the wsdl issues so that I can address as many as possible when I revisit wsdl's soon.

coenm:
Is there a fix for this?

g7ed6e:
A possible fix would be to introduce a `CustomMetadataAddressProviderBehavior` that pull let's say a `IMetadataAddressProvider` from DI. This new behavior would bypass `UseRequestHeadersForMetadataAddressBehavior` if present in DI. 
This would bring full control to user to provide a custom behavior relying on DI services (`IHttpContextAccessor`, `IConfiguration` ..).

@mconnew  I think i can submit a PR for this if the solution seems acceptable.

MarkTallentire:
@g7ed6e 

We've just run into this issue and unfortunately I don't understand the fix from the PR. Are you able to provide an example here.

Just to be clear our issue is that we need the WSDL to show https and only https but can only make it either show http with no https or both http and https.

PR #822, opened 2022-09-08
Introduce CustomEndpointAddressForMetadataBehavior

g7ed6e:
- Add UseRequestHeadersForMetadataAddressBehavior unit test
- Introduce CustomEndpointAddressForMetadataBehavior to give user control over the endpoint address in metadata. This should allow SSL termination scenario.
relates to #700


src/CoreWCF.Primitives/src/CoreWCF/Description/IEndpointAddressForMetadataProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Description
+{
+    public interface IEndpointAddressForMetadataProvider
+    {
+        System.Uri GetEndpointAddress();
+    }
```
mconnew, line 9: I think we should make this more flexible and have it accept the HttpRequest (or HttpContext, but HttpRequest feels like the better match to me, but not a strong opinion on it) as a parameter. That way an implementation can do things like read the X-Forwarded... headers etc or do something more custom like return a different address for WAN and LAN clients.

g7ed6e: I updated the signature to accept the `HttpRequest`

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
```diff
@@ -139,11 +139,20 @@ private void SetExtensionProperties(ServiceDescription description, ServiceHostB
             mex.HttpGetUrl = host.GetVia(Uri.UriSchemeHttp, _httpGetUrl ?? new Uri(string.Empty, UriKind.Relative));
             mex.HttpsGetUrl = host.GetVia(Uri.UriSchemeHttps, _httpsGetUrl ?? new Uri(string.Empty, UriKind.Relative));
 
-            UseRequestHeadersForMetadataAddressBehavior dynamicUpdateBehavior = description.Behaviors.Find<UseRequestHeadersForMetadataAddressBehavior>();
-            if (dynamicUpdateBehavior != null)
+            UseRequestHeadersForMetadataAddressBehavior useRequestHeadersForMetadataAddressBehvaior = description.Behaviors.Find<UseRequestHeadersForMetadataAddressBehavior>();
+            if (useRequestHeadersForMetadataAddressBehvaior != null)
             {
```
mconnew, line 144: I think we could simplify this by having UseRequestHeadersForMetadataAddressBehavior provide an implementation of IEndpointAddressForMetadataProvider. We could get rid of UpdateAddressDynamically and have checks for that check mex.CustomEndpointAddressProvider != null instead. Of have UpdateAddressDynamically be defined as `get => CustomEndpointAddressProvider != null;`.

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomEndpointAddressForMetadataBehavior.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public CustomEndpointAddressForMetadataBehavior(IEndpointAddressForMetadataProvider provider)
+        {
+            Provider = provider;
+        }
+
+        void IServiceBehavior.Validate(ServiceDescription serviceDescription, ServiceHostBase serviceHostBase) { }
+        void IServiceBehavior.AddBindingParameters(ServiceDescription serviceDescription, ServiceHostBase serviceHostBase, Collection<ServiceEndpoint> endpoints, BindingParameterCollection bindingParameters) { }
+        void IServiceBehavior.ApplyDispatchBehavior(ServiceDescription serviceDescription, ServiceHostBase serviceHostBase) { }
+    }
```
mconnew, line 21: Instead of simply being a container which is explicitly checked for by ServiceMetadataBehavior.ApplyDispatchBehavior, instead you could have this code directly apply the provider. Something like this:
```c#
var mexExt = host.Extensions.Find<ServiceMetadataExtension>();
mexExt.CustomEndpointAddressProvider = Provider;
```
To prevent problems with behavior order being a problem, you could use the internal api `ServiceMetadataExtension.EnsureServiceMetadataExtension(serviceHostBase)` to make sure an instance exists and return that instance instance of using host.Extensions.Find

src/CoreWCF.Primitives/src/CoreWCF/Description/IEndpointAddressForMetadataProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.AspNetCore.Http;
+
+namespace CoreWCF.Description
+{
+    public interface IEndpointAddressForMetadataProvider
+    {
```
mconnew, line 9: Using the word "For" in an interface name seems a little off. Maybe switch the subject and function to remove it, e.g. IMetadataEndpointAddressProvider

g7ed6e: Thanks i was not confortable with the name i choosed. I renamed to `IMetadataEndpointAddressProvider`

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomEndpointAddressForMetadataBehavior.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Description/IEndpointAddressForMetadataProvider.cs
mconnew: The 3 hardest problems in software development are 1) Naming, 2) Off by one errors

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomEndpointAddressForMetadataBehavior.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+
+        }
+
+        void IServiceBehavior.ApplyDispatchBehavior(ServiceDescription serviceDescription,
+            ServiceHostBase serviceHostBase)
+        {
+            ServiceMetadataExtension.EnsureServiceMetadataExtension(serviceHostBase);
+            var mex = serviceHostBase.Extensions.Find<ServiceMetadataExtension>();
+            mex.DynamicMetadataEndpointAddressProvider = _provider;
```
mconnew, line 35: I think EnsureServiceMetadataExtension actually returns the instance too. No need to find it on the extensions after

g7ed6e: Thanks i just pushed

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
```diff
@@ -87,40 +87,31 @@ public WsdlNS.ServiceDescription SingleWsdl
...
-
-            host = hostString.Host;
-            if (hostString.Port.HasValue)
-            {
-                port = hostString.Port.Value;
-            }
-            else
+            Uri customUri = DynamicMetadataEndpointAddressProvider?.GetEndpointAddress(httpRequest);
+            if (customUri != null)
             {
```
mconnew, line 105: This is a change in behavior. If DynamicMetadataEndpointAddressProvider isn't set (the default with no behaviors), we are now returning false. Previously it would directly pull from the request and if there was a Host header, it would return true. Now it returns false if an address provider isn't set.

g7ed6e: good catch. Reviewing carefully I also noticed that I may have changed behavior because previously  we were using the `_listenUri` to deduct the port using `Uri.TryCreate`. Now i no longer have access to the listenUri in the `MetadataEndpointAddressProvider` and i do not think we want to polluate the signature of `GetEndpointAddress` with it. I don't know what is the best way to get it in the `GetEndpointAddress` implementation.

g7ed6e: > This is a change in behavior. If DynamicMetadataEndpointAddressProvider isn't set (the default with no behaviors), we are now returning false. Previously it would directly pull from the request and if there was a Host header, it would return true. Now it returns false if an address provider isn't set.

I just checked and `TryGetHttpHostAndPort` is always called from `GetDynamicAddressWriter` calls which are always surrounded by a `if (UpdateAddressDynamically)` which was previously only set when `UseRequestHeadersForMetadataAddressBehavior` was applied. We can then deduct TryGetHttpHostAndPort is executed only when an address provider in the new code given that UseRequestHeadersForMetadataAddressBehavior  also provides an address provider.

mconnew: I just worked that out (the if check) trying to go through the change very carefully. HttpContext has an [Items property bag](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext.items?view=aspnetcore-6.0) you can put things in. You could stash the listenUri in there. I generally use the convention of the key being a string with the fully qualified class name of the either the object (which doesn't apply here as Uri is generic), or the owner or interested party of the object. That way it's impossible to have an accidental key name collision. In this case the value comes from `CoreWCF.Description.ServiceMetadataExtension.HttpGetImpl._listenUri` and is initially passed into a method in `TryHandleMetadataRequestAsync`. You could put the listenUri in the Items property bag with that fully qualified field name that I just mentioned.

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomEndpointAddressForMetadataBehavior.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.ObjectModel;
+using CoreWCF.Channels;
+
+namespace CoreWCF.Description
+{
+    public class CustomEndpointAddressForMetadataBehavior : IServiceBehavior
+    {
```
mconnew, line 10: I just realized, we fixed the name of the interface, but this class needs a better name. I think MetadataEndpointAddressServiceBehavior is a good name.

mconnew (approved):
Please fix the class name as mentioned in PR comment then you can merge.

src/CoreWCF.Primitives/src/CoreWCF/Description/CustomEndpointAddressForMetadataBehavior.cs
g7ed6e: I pushed the renaming of the file and also align filename with type names from previous rename.

Issue #824, opened 2022-09-08
MessageContractHelper includes MessageContract attributes from base types


lukak-msft:
MessageContractHelper includes MessageContract attributes from base types, which leads to a Null Reference Exception later when 
CreateTypedMessageDescription attemptes to retrieve the same attribute from the message type.

```
 System.NullReferenceException: Object reference not set to an instance of an object.
   at CoreWCF.Description.TypeLoader`1.CreateTypedMessageDescription(Type typedMessageType, ICustomAttributeProvider returnAttrProvider, XmlName returnValueName, String defaultNS, String action, MessageDirection direction)
   at CoreWCF.Description.TypeLoader`1.CreateMessageDescription(MethodInfo methodInfo, Boolean isAsync, Boolean isTask, Type taskTResult, XmlName returnValueName, String defaultNS, String action, XmlName wrapperName, String wrapperNamespace, MessageDirection direction)
   at CoreWCF.Description.TypeLoader`1.CreateOperationDescription(ContractDescription contractDescription, MethodInfo methodInfo, MessageDirection direction, ContractReflectionInfo reflectionInfo, ContractDescription declaringContract)
   at CoreWCF.Description.TypeLoader`1.CreateOperationDescriptions(ContractDescription contractDescription, ContractReflectionInfo reflectionInfo, Type contractToGetMethodsFrom, ContractDescription declaringContract, MessageDirection direction)
   at CoreWCF.Description.TypeLoader`1.CreateContractDescription(ServiceContractAttribute contractAttr, Type contractType, ContractReflectionInfo& reflectionInfo, Object serviceImplementation)
   at CoreWCF.Description.TypeLoader`1.LoadContractDescriptionHelper(Type contractType, Object serviceImplementation)
   at CoreWCF.Description.TypeLoader`1.CreateContractDescription(ServiceContractAttribute contractAttr, Type contractType, ContractReflectionInfo& reflectionInfo, Object serviceImplementation)
   at CoreWCF.Description.TypeLoader`1.LoadContractDescriptionHelper(Type contractType, Object serviceImplementation)
   at CoreWCF.Description.TypeLoader`1.LoadContractDescription(Type contractType)
   at CoreWCF.Description.ContractDescription.GetContract[TService](Type contractType)
   at CoreWCF.ServiceHostObjectModel`1.CreateDescription(IDictionary`2& implementedContracts)
[...]

PR #825, opened 2022-09-08
Do not retrieve inherited MessageContract attributes

lukak-msft:
MessageContractHelper includes MessageContract attributes from base types, which leads to a Null Reference Exception later when CreateTypedMessageDescription attemptes to retrieve the same attribute from the message type.
- Do not retrieve inherited MessageContract attributes

#824

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/MessageContractHelper.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/MessageContractHelper.cs
@@ -12,7 +12,7 @@ namespace CoreWCF.Description
     {
         internal static bool IsMessageContract(Type type)
         {
-            foreach (Attribute attr in type.GetCustomAttributes())
+            foreach (Attribute attr in type.GetCustomAttributes(inherit: false))
             {
                 if (attr.GetType() == typeof(MessageContractAttribute)
                 || (string.Compare(attr.GetType().FullName, ServiceReflector.SMMessageContractAttributeFullName, true) == 0))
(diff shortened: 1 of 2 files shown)


mconnew:
The test failure was timing out. Apparently we have hit our limit of how many tests can reliably be run in 10 minutes, so we need to bump the timeout up. I'm going to rerun the failed tests to see if they pass quickly enough and we'll up the timeout time as a separate work item.

src/CoreWCF.Primitives/src/CoreWCF/Description/MessageContractHelper.cs
```diff
@@ -12,7 +12,7 @@ internal static class MessageContractHelper
     {
         internal static bool IsMessageContract(Type type)
         {
-            foreach (Attribute attr in type.GetCustomAttributes())
+            foreach (Attribute attr in type.GetCustomAttributes(inherit: false))
```
mconnew, line 15: Do you think there's any value before the call to return false to do the same call with `inherit: true` and if there's any matching attributes, throw an exception saying that MessageContractAttribute can't be applied to base types?

lukak-msft: I think that would be too aggressive behaviour here. That would make scenario of some type deriving from a chain of types which at some contains a MessageContractAttribute without having a MessageContractAttribute itself unsupported, and to me that would then imply that we expect that the type desires to use message contract if any of its base types have a MessageContractAttribute, which I am not sure is right, and would be a change of behaviour as compared to FX WCF.

mconnew: You convinced me, I agree.

PR #850, opened 2022-09-25
Introduce ASP.NET Core Authorization support

g7ed6e:


g7ed6e:
reverting to draft because i think the evaluation of the policy should occur sooner in the request pipeline.

src/CoreWCF.Http/src/CoreWCF/HttpClientCredentialType.cs
```diff
@@ -28,7 +29,8 @@ internal static bool IsDefined(HttpClientCredentialType value)
                 value == HttpClientCredentialType.Ntlm ||
                 value == HttpClientCredentialType.Windows ||
                 value == HttpClientCredentialType.Certificate ||
-                value == HttpClientCredentialType.InheritedFromHost);
+                value == HttpClientCredentialType.InheritedFromHost ||
+                value == HttpClientCredentialType.Custom);
```
mconnew, line 33: I don't think we need to introduce Custom, I think InheritedFromHost is sufficient here. In WCF that means whatever is configured in IIS for the application is what gets used. That's very similar in concept to whatever is configured in asp.net core is used.

src/CoreWCF.Http/src/CoreWCF/HttpTransportSecurity.cs
```diff
@@ -22,6 +22,8 @@ public HttpTransportSecurity()
             _extendedProtectionPolicy = ChannelBindingUtility.DefaultPolicy;
         }
 
+        public string CustomAuthenticationScheme { get; set; }
+
```
mconnew, line 26: I don't think we need to specify a scheme if we are just leaving it up to asp.net core. I had a look at the the authentication code and we can pass null as the scheme which means look up the default configured authentication scheme.

src/CoreWCF.Http/tests/Authorization/AuthorizationPoliciesTests.cs
```diff
@@ -0,0 +1,258 @@
...
+
+        [ServiceBehavior(IncludeExceptionDetailInFaults = true)]
+        public class EchoService : IEchoService
+        {
+            // No attribute => defaults to the builtin default policy which is RequireAuthenticatedUser
+            public string Default(string text) => text;
+            [Authorize(Policy = Policies.AdminOnly)]
+            public string AdminOnly(string text) => text;
+            [AllowAnonymous]
+            public string Anonymous(string text) => text;
```
mconnew, line 238: I don't think we can support the AllowAnonymous attribute. The granularity of specifying authentication type is at the service endpoint. If you have authentication required on the binding of an endpoint, it will require all requests for that endpoint to be authenticated. The way you allow some methods to be allowed to be called anonymously is you define those methods on a separate contract interface which only contains the anonymous methods and add another endpoint with the binding configured for anonymous authentication.  
We can't know which operation is being called at the transport level because the operation is selected using an IOperationSelector which is the second from last extensibility point used in the ImmutableDispatchRuntime (with IOperationInvoker being the last). There's just no way to know which operation is being called at the transport layer where authentication is being requested.  
I'll explain in a separate comment on the code where you are checking the properties on the returned message about the issues with that approach.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -640,6 +667,44 @@ internal async Task<MessageRpc> ProcessMessageAsync(MessageRpc rpc)
...
+        {
+            var httpContext = OperationContext.Current.IncomingMessageProperties["Microsoft.AspNetCore.Http.HttpContext"] as HttpContext;
+            string authenticationScheme = httpContext.Items["CoreWCF.Channels.HttpTransportSettings.CustomAuthenticationScheme"] as string;
+            var principal = httpContext.User;
+            AuthorizationOperationInvoker invoker = (AuthorizationOperationInvoker)rpc.Operation.Invoker;
+
+            if (!principal.Identity.IsAuthenticated)
+            {
+                if (!invoker.IsAllowAnonymous)
+                {
```
mconnew, line 690: There's a potential DoS problem with denying anonymous requests this late in the pipe. For a request to have got this far, there might have been a lot of resources allocated. For example, using WSHttpBinding you could have established a session context which would allocate quite a bit of bookkeeping objects which are expensive to create (creates a cryptographic random session key which is always expensive to create) as well as potentially instantiated a service instance (depending on instance context mode). If we allow anonymous requests through on an endpoint which is configured with authentication to support operations attributed as allowing anonymous users, a malicious attacker can cheaply send through a large number of anonymous requests which cause a lot of memory and CPU usage, potentially on a binding which is really expensive (like WSHttpBinding) and cause a service outage. A basic principal is do as little work as possible and allocate as few resources as possible until you have authenticated the user. If you want to allow anonymous users to make calls, then that's something you are explicitly opting in to and can evaluate the risk for the binding you are using. For example, allowing anonymous users when using a sessionless binding such as BasicHttpBinding, and using InstanceContextMode.Singleton, your exposure to a malicious client forcing resource allocation is minimal.  
Basically as I said in the other comment, anonymous operations are supported through a separate endpoint with an interface which only defines the methods allowed to be called anonymously.

g7ed6e: Supporting [AllowAnonymous] is indeed a bit challenging as we have to go deep into the pipeline to know which operation will be executed before rejecting the request. In a previous commit i was delaying the policy evaluation to the IOperationInvoker and then I realized we could be exposed to DoS attack. I tried to move the policy evaluation sooner in the pipeline but it's indeed still too late.

Do you think we should warn users about [AllowAnonymous] usages ?

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+            }
+        }
+
+        private static MethodInfo GetServiceImplMethodInfo(ServiceDescription serviceDescription, OperationDescription operationDescription)
+        {
+            Type serviceType = serviceDescription.ServiceType;
+            MethodInfo[] methodInfos = serviceType.GetMethods();
+            return methodInfos.First(x => x.Name == operationDescription.Name);
+        }
```
mconnew, line 63: This isn't going to work as the operationDescription.Name is often not the same as the method name. There's some mapping we do. For example BeginFoo -> Foo, and BarAsync -> Bar. On top of that, the name might come from OperationContractAttribute.Name where you can explicitly provide the name. There's an internal property on OperationDescription which will provide you the MethodInfo which will get called for this operation. It's called OperationMethod.

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            foreach (var endpoint in serviceDescription.Endpoints)
+            {
+                foreach (var operation in endpoint.Contract.Operations)
+                {
+                    MethodInfo methodInfo = GetServiceImplMethodInfo(serviceDescription, operation);
+                    bool hasAllowAnonymous = methodInfo.GetCustomAttribute(typeof(AllowAnonymousAttribute)) is not null;
+                    var authorizeData = methodInfo
+                        .GetCustomAttributes(typeof(AuthorizeAttribute), false).OfType<IAuthorizeData>()
+                        .ToArray();
```
mconnew, line 48: Do we only want to support the specific concrete type AuthorizeAttribute or do we want to support any attribute which implements IAuthorizeData? I think the latter would provide more flexibility and I don't see it being any extra cost for us at runtime.

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                    operation.OperationBehaviors.Add(
+                        new AuthorizationOperationBehavior(authorizeData, _authorizationPolicyProvider,
```
mconnew, line 51: Rather than adding a new operation behavior, I think we should also add a property on OperationDescription to hold the IAuthorizeData list. The list would be copied across in DispatcherBuilder.BuildDispatchOperation. A reasonable small extra cost is okay on startup for an unused feature, as long as there isn't an unnecessary cost at runtime. We should populate this data in TypeLoader.CreateOperationDescription. Doing it in a behavior means any reflected data may or may not be there before a users behavior executes as it would be order dependent.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -640,6 +667,44 @@ internal async Task<MessageRpc> ProcessMessageAsync(MessageRpc rpc)
...
+                ? (true, httpContext, authenticationSchemeValue)
+                : (false, null, null);
+
+        private async ValueTask<(bool isAuthorized, Message message)> EvaluateAuthorizationPolicies(MessageRpc rpc)
+        {
+            var httpContext = OperationContext.Current.IncomingMessageProperties["Microsoft.AspNetCore.Http.HttpContext"] as HttpContext;
+            string authenticationScheme = httpContext.Items["CoreWCF.Channels.HttpTransportSettings.CustomAuthenticationScheme"] as string;
+            var principal = httpContext.User;
+            AuthorizationOperationInvoker invoker = (AuthorizationOperationInvoker)rpc.Operation.Invoker;
+
```
mconnew, line 686: You can't depend on the operation invoker being a specific type. Just as you wrapped the existing invoker with your own, a developer might do the same. You are using the invoker as a means to hold properties about the operation, but that's the wrong place to store that. I had a look at the code that uses the asp.net core policy implementation and the same set of attributes are universally used by gRPC, WebAPI, MVC, and SignalR so I think it's a good idea to build in to the core code of CoreWCF knowledge about the concept of policies. I believe we should have policies also work for non-http transports too so we're not adding http specific implementation in primitives.   

I would make a public api on DispatchOperation which holds a collection of IAuthorizeData. This would allow attributing (which CoreWCF would populate), or a custom behavior which populates it more directly (for example, might not want to populate it the same for unit tests or dev environment).

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+    public class AuthorizationServiceBehavior : IServiceBehavior
+    {
+        private readonly IAuthorizationPolicyProvider _authorizationPolicyProvider;
+        private readonly IAuthorizationService _authorizationService;
+
+        public AuthorizationServiceBehavior(IAuthorizationPolicyProvider authorizationPolicyProvider,
+            IAuthorizationService authorizationService)
+        {
+            _authorizationPolicyProvider = authorizationPolicyProvider;
+            _authorizationService = authorizationService;
```
mconnew, line 23: I think we should store the IAuthorizationPolicyProvider on DispatchRuntime as that's where ServiceAuthenticationManager and ServiceAuthorizationManager live. DispatchRuntime has a RequiresAuthorization bool which controls whether an AuthorizationBehavior is returned from AuthorizationBehavior.TryCreate. The AuthorizationBehavior would store the IAuthorizationPolicyProvider as well as an IAuthorizationService. We can either add a method called AuthorizePolicyAsync which takes the principal and the AuthorizationPolicy as parameters and gets called in the same place as you are currently inserting the logic, or we could add parameters to the existing Authorize method (and make it async) so the one method can make decisions about both mechanisms. The existing Authorize code throws a FaultException with the the relevant namespace and faultcode etc which the client will understand and throw SecurityAccessDeniedException.

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
mconnew: Thinking about this more as I've explored the code, I think OperationDescription should have a collection of IAuthorizeData and DispatchOperation should store an AuthorizePolicy. OperationDescription holds object models to describe the operation whereas DispatchOperation holds concrete types with functionality so this seems like the right way to do things.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -640,6 +667,44 @@ internal async Task<MessageRpc> ProcessMessageAsync(MessageRpc rpc)
...
+                && authenticationScheme is string authenticationSchemeValue
+                && !string.IsNullOrEmpty(authenticationSchemeValue)
+                ? (true, httpContext, authenticationSchemeValue)
+                : (false, null, null);
+
+        private async ValueTask<(bool isAuthorized, Message message)> EvaluateAuthorizationPolicies(MessageRpc rpc)
+        {
+            var httpContext = OperationContext.Current.IncomingMessageProperties["Microsoft.AspNetCore.Http.HttpContext"] as HttpContext;
+            string authenticationScheme = httpContext.Items["CoreWCF.Channels.HttpTransportSettings.CustomAuthenticationScheme"] as string;
+            var principal = httpContext.User;
```
mconnew, line 684: These two lines will throw on non-http stacks. What we should do is make the principal available in the request message properties. Message.Properties.Security.ServiceSecurityContext contains a ServiceSecurityContext. This class has an AuthorizationContext which has a Properties class where you can add arbitrary named objects. If you look at ServiceSecurityContext.GetIdentities, you can see how we use this to store an `IList<IIdentity>` in the authorization context as AuthorizationContext doesn't directly support exposing IIdentity objects. We could to a similar thing to store the ClaimsPrincipal from HttpContext.User in the Http code. Then when we add later add support for non-http transports to use the authentication and policy mechanisms from asp.net core in a general purpose way, we can store the ClaimsPrincipal in the same way and this code will automatically light up.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
mconnew: Do you think we should emit a warning log or should we throw and refuse to startup if it's used? I worry about discoverability if it's a warning, but it's also harmless in that they won't accidentally process an unauthenticated request if they miss the warning.

src/CoreWCF.Http/src/CoreWCF/HttpTransportSecurity.cs
g7ed6e: done

src/CoreWCF.Http/tests/Authorization/AuthorizationPoliciesTests.cs
g7ed6e: I dropped Anonymous support

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
g7ed6e: `OperationMethod` property is now used.

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
g7ed6e: I updated the logic to fetcth all attributes that implement `IAuthorizeData`

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
g7ed6e: `OperationDescription` now holds a `Collection<IAuthorizeData>` named `AuthorizeData`.
`DispatchOperation` now hols the `AuthorizationPolicy` in a property named `AuthorizationPolicy`.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: done. please see the comment above.

g7ed6e: done. please see the comment above.

src/CoreWCF.Primitives/src/CoreWCF/Security/Authorization/AuthorizationServiceBehavior.cs
g7ed6e: I figured that I no longer need the `IAuthorizationPolicyProvider` when handling requests. Instead i'm using it in `TypeLoader` when building `OperationDescription`.
The `IAuthorizationService` is populated into `AuthorizationBehavior` using `ServiceAuthorizationManager` then `DispatchRuntime` and finally `AuthorizationBehavior.TryCreate`.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: I dropped this code and created a property named `ClaimsPrincipal` on `ServiceSecurityContext`. This property returns the `ClaimsPrincipal` from `HttpContext.User` that has been previsously stored in `AuthorizationContext.Properties`.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
samsp-msft: Did this get resolved by dropping the Anonymous support?

samsp-msft:
Sweet, I can't wait to update the sample to use this to enforce the authorization rules.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: Yes. Non authenticated requests are now dropped earlier in the request pipeline.

g7ed6e: Regarding warning we may add a roslyn analyzer with code fix to detect:
- `[AllowAnonymous]` usages at operation contract level
- `[Authorize]` at service contract level

samsp-msft:
@HaoK - can you take a look at this to make sure it fits in with the patterns that we want from an ASP.NET Core perspective.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: Added to https://github.com/CoreWCF/CoreWCF/pull/850/commits/ce7e91b13a1204ae1565172975accb7386256fa6 to detect above usages

g7ed6e:
@mconnew there's still an issue in this PR if someone mix the `OperationContractParameterInjector` feature from the source generator with authorization attributes. I can see 2 solutions:
- Either the source gen has to copy the attributes from the user provided implementation... but we will need to deal with possible namespace imports etc used by the `AuthorizeAttribute`. I can think to an authorization policy name constant defined in another namespace for example.
- Or i need to adapt the way that we populate the `AuthorizeData` property of an `OperationDescription` to get them from the user defined operation contract implementation.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -590,6 +593,16 @@ private static void BuildDispatchOperation(OperationDescription operation, Dispa
 
                 parent.UnhandledDispatchOperation = child;
             }
+
+            var authorizationPolicyProvider = services.GetService<IAuthorizationPolicyProvider>();
```
HaoK, line 597: We ended up adding some perf tweaks to enable caching when the IAuthorizationPolicyProvider is the Default implementation, because we know this always returns the same policies, which allows caching of the combined policies.  I'm not sure if this code is called more than once, but food for thought.  For 7.0 we ended up sticking the combined policy into the endpoint metadata itself. PR for context: https://github.com/dotnet/aspnetcore/pull/43124

src/CoreWCF.Http/tests/Authorization/AuthorizationStartup.cs
```diff
@@ -0,0 +1,109 @@
...
+    public Task<string> Generated(string text) => Task.FromResult(text);
+}
+
+[ServiceBehavior(IncludeExceptionDetailInFaults = true)]
+public partial class SecuredService : ISecuredService
+{
+    // No attribute => defaults to the builtin default policy which is RequireAuthenticatedUser
+    public string Default(string text) => text;
+
+    [Authorize(Policy = Policies.AdminOnly)]
```
HaoK, line 88: So AuthZ policies also allow specifying the authentication schemes as part of the policy, I don't see any code or tests for that, so I assume this is something not supported for WCF + Authorize?

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
g7ed6e: This code should only be called once at application startup, the combined policy is stored in the `DispatchOperation`.

src/CoreWCF.Http/tests/Authorization/AuthorizationStartup.cs
g7ed6e: Indeed that's no currently supported. you can see [there](https://github.com/CoreWCF/CoreWCF/pull/850/files#diff-b51abb1fc1d805a3936e07b10b0bad6bb889566195f11573ebcc8103aa1a565e) that we are explicitly passing null to `AuthenticateAsync`.

g7ed6e: We will now trigger build warnings when AuthenticationSchemes or Roles property are set

src/CoreWCF.BuildTools/src/DiagnosticDescriptors.cs
```diff
@@ -15,5 +15,41 @@ internal static class DiagnosticDescriptors
...
+            messageFormat: "[AllowAnonymous] attribute is not supported",
+            category: nameof(AuthorizationAttributesAnalyzer),
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true);
+
+        internal static Diagnostic AllowAnonymousAttributeIsNotSupportedWarning(Location location) => Diagnostic.Create(AllowAnonymousAttributeIsNotSupported, location);
+
+        internal static readonly DiagnosticDescriptor AuthorizeAttributeIsNotSupportedOnClass = new DiagnosticDescriptor(id: "COREWCF_0201",
+            title: "[Authorize] attribute is not supported",
+            messageFormat: "[Authorize] attribute is not supported on class '{0}'",
```
mconnew, line 30: Should we clarify in the message that it's not supported on classes? This message sounds more generally like it's not supported by CoreWCF at all.

src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs
```diff
@@ -136,9 +137,20 @@ internal TransportBindingElement GetTransport()
...
 
             BasicHttpSecurity basicHttpSecurity = BasicHttpSecurity;
+
             if (basicHttpSecurity.Mode == BasicHttpSecurityMode.Message)
             {
-                throw new PlatformNotSupportedException(nameof(BasicHttpSecurityMode.TransportWithMessageCredential));
+                throw new PlatformNotSupportedException(nameof(BasicHttpSecurityMode.Message));
+            }
+            else if (basicHttpSecurity.Mode == BasicHttpSecurityMode.None && basicHttpSecurity.Transport.ClientCredentialType == HttpClientCredentialType.InheritedFromHost)
+            {
```
mconnew, line 146: What's the purpose of this change? I have a suspicion the scenario here that you were looking to affect was with BasicHttpSecurityMode.TransportCredentialOnly

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -425,22 +434,30 @@ protected override Task OnCloseAsync(CancellationToken token)
...
                 {
                     WindowsSecurityTokenAuthenticator tokenAuthenticator = new WindowsSecurityTokenAuthenticator();
                     SecurityToken genericToken = new GenericIdentitySecurityToken(gid, SecurityUniqueId.Create().Value);
                     ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = await tokenAuthenticator.ValidateTokenAsync(genericToken);
                     return new ServiceSecurityContext(authorizationPolicies);
                 }
+                else if (principal.Identity is ClaimsIdentity cid)
+                {
+                    Fx.AssertAndThrow(cid.IsAuthenticated, "ClaimsPrincipal should be authenticated");
+                    AuthorizationContext authorizationContext = AuthorizationContext.CreateDefaultAuthorizationContext(null);
```
mconnew, line 456: Can this assert be hit? ValidateAuthentication already checks IsAuthenticated so if that returns Unauthorized, this code won't be hit.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -222,7 +224,7 @@ internal static void InitializeServiceHost(ServiceHostBase serviceHost, IService
                     ServiceEndpoint endpoint = stuff.Value.Endpoints[i];
 
                     //EndpointFilterProvider provider = new EndpointFilterProvider();
-                    EndpointDispatcher dispatcher = BuildEndpointDispatcher(description, endpoint);
+                    EndpointDispatcher dispatcher = BuildEndpointDispatcher(description, endpoint, services);
```
mconnew, line 228: ServiceDescription holds a reference to IServiceProvider so no need to add a parameter to pass it through.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -1117,9 +1119,73 @@ private void VerifyConsistency(OperationConsistencyVerifier verifier)
...
             return operationDescription;
         }
 
+        private static MethodInfo FindServiceImplementationMethodInfo(OperationDescription operationDescription, ContractDescription contractDescription)
+        {
+            Type serviceType = typeof(TService);
+            MethodInfo[] methodInfos = serviceType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+
+            // The source generator will produce an method with identical name but with attrributes marked with either CoreWCF.Injected or FromServices attributes.
+            // We need to to fetch AuthorizeAttribute from user provided implementation.
```
mconnew, line 1148: I think we should have the source generator copy all  attributes across to the generated method. We have the other existing attribute for AuthorizeRoleAttribute to provide equivalent behavior to PrincipalPermssionAttribute in NetFx. Doing this in the generator feels like the right way to do this too.

src/CoreWCF.Primitives/src/CoreWCF/ServiceSecurityContext.cs
```diff
@@ -159,6 +160,12 @@ public AuthorizationContext AuthorizationContext
             }
         }
 
+        public ClaimsPrincipal ClaimsPrincipal => _authorizationContext != null
+                   && _authorizationContext.Properties.TryGetValue("ClaimsPrincipal", out object principal)
```
mconnew, line 164: We already have a ClaimsPrincipal property on OperationContext. We aren't currently setting it as the code in `DispatchOperationRuntime.SetClaimsPrincipalToOperationContext` hasn't been re-enabled yet.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/AuthorizationBehavior.cs
```diff
@@ -17,12 +19,21 @@ internal sealed class AuthorizationBehavior
...
         private ServiceAuthorizationManager _serviceAuthorizationManager;
+        private IAuthorizationService _authorizationService;
 
         private AuthorizationBehavior() { }
 
         public async ValueTask<MessageRpc> AuthorizeAsync(MessageRpc rpc)
         {
             // TODO: Events
+            ClaimsPrincipal claimsPrincipal = rpc.Request.Properties.Security?.ServiceSecurityContext?.ClaimsPrincipal;
+            AuthorizationPolicy authorizationPolicy = rpc.Operation.AuthorizationPolicy;
```
mconnew, line 30: We should move the call to `DispatchOperationRuntime.SetClaimsPrincipalToOperationContext` to happen either as part of this method, or before this method is called. Then you can get the ClaimsPrincipal from rpc.OperationContext.ClaimsPrincipal

g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/0da8aca434f3bcb07d3fd55955a69d39cb3cd352

src/CoreWCF.Primitives/src/CoreWCF/ServiceSecurityContext.cs
g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/0da8aca434f3bcb07d3fd55955a69d39cb3cd352

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/4961f2aef7d575f7549fb95bb44bc60147747326

src/CoreWCF.BuildTools/src/AnalyzerReleases.Unshipped.md
```diff
@@ -1 +1,8 @@
-﻿
\ No newline at end of file
+﻿### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+COREWCF_0200 | AuthorizationAttributesAnalyzer | Warning  | COREWCF_0200_DiagnosticDescriptors
+COREWCF_0201 | AuthorizationAttributesAnalyzer | Error    | COREWCF_0201_DiagnosticDescriptors
```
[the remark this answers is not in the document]
g7ed6e: I believe best documentation is code thus i broke down the `DiagnosticDescriptors` class with inner classes which name contain the scope. Please see https://github.com/CoreWCF/CoreWCF/pull/850/commits/6e91a4a1c572a5b87de6be9fe79b9e488dd0e0a5

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/a2c8dbc237534bd8922cfcb7f2b2759694841cf7

src/CoreWCF.BuildTools/src/DiagnosticDescriptors.cs
g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/dc5c90ab00ffd1908065e943ee07f5530ee5b464

src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs
g7ed6e: Indeed, i reverted back this file and change the unit test configuration. Please see https://github.com/CoreWCF/CoreWCF/pull/850/commits/fc4c36ce39e3a40ae5fac57bdac6f308903ef775

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
g7ed6e: indeed i removed it here https://github.com/CoreWCF/CoreWCF/pull/850/commits/64ed1127e474fddb939539fc8bb780fa003ce018

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
```diff
@@ -425,22 +434,29 @@ protected override Task OnCloseAsync(CancellationToken token)
...
                 {
                     WindowsSecurityTokenAuthenticator tokenAuthenticator = new WindowsSecurityTokenAuthenticator();
                     SecurityToken genericToken = new GenericIdentitySecurityToken(gid, SecurityUniqueId.Create().Value);
                     ReadOnlyCollection<IAuthorizationPolicy> authorizationPolicies = await tokenAuthenticator.ValidateTokenAsync(genericToken);
                     return new ServiceSecurityContext(authorizationPolicies);
                 }
+                else if (principal.Identity is ClaimsIdentity)
+                {
+                    AuthorizationContext authorizationContext = AuthorizationContext.CreateDefaultAuthorizationContext(null);
+                    authorizationContext.Properties.Add("ClaimsPrincipal", principal);
```
g7ed6e, line 456: @mconnew Here the magic string `"ClaimsPrincipal"` refers to the constant `CoreWCF.IdentityModel.Tokens.AuthorizationPolicy.ClaimsPrincipalKey`.
Please do notice that this const is internal to CoreWCF.Primitives and is used to pull `ClaimsPrincipal` from `AuthorizationContext` in 2 scenarii:
- The scenario of this PR (using ASP.NET Core AuthZ)
- Another scenario that uses IdentityModel

src/CoreWCF.BuildTools/src/Extensions.cs
```diff
@@ -17,11 +54,11 @@ public static bool IsMatchingParameter(this IParameterSymbol symbol, IParameterS
 
     internal static class SymbolExtensions
     {
-        public static bool HasOneOfAttributes(this ISymbol symbol, params INamedTypeSymbol?[] attributeTypeSymbols)
+        public static (bool Value, AttributeData? AttributeData)  HasOneAttributeOf(this ISymbol symbol, params INamedTypeSymbol?[] attributeTypeSymbols)
         {
```
mconnew, line 58: This is a synchronous method so no need to use a named tuple to return multiple values. Just use an out parameter instead. Or you could just return AttributeData as false is only returned when this is null so you could equate returning null as being equivalent to Value being false.

src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs
mconnew: You did make one correct fix which was the nameof is the PlatformNotSupportedException is using the wrong enum value. Could you refix that?

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
mconnew: Although strictly speaking the principal might not be a ClaimsPrincipal, it likely is if the Identity is a ClaimsIdentity. What do you think of using `nameof(ClaimsPrincipal)` here?

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -1117,9 +1119,74 @@ private void VerifyConsistency(OperationConsistencyVerifier verifier)
...
+                {
+                    operationDescription.AuthorizeData.Add(methodAttribute);
+                }
+            }
+
             return operationDescription;
         }
 
+        private static MethodInfo FindServiceImplementationMethodInfo(OperationDescription operationDescription, ContractDescription contractDescription)
+        {
```
mconnew, line 1143: There's a lot easier way to do this using an [interface map](https://learn.microsoft.com/en-us/dotnet/api/system.type.getinterfacemap?view=net-6.0). I think code like this should work:
```c#
Type serviceType = typeof(TService);
var ifaceMap = serviceType.GetInterfaceMap(operationDescription.OperationMethod.DeclaringType);
var index = Array.IndexOf(ifaceMap.InterfaceMethod, operationDescription.OperationMethod);
if (index < 0) return null;
return ifaceMap.TargetMethods[index];
```

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -550,6 +550,7 @@ internal async Task<MessageRpc> ProcessMessageAsync(MessageRpc rpc)
                 rpc = await _authenticationBehavior.AuthenticateAsync(rpc);
             }
 
+            rpc.Operation.SetClaimsPrincipalToOperationContext(rpc);
             if (_authorizationBehavior != null)
```
mconnew, line 554: This should be wrapped in `if (Parent.RequireClaimsPrincipalOnOperationContext)` otherwise the call might throw. This means you need to make sure that whatever mechanism sets `RequireClaimsPrincipalOnOperationContext` to true is triggered in this scenario.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: It works but break 2 tests:
- one test named `InterfaceOnlyServiceTest` that register service in DI using interface and pull implementation using `ServiceProvider`.
- another one named `DuplexEcho`

g7ed6e: This seems to work
```c#
            Type serviceType = typeof(TService);
            if (serviceType.IsInterface || operationDescription.OperationMethod.DeclaringType!.IsInterface)
            {
                return operationDescription.OperationMethod;
            }

            InterfaceMapping interfaceMapping = serviceType.GetInterfaceMap(operationDescription.OperationMethod.DeclaringType);
            int index = Array.IndexOf(interfaceMapping.InterfaceMethods, operationDescription.OperationMethod);
            return index < 0
                ? null
                : interfaceMapping.TargetMethods[index];
```

src/CoreWCF.BuildTools/src/Extensions.cs
g7ed6e: done in https://github.com/CoreWCF/CoreWCF/pull/850/commits/dbfa96179c9880b7f43b9fbe8a8e0adf6fab0957

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/719999fb8b5771d6060ce4d2e59a307406415718

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/bc84c9202bbc8cb1df0ca6e430366e599afa4e35

src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/f2721caf8a5829ea7d741e151452afcb71501fdc

src/CoreWCF.Http/src/CoreWCF/Channels/HttpRequestContext.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/5ac13a79cedcd259ff806e710c6cd3e51640949c

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: Thanks i did not know `InterfaceMapping`!

mconnew: This means this feature won't work when CoreWCF doesn't know the type ahead of time.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
```diff
@@ -191,6 +212,12 @@ void IServiceBehavior.ApplyDispatchBehavior(ServiceDescription description, Serv
                         {
                             ApplyAuthorizationPoliciesAndManager(behavior);
                         }
+
+                        if (_isAuthorizationServiceSet)
+                        {
+                            behavior.AuthorizationService = _authorizationService;
+                            behavior.RequireClaimsPrincipalOnOperationContext = true;
+                        }
```
mconnew, line 220: Doesn't _isAuthorizationServiceSet get set to true if there's an IAuthorizationPolicyProvider available in DI? If they have it available but aren't using it for their CoreWCF service, won't this will cause an exception because the ClaimsPrincipal isn't available?

g7ed6e: Indeed. In case user added Authorization services in DI but does not use it fore CoreWCF we will have `RequireClaimsPrincipalOnOperationContext` set to true which is not correct, I will add a unit test to highlight it.
I still need to find a way to set `RequireClaimsPrincipalOnOperationContext` to true when `HttpClientCredentialType.InheritedFromHost` is configured in the binding.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -590,6 +593,16 @@ private static void BuildDispatchOperation(OperationDescription operation, Dispa
 
                 parent.UnhandledDispatchOperation = child;
             }
+
+            var authorizationPolicyProvider = services.GetService<IAuthorizationPolicyProvider>();
+            if (authorizationPolicyProvider != null)
+            {
+                // TODO: Make this chain call async
+                Task<AuthorizationPolicy> getPolicyTask = operation.AuthorizeData.Count > 0
+                    ? AuthorizationPolicy.CombineAsync(authorizationPolicyProvider, operation.AuthorizeData)
```
g7ed6e, line 602: The `CombineAsync` call need a proper unit test that actually combine several policies set through multiple `AuthorizeAttribute`

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: When the "Fully register CoreWCF service with interface" has been added it assumed the concrete service type will be in DI. What do you think about trying to pull an instance from DI using a ServiceScope, getting the concrete type and pull the attributes applied to it ?

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -590,6 +593,16 @@ private static void BuildDispatchOperation(OperationDescription operation, Dispa
...
+                    : authorizationPolicyProvider.GetDefaultPolicyAsync();
```
g7ed6e, line 603: @mconnew please notice that ASP.NET Core 3 introduced a `FallbackPolicy` which differs of the DefaultPolicy.
- `DefaultPolicy` is applied when `AuthorizeData` is applied without specifying a policy name.
- `FallbackPolicy` is applied when no Authorize attribute is applied.

The behavior implemented here is that we apply the `DefaultPolicy` when no `AuthorizeData` is provided.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
g7ed6e: fixed in https://github.com/CoreWCF/CoreWCF/pull/850/commits/8da1ca6e9f916bf3eb6a64f0fccdd56c859fc84d

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
g7ed6e: This is now done

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: @mconnew I brought support of Authorize at ServiceContract level and dropped the analyzer that was emitting build errors

src/CoreWCF.Primitives/src/CoreWCF/Channels/IAuthorizationCapabilities.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Channels;
+
+public interface IAuthorizationCapabilities
```
g7ed6e, line 6: Make this interface and its implementation internal and move to shared code folder

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
```diff
@@ -222,6 +224,27 @@ public override T GetProperty<T>(BindingContext context)
             {
                 return (T)(object)new HttpTransportServiceBuilder();
             }
+
+            if (typeof(T) == typeof(IAuthorizationCapabilities))
+            {
+                var binding = context.BindingParameters.OfType<Binding>().FirstOrDefault();
+                if (binding is HttpBindingBase httpBindingBase)
+                {
```
mconnew, line 232: Also need code for WSHttpBinding (which has an WSHttpBindingBase base class which I think contains the equivalent of HttpBindingBase)

src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
```diff
@@ -96,6 +96,7 @@ protected override ServiceDescription CreateDescription(out IDictionary<string,
 
             ReflectedContractCollection reflectedContracts = new ReflectedContractCollection();
             List<Type> interfaces = ServiceReflector.GetInterfaces<TService>();
+            TypeLoader<TService>.ServiceProvider = _serviceProvider;
             for (int i = 0; i < interfaces.Count; i++)
```
mconnew, line 100: This is fragile and can cause problems in a rare scenario. With a single WebHost being built, this is fine, but if you had 2 different threads creating 2 different WebHost instances for the same service type at the same time, you could get the service provider leak between WebHost build pipelines and end up with some really difficult to diagnose bugs.

src/CoreWCF.Primitives/src/CoreWCF/Channels/DefaultAuthorizationCapabilities.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Channels;
+
+public class DefaultAuthorizationCapabilities : IAuthorizationCapabilities
+{
```
mconnew, line 7: Do we want this class public? The general pattern is for the interface to be public, and the default implementation to be internal. Take a look at ISecurityCapabilities and SecurityCapabilities as an example. If it's internal, you don't need a constructor, a setter will work.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
```diff
@@ -154,8 +176,17 @@ private void CopyAuthorizationPoliciesAndManager(ServiceAuthorizationBehavior ot
...
         }
 
-        void IServiceBehavior.Validate(ServiceDescription description, ServiceHostBase serviceHostBase)
+        void IServiceBehavior.Validate(ServiceDescription serviceDescription, ServiceHostBase serviceHostBase)
         {
+            foreach (ServiceEndpoint endpoint in serviceDescription.Endpoints)
+            {
+                TransportBindingElement transportBindingElement = endpoint.Binding.CreateBindingElements().Find<TransportBindingElement>();
+                if (transportBindingElement != null)
+                {
```
mconnew, line 185: A binding is required to have a TransportBindingElement. I just looked up the code to link it and discovered we have a bug. ServiceEndpoint.EnsureInvariants is supposed to ball Binding.EnsureInvariants but that's been commented out. Subject to that bug being fixed (should just require uncommenting the one line) you should be able to presume transportBindingElement is not null. I can see an Fx.Assert being useful here still, to catch bugs.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
```diff
@@ -1117,9 +1122,41 @@ private void VerifyConsistency(OperationConsistencyVerifier verifier)
...
             return operationDescription;
         }
 
+        private static IEnumerable<IAuthorizeData> GetAuthorizeData(OperationDescription operationDescription)
+        {
+            Type serviceType = typeof(TService);
+            if (serviceType.IsInterface)
+            {
+                using var scope = ServiceProvider.CreateScope();
+                object serviceInstance = scope.ServiceProvider.GetService(serviceType);
```
mconnew, line 1140: This can still breaks some users. There's one user that I made a fix for where their service type has an IHttpContextAccessor constructor parameter and their app breaks when it can't get one. I don't think this will break them specifically as I think they specify a concrete type to AddService, but if they switched to the interface this would break them.

mconnew: I think I'm okay with saying this feature doesn't work when service type is an interface as long as we can detect that they are using an interface and IAuthorizationCapabilities.SupportsAuthorizationData returns true and either fail startup or output a warning. I'm okay with failing startup and rolling back to a warning if that's a problem as going the other direction is worse developer experience.

g7ed6e: This is ran lazy so it should work. I wrote a specific unit test to check HttpContextAccessor.HttpContext is not null. please see https://github.com/CoreWCF/CoreWCF/pull/850/commits/bd1fb5fdad760909bd22a5d62b4b8e63bdfb082e

src/CoreWCF.Primitives/src/CoreWCF/Channels/DefaultAuthorizationCapabilities.cs
g7ed6e: indeed i renamed the class too to ensure I follow the general pattern you mentioned https://github.com/CoreWCF/CoreWCF/pull/850/commits/830320d1af7dd6c019176c06aab0a7cc345fefe4

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
```diff
@@ -222,6 +224,27 @@ public override T GetProperty<T>(BindingContext context)
...
+                    context.BindingParameters.Remove(httpBindingBase);
+                    return (T)(object)new DefaultAuthorizationCapabilities(httpBindingBase?.BasicHttpSecurity.Transport.ClientCredentialType ==
+                                                                           HttpClientCredentialType.InheritedFromHost);
+                }
+                if (binding?.GetType().FullName == "CoreWCF.WebHttpBinding")
+                {
+                    context.BindingParameters.Remove(binding);
+                    dynamic webHttpBinding = binding;
```
mconnew, line 240: >dynamic webHttpBinding = binding; [](http://example.com/codeflow?start=20&length=33)

 It's been a while since I deep a deep dive on how the dynamic keyword works and I forgot you could do things like that. If you look at the generated IL it's an expensive operation with a lot of stuff behind that one keyword so it should be used sparingly. We might need to remove it in the future too and find another way to do this (add a property on HttpTransportBindingElement?) as asp.net core is working to add support for aot compilation and we might want to follow. I have a suspicion this will break aot compilation. It's okay for now, just saying we need to minimize usage overall.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
```diff
@@ -222,6 +224,27 @@ public override T GetProperty<T>(BindingContext context)
             {
                 return (T)(object)new HttpTransportServiceBuilder();
             }
+
+            if (typeof(T) == typeof(IAuthorizationCapabilities))
+            {
+                var binding = context.BindingParameters.OfType<Binding>().FirstOrDefault();
+                if (binding is HttpBindingBase httpBindingBase)
```
mconnew, line 231: OfType is a Linq method. BindingParameters is of type BindingParameterCollection which derives from KeyedByTypeCollection\<object\>. You should use Find<Binding>() instead, which returns default if it can't be found. You avoid intermediate IEnumerable instance being created.

src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
g7ed6e: I now store the `ServiceProvider` in `ContractDescription`. That's not an issue to store it after object creation as it is used lazyly when first request comes in to populate the lazy AuthorizeData property. This allow us not to break the public API `ContractDescription.GetContract`. please see https://github.com/CoreWCF/CoreWCF/pull/850/commits/d9e9dac4bf5685df166bf9d85fceb3a4602a2a34

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
mconnew: Okay, I see that. It's depending on a statically set ServiceProvider which as I mentioned in another comment is fragile and will break. If we had concurrent execution of unit tests (on my todo that we switch to dynamic port usage for all tests so we can) this would break.

mconnew: It also still doesn't use the concrete service type from the InstanceContext which can theoretically change from call to call.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/87b5c264ab6c606dcabe988487ff9966b939f639

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/abc73112939c4622216588547e548eaddc68d36e

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
mconnew: @g7ed6e, ping in case you missed this comment.

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
g7ed6e: Totally agree. I will create a separate issue to remediate this.

src/CoreWCF.Primitives/src/CoreWCF/Description/TypeLoader.cs
g7ed6e: @mconnew i pushed https://github.com/CoreWCF/CoreWCF/pull/850/commits/d7481f0c2524680134701bcb7241ddd374e95067. It now relies on InstanceContext instead of ServiceProvider. An AuthorizeData collection is cached per type into `OperationDescription`.

src/CoreWCF.Primitives/src/CoreWCF/ServiceHostObjectModel.cs
g7ed6e: I finally dropped this code

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
g7ed6e: @mconnew  Please see here https://github.com/CoreWCF/CoreWCF/pull/850/commits/36f9062d236c9a8f528355d9993929bf219e9568. Implemented at WSHttpBinding level which also exposes a Security property that allow HttpClientCredentialType configuration.

src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs
```diff
@@ -599,9 +600,25 @@ private static void BuildProxyOperation(OperationDescription operation, ClientRu
             {
                 child.AuthorizationPolicy = new Lazy<AuthorizationPolicy>(() =>
                 {
+                    object serviceInstance = OperationContext.Current.InstanceContext.GetServiceInstance();
```
mconnew, line 603: This code has a few unintended consequences for some scenarios.

As you've noticed, the call to InstanceBehavior.EnsureInstanceContextAsync happens after the call to authorizationBehavior.AuthorizeAsync so you moved it earlier. There's another call that needs to also happen earlier. Take a look at InstanceBehavior.EnsureServiceInstance. It has an optional capability of releasing an existing service instance before creating a new one. If that's the case, you could double allocate with this change. If some work is done in the constructor of the instance, this could have some bad side effects. We need to use the call to EnsureServiceInstance to allocate the instance and only call it once. InstanceContext.GetInstance will then return a cached value so won't double allocate. This would fix the next issue I discovered which I'm leaving here to increase your understanding but will be resolved by fixing this.  

If a custom IInstanceProvider is being used, this results in GetInstance being called without the request message. The built-in implementation won't have a change in behavior, but a custom implementation might. We need to pass the request message, which you can get from  OperationContext.RequestContext.RequestMessage. InstanceContext.GetServiceInstance caches the instance so if we use EnsureServiceInstance before this code, this becomes a non-issue as it will already have been done correctly.   

I was initially thinking we should optionally call EnsureInstanceContextAsync and EnsureServiceInstance earlier when enabling this feature but then I took a look at what else happens between authorizationBehavior.AuthorizeAsync and EnsureServiceInstance and there are just too many potential changes in behavior. One example is we allocate the service instance while running within a captured synchronization context. This could be important if running in a UI app like WPF of WinForms and the instance constructor accesses UI elements. It also happens after any concurrency locks are acquired. This means we could now allocate more instances than intended which could have a knock on effect elsewhere (more connections to backend systems for example). I've been working with a customer on WCF recently that was using some of these extensibility points which moving the instance creation earlier would have broken badly.

I now think we need to invert the thinking and instead, only if policies are being used move the authorization to later in the pipeline after the call to EnsureServiceInstance. If policies aren't being used, then we leave it where it is.  

This will need to be mentioned in your blog post. There are some things which will now happen before authorization which didn't before. 
1. If using a dynamic quota throttle, it will be acquired before authorization
2. IDispatchMessageInspector.AfterReceiveRequest will be called
3. Concurrency lock will be acquired
4. Authorization will now run on a captured SynchronizationContext

I don't think this is a big problem as the old Windows Permissions attribute for requiring the caller to be in a particular Windows domain group for example was done by the CLR just before calling the operation method which is as late as you can get and that was an authorization mechanism.  

Can you open an issue to move the authorization step to permanently be later for the 2.0 version with an AppContext switch to revert the behavior. The plan would be to remove the AppContext switch in 3.0 if it doesn't cause problems.

src/CoreWCF.Primitives/src/CoreWCF/Description/OperationDescription.cs
```diff
@@ -126,7 +127,7 @@ public string Name
 
         internal KeyedByTypeCollection<IAuthorizeOperation> AuthorizeOperation { get; }
 
-        internal Lazy<ReadOnlyCollection<IAuthorizeData>> AuthorizeData { get; set; }
+        internal ConcurrentDictionary<Type, ReadOnlyCollection<IAuthorizeData>> AuthorizeData { get; set; }
```
mconnew, line 130: ConcurrentDictionary is optimized for multiple concurrent readers and writers and has some performance issues. For example it has buckets which are affinitized a thread (or multiple threads) to reduce contention, but then when reading it needs to check multiple buckets. There's a little known pattern you can do when you have a seldom write read often situation like we have here. The HashTable class has an interesting property that it can safely return results from multiple readers at once concurrent to a write, as long as you only ever have a single write happening. In other words, reads are always thread safe and a single write is thread safe with multiple reads. The failure mode with concurrent read and write to an entry being modified (value added, changed, or deleted) is the read might return a stale value, but it won't ever throw or return an incorrect value (value from different key). I was going to quickly code up a helper class for this and went looking at the implementation and today was one of those days when I learn something new :smile:. .NET has had a s[tatic method to return a wrapped implementation](https://learn.microsoft.com/en-us/dotnet/api/system.collections.hashtable.synchronized?view=net-7.0) which only locks on write since .NET Framework 1.1. You would create a new instance with `Hashtable.Synchronized(new Hashtable());`. You can either use this directly and add cases into the code where necessary, or I [quickly threw together a generic implementation](https://gist.github.com/mconnew/772049ac7e202d8e7468e5df74ecfb93) of `IDictionary<Tkey,TValue>` based on this which I did some basic testing on to validate. You can see a couple of methods still need to take a lock so avoid those if you choose to use this.

mconnew:
The latest change doesn't run authorizationBehavior.AuthorizeAsync at all if you are using authorization attributes. Is that the intended behavior? Now you've split the functionality into a different call, I think we should still call AuthorizeAsync like normal at the original place in the code.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -550,7 +553,7 @@ internal async Task<MessageRpc> ProcessMessageAsync(MessageRpc rpc)
                 rpc = await _authenticationBehavior.AuthenticateAsync(rpc);
             }
 
-            if (_authorizationBehavior != null)
+            if (_authorizationBehavior != null && !SupportsAuthorizationData)
             {
```
mconnew, line 557: If you remove the !SupportsAuthorizationData so that the original behavior still remains, I'd call this feature done and you can merge.

mconnew (approved):
Just the one small change and then you can merge.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
g7ed6e: @mconnew I removed the `!SupportsAuthorizationData` flag check but it had a side effect as `AuthorizationBehavior.AuthorizeAsync` internally set the ServiceSecurityContext which lead to the lost of the ClaimsPrincipal we previously stored during authentication. 
I then moved the call to `SetClaimsPrincipalToOperationContext` before the original call to `AuthorizationBehavior.AuthorizeAsync`.

g7ed6e:
> Just the one small change and then you can merge.

@mconnew Having a look back to all discussions i figured out that i forgot this https://github.com/CoreWCF/CoreWCF/pull/850#discussion_r1008816833. That's now done here https://github.com/CoreWCF/CoreWCF/pull/850/commits/795b5a9e6753020e2d724657687ed76c00ff6998

src/CoreWCF.Http/src/CoreWCF/Channels/HttpTransportBindingElement.cs
g7ed6e: finally done in this [commit](https://github.com/CoreWCF/CoreWCF/pull/850/commits/3c2d4b92aee4bc625b4275da639172a0678d661d)

src/CoreWCF.Http/tests/Authorization/AuthorizationTests.NotSupported.cs
```diff
@@ -0,0 +1,198 @@
...
+
+    [Theory]
+    [MemberData(nameof(Get_Authorization_Features_Are_Mutually_Exclusive_TestVariations))]
+    public void Authorization_Features_Are_Mutually_Exclusive_Test(Type startupType)
+    {
+        using IWebHost host = ServiceHelper.CreateWebHostBuilder(_output, startupType).Build();
+        var exception = Assert.Throws<NotSupportedException>(() =>
+        {
+            host.Start();
+        });
```
mconnew, line 47: This will leak an open host if there's a bug which stops it from throwing. Do this to ensure that doesn't happen:
```c#
using(host)
{
    host.Start();
}
```
This will guarantee the host is disposed regardless of the outcome of Start().

src/CoreWCF.Primitives/src/Resources/Strings.resx
```diff
@@ -9726,4 +9726,7 @@
   <data name="XmlWriterBackedStreamPropertySetNotSupported" xml:space="preserve">
     <value>The set accessor of the '{0}' property is not supported by the stream.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="AuthorizationFeaturesAreMutuallyExclusive" xml:space="preserve">
+    <value>Invalid configuration. The security of the Binding is configured with ClientCredentialType set to 'InheritedFromHost' which imply that only native ASP.NET Core Authorization will be ran. Applying a custom 'ServiceAuthorizationManager' and/or setting the 'ServiceAuthorizationBehavior.ExternalAuthorizationPolicies' property is not supported. Please see https://learn.microsoft.com/en-us/aspnet/core/security/authorization to configure ASP.NET Core Authorization.</value>
```
mconnew, line 9730: Typo: ...which implies that....

PR #851, opened 2022-09-26
Make ServiceAuthenticationManager async

g7ed6e:


mconnew:
Same issue as the other PR for ServiceAuthorizationManager, the default implementation needs to stay in the sync methods and have the async methods call them.

g7ed6e:
> Same issue as the other PR for ServiceAuthorizationManager, the default implementation needs to stay in the sync methods and have the async methods call them.

I inverted the delegation to ensure compatibility with existing implementations.

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthenticationManager.cs
```diff
@@ -45,7 +53,7 @@ public override ReadOnlyCollection<IAuthorizationPolicy> Authenticate(ReadOnlyCo
                 authPolicy = authPolicies.AsReadOnly();
             }
 
-            return _wrappedAuthenticationManager.Authenticate(authPolicy, listenUri, ref message);
+            return await _wrappedAuthenticationManager.AuthenticateAsync(authPolicy, listenUri, message);
         }
```
mconnew, line 57: When the only await is the last line of the method, you don't need to make the method async. You can directly return the ValueTask from _wrappedAuthenticatonManager

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthenticationManager.cs
```diff
@@ -84,18 +92,18 @@ public override ReadOnlyCollection<IAuthorizationPolicy> Authenticate(ReadOnlyCo
...
                     {
-                        return authPolicy;
+                        return (authPolicy, message);
                     }
                 }
             }
 
-            return _wrappedAuthenticationManager.Authenticate(authPolicy, listenUri, ref message);
+            return await _wrappedAuthenticationManager.AuthenticateAsync(authPolicy, listenUri, message);
         }
```
mconnew, line 101: Same as above with not needing to use await

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthenticationManager.cs
g7ed6e: i removed the async /await

mconnew (approved):

PR #852, opened 2022-09-27
Make ServiceAuthorizationManager async

g7ed6e:


mconnew:
You've got the delegation the wrong way around between the async and sync methods. Think about the scenario where someone has an implementation where they already have a derived ServiceAuthorizationManager with an implementation of any of the sync methods. CoreWCF will now be calling the async methods so that a provided async implementation is called, but if the async method hasn't been overridden, we still need to end up calling the overridden sync method. Additionally, if a sync method has been overridden, the implementation needs to be able to call the base method to execute the default behavior.  
This means the default behavior needs to remain in the sync method (which is fine as the default behavior is still a synchronous implementation), and the async methods need to call the sync methods. This has the added benefit of completely avoiding any issues of synchronous code calling an async method and waiting on the result.

g7ed6e:
> You've got the delegation the wrong way around between the async and sync methods. Think about the scenario where someone has an implementation where they already have a derived ServiceAuthorizationManager with an implementation of any of the sync methods. CoreWCF will now be calling the async methods so that a provided async implementation is called, but if the async method hasn't been overridden, we still need to end up calling the overridden sync method. Additionally, if a sync method has been overridden, the implementation needs to be able to call the base method to execute the default behavior. This means the default behavior needs to remain in the sync method (which is fine as the default behavior is still a synchronous implementation), and the async methods need to call the sync methods. This has the added benefit of completely avoiding any issues of synchronous code calling an async method and waiting on the result.

I inverted the delegation to make all async virtual methods call their sync couterpart. 
Also I changed implementation of sync methods to call async methods when they were previously calling an overload or a XXCore method

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
```diff
@@ -11,18 +15,52 @@ namespace CoreWCF
...
+            Type implementorType = GetType();
+            var methods = implementorType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+
+            var checkAccessCoreAsyncMethodInfo = methods.Single(static x => x.Name == nameof(CheckAccessCoreAsync));
+            var baseCheckAccessCoreAsyncMethodInfo = checkAccessCoreAsyncMethodInfo!.GetBaseDefinition();
+
+            _isCheckAccessCoreAsyncOverridden = baseCheckAccessCoreAsyncMethodInfo.DeclaringType != checkAccessCoreAsyncMethodInfo.DeclaringType;
+
+            var checkAccessAsyncWithSingleParameterMethodInfo = methods.SingleOrDefault(static x =>
+                x.Name == nameof(CheckAccessAsync) && x.GetParameters().Length == 1);
```
mconnew, line 32: If the implementation has another method with the same name and a single parameter, eg a helper method with the same name, how will this behave?

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
```diff
@@ -39,13 +77,33 @@ public virtual bool CheckAccess(OperationContext operationContext)
             operationContext.IncomingMessageProperties.Security.ServiceSecurityContext =
                 new ServiceSecurityContext(authorizationPolicies ?? EmptyReadOnlyCollection<IAuthorizationPolicy>.Instance);
 
-            // 3) Call the CheckAccessCore
+            // 3) Call the CheckAccessCoreAsync(OperationContext operationContext)
             return CheckAccessCore(operationContext);
```
mconnew, line 81: I think this comment should say CheckAccessCore

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
```diff
@@ -11,18 +15,52 @@ namespace CoreWCF
...
         // 4) Return the authoritive decision true/false (allow/deny).
-        public virtual bool CheckAccess(OperationContext operationContext, ref Message message)
+        [Obsolete("Implementers should override CheckAccessAsync.")]
+        public virtual bool CheckAccess(OperationContext operationContext, ref Message message) => CheckAccess(operationContext);
+
+        public virtual async ValueTask<(bool isAuthorized, Message message)> CheckAccessAsync(OperationContext operationContext, Message message)
         {
-            return CheckAccess(operationContext);
+            if (_isCheckAccessAsyncWithSingleParameterOverridden || _isCheckAccessCoreAsyncOverridden)
+            {
```
mconnew, line 53: Should we combine these two into a single field with the two values combined already? I couldn't see anywhere that either is used independently.

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
g7ed6e: I refined the logic of both flags

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
g7ed6e: Good catch! thanks!

src/CoreWCF.Primitives/src/CoreWCF/ServiceAuthorizationManager.cs
g7ed6e: I merged both boolean values in a single field

mconnew (approved):

PR #856, opened 2022-09-28
Fix NetTcp MaxReceivedMessageSize

### LINKED ISSUE #843: MaxReceivedMessageSize is exceeded when transferring a large file with NetTcpBinding
I got QuotaExceededException when trying to transfer a large file(~3GB) from client to corewcf server. The exception complains the MaxReceivedMessageSize(2147483647 == 2GB) has been exceeded and suggests to increase the property value in binding element. I tried to set MaxReceivedMessageSize to Int64.MaxValue(== 9223372036854775807) at both side, but it still throws the same exception like the property set by user is ignored. The file size being temporarily transferred is ~2GB as expected.

Server side exception:
QuotaExceededException: The maximum message size quota for incoming messages (2147483647) has been exceeded. To increase the quota, use the MaxReceivedMessageSize property on the appropriate binding element.

My NetTcpBinding:
var netTcpBinding = new NetTcpBinding(SecurityMode.None)
                {
                    TransferMode = TransferMode.Streamed,
                    **MaxReceivedMessageSize = Int64.MaxValue,**
                    MaxBufferSize = 1024 * 1024,
                    ReaderQuotas =
                    {
                        MaxDepth = 32,
                        MaxStringContentLength = 10485760,
                        MaxArrayLength = 16384,
                        MaxBytesPerRead = 4096,
                        MaxNameTableCharCount = 16384
                    },
                    SendTimeout = TimeSpan.FromHours(2),
                        ReceiveTimeout = TimeSpan.FromHours(2),
                        OpenTimeout = TimeSpan.FromHours(2),
                        CloseTimeout = TimeSpan.FromHours(2),
                    };

mconnew:
Fixes #843
If `MaxReceivedMessageSize` was greater than `int.MaxValue`, it would get capped at `int.MaxValue`.  
This does change the public field `FramingConnection.MaxReceivedMessageSize` from an `int` to a `long`. I'm not going to consider this a breaking change for 3 reasons.
1. For something to be a breaking change, it has to break compilation. We don't guarantee binary compatibility, a new release may require a recompile, but I don't expect this change to break that.
2. We're increasing the size of the field so any previous values can still be stored with no data loss.
3. While technically a public API, you would need to create a custom .NET Framing middleware for this API to be useful. I would be incredibly surprised if anyone is doing that.


mconnew:
Btw, here's some interesting information about how long the unit test takes to run on different versions of .NET. I've disabled the test with a Skip reason because of how long it takes to run.
| Framework | Test Run Time |
| ---- | ---- |
| .NET Framework | 15.6 minutes |
| .NET Core 3.1 | 5.5 minutes |
| .NET 6 | 4.7 minutes |

While this isn't a precise test, it does show the performance improvements through the .NET releases in a qualitative way.

g7ed6e (approved):

PR #863, opened 2022-10-04
Async global exception logging missing user-code stacktrace

OlofBlomqvist:
Currently when an async task throws an exception after an await, the stack trace of the original exception is lost so that it is not possible to see where in user-code the exception was thrown from (since only the corewcf specific code is included).

This change makes it so that information regarding the original exception location is included in the final exception message by capturing the original exception and its stack trace.


mconnew:
Test failures are caused by too low a test run timeout for debug builds. It's not a real failure.

mconnew:
Can you rebase onto the latest main? That should fix the test failures.

mconnew:
Thank you, merged.

PR #879, opened 2022-10-18
Added Check That Config File Exists

### LINKED ISSUE #678: throw exception when config file path not found
services.AddServiceModelConfigurationManagerFile(filepath), should throw exception if file is not present.... during startup itself...

lukedukeus:
Fixes #678

### DIFF
--- a/src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ConfigurationManagerExtensions.cs
+++ b/src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ConfigurationManagerExtensions.cs
@@ -13,10 +13,7 @@ namespace CoreWCF.Configuration
     {
         public static IServiceCollection AddServiceModelConfigurationManagerFile(this IServiceCollection builder, string path)
         {           
-            if (!File.Exists(path))
-            {
-                throw new ArgumentException($"File {path} Does not Exist");
-            }
+            if (!File.Exists(path)) { throw new ArgumentException($"File {path} Does not Exist"); }
             
             builder.TryAddSingleton<IConfigurationHolder, ConfigurationHolder>();
             builder.TryAddSingleton<IBindingFactory, BindingFactory>();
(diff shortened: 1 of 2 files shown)


mconnew:
You're missing a using statement. As it's ArgumentException and File that the compiler is complaining about, I think you need to add `using System;` at the top of the file.

lukedukeus:
Ah, thanks. I was too lazy to clone it to my machine, so I just made the edits on web and forgot the using.

mconnew:
Can you rebase onto the latest main, that should fix the test failures. It's just an issue that we added too many tests for the existing timeout so I increased the timeout.

lukedukeus:
Should work now

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ConfigurationManagerExtensions.cs
```diff
@@ -4,13 +4,17 @@
...
+using System.IO;
+using System;
 
 namespace CoreWCF.Configuration
 {
     public static class ConfigurationManagerExtensions
     {
         public static IServiceCollection AddServiceModelConfigurationManagerFile(this IServiceCollection builder, string path)
         {           
+            if (!File.Exists(path)) { throw new ArgumentException($"File {path} Does not Exist"); }
```
mconnew, line 16: Can you add this to the string resources file (Strings.resx). You will need to switch to the old string format style using `{0}` and use `SR.Format(SR.ResourceName, path)`. Also the casing needs to be changed to follow regular English grammar, so the words does and exist should have a lowercase first letter.

src/CoreWCF.ConfigurationManager/src/CoreWCF/Configuration/ConfigurationManagerExtensions.cs
```diff
@@ -4,13 +4,21 @@
...
+using System;
 
 namespace CoreWCF.Configuration
 {
     public static class ConfigurationManagerExtensions
     {
         public static IServiceCollection AddServiceModelConfigurationManagerFile(this IServiceCollection builder, string path)
         {           
+            if (!File.Exists(path)) { throw new FileNotFoundException(SR.Format(SR.FileNotFound, path)); }
+            using (var fs = new FileStream(path, FileMode.Open))
```
mconnew, line 17:
```suggestion
            using (var fs = new FileStream(path, FileMode.Open, FileAccess.Read))
```
 The tests use `TemporaryFileStream` which opens the file with the default sharing mode of `FileShare.Read`. As `Write` wasn't specified in the `FileShare` parameter, it denies subsequently opening the file for write (as long as it's still open by `TemporaryFileStream`). Your code wasn't specifying any `FileAccess` mode, so it requested the default of `ReadWrite`, which gets denied because `TemporaryFileStream` is still holding a file handle which denies opening the file in write mode.  

This small change should fix the tests.

PR #882, opened 2022-10-20
Update CoreWCF.Templates target runtimes

g7ed6e:
Add net7.0
Remove net5.0

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,8 @@ pr:
   paths:
     exclude:
     - src/CoreWCF.Templates/
+    - src/CoreWCFTemplates.sln
+    - templates/TestTemplates.yml
 
 variables:
   _solution: src/CoreWCF.sln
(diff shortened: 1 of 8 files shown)


azure-pipelines.yml
```diff
@@ -9,6 +9,8 @@ pr:
   paths:
     exclude:
     - src/CoreWCF.Templates/
+    - src/CoreWCFTemplates.sln
+    - templates/TestTemplates.yml
```
mconnew, line 13: Did you mean TestTemplatesStage.yml

mconnew:
LGTM. Only question is whether you specified the correct filename in the pipeline excludes. Feel free to merge as that won't need a second review if you modify it.

mconnew (approved):

azure-pipelines.yml
g7ed6e: Yes I pushed a fix. thanks!

PR #895, opened 2022-11-03
Make templates findable in VS when searching for 'Web'

### LINKED ISSUE #893: dotnet new template doesn't show up in VS when filtering for Web projects
The first time I searched with the "Project Type" selected as "Web", the CoreWCF template wasn't found. When clearing the "Project Type", it was then found. On subsequent searches using "Web", it now says it wasn't found using the search criteria, but it then searches generally and finds it. We should add whatever metadata is needed to categorize the template as Web.

image

g7ed6e:
Fixes #893 

image


g7ed6e:
@mconnew this is fixed please see below. Please notice that you have to close all VS instances after installing templates to get the templates loaded at next opening.
image

mconnew (approved):

Issue #906, opened 2022-11-10
RequestDelegateHandler.AcceptWebSocketAsync does not support SubProtocol == null


arontsang:

mconnew:
Btw, if you put in the description of the pull request "Fixes #issuenumber" it will auto close the issue when it's merged.

PR #909, opened 2022-11-10
RequestDelegateHandler.AcceptWebSocketAsync does not support SubProtocol == null

arontsang:
https://github.com/CoreWCF/CoreWCF/issues/906


mconnew:
Thanks, I really appreciate the unit tests to go with the fix. That's also a really interesting approach injecting the ServiceEndpoint into DI, I might use that to clean up some other tests.

Issue #916, opened 2022-11-20
Error using TransportWithMessageCredential with MessageCredentialType.Windows


nmangue:
Hello,

I get an error when I try to setup a WCF service with `WsHttpBinding`, `TransportWithMessageCredential` and `MessageCredentialType.Windows` targeting `net6.0`.

Here is the exception stacktrace :   
   
```
System.Reflection.TargetParameterCountException: Parameter count mismatch.
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at CoreWCF.Security.NegotiateInternal.NegotiateInternalState.Encrypt(Byte[] input)
   at CoreWCF.Security.SspiNegotiationTokenAuthenticator.IssueServiceToken(SspiNegotiationTokenAuthenticatorState sspiState, ReadOnlyCollection`1 authorizationPolicies, SecurityContextSecurityToken& serviceToken, WrappedKeySecurityToken& proofToken, Int32& issuedKeySize)
   at CoreWCF.Security.SspiNegotiationTokenAuthenticator.ProcessNegotiationAsync(SspiNegotiationTokenAuthenticatorState negotiationState, Message incomingMessage, BinaryNegotiation incomingNego)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
```

I have put a solution to reproduce the problem [here](https://github.com/nmangue/TransportWithWindowsMessageCredential) : 

The `System.Net.NTAuthentication.Encrypt` method signature, [invoked from NegotiateInternalState](https://github.com/CoreWCF/CoreWCF/blob/6772206daa830666d29e229924eaea3e458528f8/src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiateInternal/NegotiateInternalState.cs#L201), has changed since `net5.0`.

References:
	[Encrypt in net5.0](https://github.com/dotnet/runtime/blob/cf258a14b70ad9069470a108f13765e0e5988f51/src/libraries/System.Net.Security/src/System/Net/NTAuthentication.cs#L117)
	[Encrypt in net7.0](https://github.com/dotnet/runtime/blob/26da83df77b2391d6a74a4f24b663dc2fb4eda3f/src/libraries/Common/src/System/Net/NTAuthentication.Common.cs#L333)

I will submit a PR to propose a fix.

PR #917, opened 2022-11-20
Update the use of System.Net.NTAuthentication from NegotiateInternalState

nmangue:
Update the calls to System.Net.NTAuthentication to be compatible with net5.0, net6.0 and net7.0
 - Move the reflection code to a specific class
 - Invoke NTAuthentication.Encrypt() with the new signatures
 - Update GetOutgoingBlob search condition since there are new overloads in net7.0

#916

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiateInternal/INTAuthenticationFacade.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiateInternal/INTAuthenticationFacade.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Security.NegotiateInternal
+{
+    internal interface INTAuthenticationFacade
+    {
+        object Instance { get; }
+
+        bool IsCompleted { get; }
+
+        string Protocol { get; }
+
+        bool IsValidContext { get; }
+
+        byte[] GetOutgoingBlob(byte[] incomingBlob, bool throwOnError, out object statusCode);
+
+        int Encrypt(byte[] input, ref byte[] output);
+
+        void CloseContext();
+    }
+}
--- a/src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiateInternal/NTAuthentication.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Security/NegotiateInternal/NTAuthentication.cs
@@ -1,0 +1,89 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Net;
+using System.Reflection;
+using System.Security.Authentication;
+using System;
+using System.Linq;
+
+namespace CoreWCF.Security.NegotiateInternal
+{
+    internal class NTAuthentication : INTAuthenticationFacade
+    {
+        protected static readonly Type s_ntAuthenticationType;
+        protected static readonly ConstructorInfo s_constructor;
+        protected static readonly MethodInfo s_getOutgoingBlob;
+        protected static readonly MethodInfo s_isCompleted;
+        protected static readonly MethodInfo s_isValidContext;
+        protected static readonly MethodInfo s_protocol;
+        protected static readonly MethodInfo s_closeContext;
+        protected static readonly MethodInfo s_encrypt;
+
+        public object Instance { get; }
+
+        public bool IsCompleted => (bool)s_isCompleted.Invoke(Instance, Array.Empty<object>());
+
+        public string Protocol => (string)s_protocol.Invoke(Instance, Array.Empty<object>());
+
+        public bool IsValidContext => (bool)s_isValidContext.Invoke(Instance, Array.Empty<object>());
+
+        internal NTAuthentication() => Instance = CreateInstance();
+
+        static NTAuthentication()
+        {
+            Assembly secAssembly = typeof(AuthenticationException).Assembly;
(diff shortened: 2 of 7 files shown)


birojnayak:
could please run this test https://github.com/CoreWCF/CoreWCF/blob/c957191f733a5c49daa76e2550e089c6cf86bdf2/src/CoreWCF.NetTcp/tests/AuthorizationTest.cs#L33 across platform and see nothing broken... and thanks for submitting the PR

nmangue:
I can't get the test to pass without my changes. I get `System.ServiceModel.Security.SecurityAccessDeniedException : Access is denied` when calling `EchoForAuthorizarionOneRole`. 
The test seems to use a local LDAP server. How should it be set up ?

nmangue:
To run the test I have used my local domain. I have created a "CoreWCFGroupAdmin" user group and assigned it to my user. Then I have executed the tests and they do pass.

Please let me know if I should proceed otherwise.

````
>dotnet test .\CoreWCF.NetTcp\tests\CoreWCF.NetTcp.Tests.csproj --filter "FullyQualifiedName~CoreWCF.NetTcp.Tests.AuthorizationTest"
Test run for C:\repos\CoreWCF\bin\Debug\CoreWCF.NetTcp.Tests\netcoreapp3.1\CoreWCF.NetTcp.Tests.dll (.NETCoreApp,Version=v3.1)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: < 1 ms - CoreWCF.NetTcp.Tests.dll (netcoreapp3.1)
Test run for C:\repos\CoreWCF\bin\Debug\CoreWCF.NetTcp.Tests\net6.0\CoreWCF.NetTcp.Tests.dll (.NETCoreApp,Version=v6.0)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: < 1 ms - CoreWCF.NetTcp.Tests.dll (net6.0)
Test run for C:\repos\CoreWCF\bin\Debug\CoreWCF.NetTcp.Tests\net7.0\CoreWCF.NetTcp.Tests.dll (.NETCoreApp,Version=v7.0)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: < 1 ms - CoreWCF.NetTcp.Tests.dll (net7.0)
Test run for C:\repos\CoreWCF\bin\Debug\CoreWCF.NetTcp.Tests\net472\CoreWCF.NetTcp.Tests.dll (.NETFramework,Version=v4.7.2)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration:  - CoreWCF.NetTcp.Tests.dll (net472)
```

mconnew:
We need to back port this to at least 1.2 as it's a necessary fix for .NET 7 support. We also need to find a way to test this either in CI or at least easily locally using docker containers.

mconnew:
@nmangue, thank you for this fix. I'd actually like to modify the fix a bit to make it more stable going forwards. There's a new public API in .NET 7 that I'd like to use from .NET 7 onwards. You will still need to call it using reflection as we don't multi-target for different .NET versions, but it's a public API which means it can't be broken by changing the API shape in future versions. Continuing to always use the internal implementation makes us vulnerable to future breaks with further internal refactoring.  
The .NET5 version (which works for .NET6 too) is going to be stable as they aren't going to go back and mess around with the previous versions api's. Adopting the public api through reflection should insulate us from changes going forwards. When we eventually drop .NET Framework support and .NET 7 is the oldest supported version (it will likely be .NET 8 we would be targeting and not 7 as .NET 6 will outlive .NET 7), we can make the calls directly instead of reflected.  

[This is the new .NET7+ api](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.negotiateauthentication?view=net-7.0) for Negotiate authentication. [Here's the PR where asp.net core dropped](https://github.com/dotnet/aspnetcore/pull/42364) their reflection based implementation and switched to the new public api.

nmangue:
I have looked at the differences between `NTAuthentication` and `NegotiateAuthentication.` Unless I'm mistaken, it doesn't look like the switch is straightforward. Unlike asp.net core, CoreWCF uses `IsValidContext` and `Encrypt` which are not part of `NegotiateAuthentication`. It seems that using the new methods implies more significant changes. For example `IsValidContext` bubbles up in `WindowsSspiNegotiation`.
For now is it possible to integrate the bug fix with `NTAuthentication` and plan the switch to `NegotiateAuthentication` as a separate modification ?

mconnew:
I think Encrypt and Decrypt are equivalent to Wrap and Unwrap and the name was changed because just signing and not encrypting is an option. I didn't see an easy way to get equivalent info from IsValidContext. I think your suggestion is reasonable. I'll merge it. Can you open an issue for the work to switch this over? There's no obligation on you to do that work.

nmangue:
Thank you ! I have created the issue #978.

PR #927, opened 2022-11-25
WebHttpBinding config section support added

### LINKED ISSUE #805: How to use Config Section Map for WebHttpBinding
I have a tried to load a config which has webHttpBinding but it fails to load and gave a error Unrecognized element webHttpBinding

jvnvenu:
Resolves #805
WebHttpBinding config section support added. This will be helpful when we are using config file based hosting


g7ed6e (approved):
Thanks @jvnvenu for the PR. Having a look to other binding support unit tests I noticed each binding has a specific unit test file which test default values. Do you think you could add one for webHttpBinding too? If you can't that's no problem I will add it later.

jvnvenu:
Thanks @g7ed6e for accepting the changes and I have added test cases

PR #928, opened 2022-11-27
Add hint to find test causing CI error 'There is no currently active test.'

g7ed6e:


mconnew:
This is caused by a test not cleaning up properly and the WebHost wasn't stopped. Rather than swallow the exception and stopping the logger from causing the test run to be aborted, can we instead add something to inform us of the test which caused this? We could embed the test name into the XunitLogger. There are helper methods which are called by the tests to setup the WebHost and these helper methods are what instantiate the XunitLoggerProvider. There are only 12 of them according to Intellisense. We could add an extra parameter which is attributed with `[CallerMemberName]` so that we get the test method name in a string. We can pass that to the XUnitLogerProvider and then where you are detecting the exception, add this information to the exception (I think you need to replace it with a new Exception instance to replace the message) so that we see on the console which test caused this.

mconnew (approved):

PR #929, opened 2022-11-27
Fix logger category in ServiceModelApplicationBuilderExtensions

### LINKED ISSUE #926: Wrong logger category in ServiceModelApplicationBuilderExtensions
The logger created here in `ServiceModelApplicationBuilderExtensions` should be using `typeof` instead of `nameof`. Currently the "CoreWCF" prefix is missing in the logger category, so the messages here are still logged, even if the logger configuration has a separate log level for CoreWCF.

https://github.com/CoreWCF/CoreWCF/blob/6772206daa830666d29e229924eaea3e458528f8/src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceModelApplicationBuilderExtensions.cs#L39

mus65:
Fixes #926


g7ed6e (approved):
Thanks @mus65 for the quick fix.

PR #936, opened 2022-12-02
Use DataProtection api

g7ed6e:


src/CoreWCF.Primitives/src/CoreWCF/Security/DataProtectionSecurityStateEncoder.cs
```diff
@@ -56,11 +49,25 @@ public override string ToString()
             return result.ToString();
         }
 
+
+        private IDataProtector GetDataProtector()
+        {
+            HttpContext context = null;
+            if (OperationContext.Current.RequestContext.RequestMessage.Properties.TryGetValue("Microsoft.AspNetCore.Http.HttpContext", out object contextObj))
```
g7ed6e, line 56: i guess it will only work with http based Binding, we need to a way to easyly retrieve the root `ServiceProvider` whatever binding is used

mconnew: Why can't you use the same mechanism as is used by the injection code generator?

g7ed6e: @mconnew Retrieving the `IServiceProvider` using `OperationContext.Current.InstanceContext.Extensions.Find<IServiceProvider>()` is possible when the service has been added to DI.
When the service is not in DI, it's only possible when service is configured with `InstanceContextMode.Single`.

https://github.com/CoreWCF/CoreWCF/blob/bbc1a9f3c97efad629598926e11852e6d2dfefb2/src/CoreWCF.Primitives/src/CoreWCF/ServiceBehaviorAttribute.cs#L259

birojnayak:
why are we changing this ? I mean issues details will be helpful....

g7ed6e:
> why are we changing this ? I mean issues details will be helpful....

The intention is related to #624. Could you point me to a test of this feature ? I did not find yet how to configure the service to use it.

PR #938, opened 2022-12-06
Fix OneWayPatternTest

g7ed6e:


src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
```diff
@@ -20,19 +21,26 @@ public AsyncNetAdoptionTests(ITestOutputHelper output)
...
+                Services.OneWayService.TestOutputHelper = _output;
                 var httpBinding = ClientHelper.GetBufferedModeBinding();
                 var factory = new System.ServiceModel.ChannelFactory<ClientContract.IOneWayContract>(httpBinding,
                     new System.ServiceModel.EndpointAddress(new Uri("http://localhost:8080/OneWayPatternTest/basichttp.svc")));
-                var channel = factory.CreateChannel();                              
-                Task task = channel.OneWay("Hello");
+                var channel = factory.CreateChannel();
+                Parallel.ForEach(inputs, s => channel.OneWay(s));
+                await Task.WhenAll(Services.OneWayService.Tasks);
+                foreach (string input in inputs)
```
mconnew, line 38: There is still a race condition here as it's possible not all the request will have added a Task to OneWayService.Tasks yet. You could get a concurrent modification exception if WhenAll tries to iterate the list of Tasks before the final Task has been added. There's the same problem in the following foreach. You might have 9 Tasks in the list of Tasks which have completed, and the 10th is still pending execution and hasn't put its string into the Inputs yet.  
There's an easy fix though. Instantiate a CountdownEvent with the number of calls you want to have happened. In each service call, decrement the CountdownEvent using Signal. In the test, wait with a timeout on the CountdownEvent.

src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
```diff
@@ -29,26 +30,50 @@ public void OneWayPatternTest()
...
+                var httpBinding = ClientHelper.GetBufferedModeBinding();
+                var factory = new System.ServiceModel.ChannelFactory<ClientContract.IOneWayContract>(httpBinding,
+                    new System.ServiceModel.EndpointAddress(new Uri("http://localhost:8080/OneWayPatternTest/basichttp.svc")));
+                var channel = factory.CreateChannel();
+                var tasks = inputs.AsParallel().Select(x => channel.OneWay(x)).ToList();
+                await Task.WhenAll(tasks);
+                foreach (string input in inputs)
+                {
+                    Assert.Contains(input, host.Services.GetService<ConcurrentBag<string>>());
+                }
```
mconnew, line 56: You still have a potential race here. It's unlikely to happen, but the task created in the service which adds the item to the concurrent bag might not execute before this validation step runs. In other words you might still have an entry missing. It would require the CPU to be heavily loaded and be running on a low core count machine, but the CI machines tend to fit this criteria (2 cores, highly contended CPU).

mconnew (approved):

PR #939, opened 2022-12-06
Enable connections to remote RabbitMQ hosts

jonlouie:
## Summary 
* Enables connections to remote RabbitMQ hosts
* Modify RabbitMqBinding to use a URI

## Testing
* Unit tests for queue configurations
* End-to-end integration tests for binding usage and queue creation/usage


src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -2,89 +2,68 @@
...
         /// <summary>
-        /// Uses the specified protocol. The broker must be set before use.
+        /// Gets the scheme used by the binding
         /// </summary>
-        /// <param name="protocol">The protocol version to use</param>
-        public RabbitMqBinding(IProtocol protocol)
+        public override string Scheme
+        {
+            get { return CurrentVersion.Scheme; }
+        }
```
mconnew, line 45: It's typical for a binding to defer to the transport for the scheme. So this would become:
```c#
public override string Scheme => _transport.Scheme;
```

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,177 @@
...
+using System.Linq;
+using System.Net;
+using RabbitMQ.Client;
+
+namespace CoreWCF.Channels
+{
+    public class RabbitMqConnectionSettings
+    {
+        public const string TempQueuePrefix = "corewcf-temp-";
+        private const string AMQPScheme = "amqp";
```
mconnew, line 14: const values are usually private or internal unless it's a value we expect users to consume.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,177 @@
...
+        public bool AutomaticRecoveryEnabled => true;
+
+        public static RabbitMqConnectionSettings FromUri(
+            Uri uri,
+            ICredentials credentials = null,
+            SslOption sslOption = null,
+            string virtualHost = DefaultVirtualHost)
+        {
+            sslOption = ConfigureSslOption(sslOption, uri);
+            var queueName = GetQueueNameFromUri(uri);
```
mconnew, line 36: Need a null check on uri otherwise ConfigureSslOption can throw NRE if sslOption is null.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,177 @@
...
+                return ConnectionFactory.DefaultPass;
+            }
+
+            return uri.UserInfo.Split(':').Last();
+        }
+
+        private static SslOption ConfigureSslOption(SslOption sslOption, Uri uri)
+        {
+            if (sslOption != null)
+            {
```
mconnew, line 165: As uri can't be null as it's needed in other methods, what happens if the passed in sslOption doesn't match the uri? Do we care? Is it legitimate to have a different ServerName than uri? Is ServerName used to validate the certificate? If that's the case, then that could be the case. What if uri contains SecureAMQPSScheme but SslOption is false?

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,177 @@
...
+                AutomaticRecoveryEnabled = AutomaticRecoveryEnabled
+            };
+        }
+
+        private static string GetQueueNameFromUri(Uri uri)
+        {
+            if (uri.Segments.Length < 2)
+            {
+                throw new ArgumentException($"Expected a non-empty segment in the uri {uri}");
+            }
```
mconnew, line 75: Put exception text into string resource. Also should state the expected format of the uri in the exception message.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (!string.IsNullOrEmpty(passwordFromCredentials))
+            {
+                return passwordFromCredentials;
+            }
+
+            return GetPasswordFromUri(uri);
+        }
+
+        private static string GetPasswordFromUri(Uri uri)
+        {
```
mconnew, line 153: I don't believe we should support or encourage encoding the password as part of the uri. It's a bad idea from a security perspective. Uri's can get logged in log files for example.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqQueueMessageContext.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace CoreWCF.Channels
+{
+    public class RabbitMqQueueMessageContext : QueueMessageContext
+    {
+        /// <summary>
+        /// Delivery confirmation number for a message read from a RabbitMQ queue
+        /// </summary>
+        public ulong DeliveryTag { get; set; }
+    }
```
mconnew, line 14: Is ulong necessary? Are the numbers ever > int64.MaxValue? The reason I ask is ulong isn't cls compliant which means there are some .NET languages which can't use this api. Generally using non-cls compliant types should be avoided if possible. If it really does need an unsigned 64 bit number, then it's okay.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -12,26 +13,25 @@ namespace CoreWCF.Channels
...
-        /// <summary>
-        /// Creates a new instance of the RabbitMQTransportBindingElement Class using the default protocol.
-        /// </summary>
         public RabbitMqTransportBindingElement()
         {
         }
-
+        
         private RabbitMqTransportBindingElement(RabbitMqTransportBindingElement other)
         {
```
mconnew, line 21: Should call the base constructor passing other as there might be some properties on QueueBaseTransportBindingElement or TransportBindingElement which need to be copied.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -68,20 +68,29 @@ public override string Scheme
         public override long MaxReceivedMessageSize { get; set; }
```
mconnew, line 69: This should have a default value otherwise it will be 0. This is important for when using a CustomBinding

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -68,20 +68,29 @@ public override string Scheme
...
 
         /// <summary>
-        /// The credentials to use when authenticating with the broker
+        /// Specifies the version of the AMQP protocol that should be used to
+        /// communicate with the broker
         /// </summary>
-        internal ICredentials Credentials { get; set; }
+        public IProtocol BrokerProtocol { get; set; }
```
mconnew, line 75: Should a default value be set?

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitmqTransportPump.cs
```diff
@@ -17,59 +20,117 @@ namespace CoreWCF.Channels
 {
     public class RabbitMqTransportPump : QueueTransportPump
     {
```
mconnew, line 22: This isn't a type that would need to be instantiated by developer code. I think this should be internal.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/Configuration/QueueDeclareConfiguration.cs
```diff
@@ -0,0 +1,88 @@
...
+            get
+            {
+                return _prefetchCount;
+            }
+            set
+            {
+                if (value < 1)
+                {
+                    throw new ArgumentException($"PrefetchCount value of {value} is invalid. PrefetchCount must be greater than 0.");
+                }
```
mconnew, line 63: Use ArgumentOutOfRangeException. Also move text to string resources.

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
```diff
@@ -1,63 +1,247 @@
...
             using (host)
             {
-                host.Start();
-                await Task.Delay(3000);
-                MessageQueueHelper.SendMessageInQueue();
+                // When the host is started, queues are created if they do not already exist.
+                // By waiting for the host to initialize completely, we avoid the race condition
+                // of trying to access a queue that has not finished creation.
+                Task.WaitAll(Task.Run(() => host.Start()));
+
```
mconnew, line 43: I don't understand what you are doing here. host.Start() doesn't return until the host has completely started. You can [see the code here](https://github.com/dotnet/aspnetcore/blob/b10b40ae644995533e59f40f89893c873b3afa42/src/Hosting/Hosting/src/Internal/WebHost.cs#L130-L133) where it calls StartAsync() and waits for the Task to be completed. By using Task.Run, you are starting a Task running on another thread which calls host.Start(), which calls host.StartAsync(), then blocks the thread until that Task has completed, which then completes the thread from Task.Run(). Task.WaitAll() blocks until the passed in Task has completed. If things aren't completely set up by the time that host.Start() completes normally, then there's some work being completed after Start() returns, and all the rest of the code is just adding in a small bit of delay and things are accidentally working. Can you add a step to enumerate all the queues on the server in MessageQueueHelper.SendMessageToQueue to ensure things are setup with a retry with a delay.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
jonlouie: * The idea here was to provide a default sslOption if one isn't provided, but if one is provided, use that even if it's a misconfiguration. 
* ServerName is supposed to match the SAN or CN of the host in the SSL cert [(source)](https://www.rabbitmq.com/ssl.html#dotnet-connection-params). Based on this, my understanding is that the ServerName should match the host uri. Does this sound correct to you?
* If the uri contains SecureAMQPSScheme and SslOption.Enabled is false, connection to the RabbitMQ host will fail (`BrokerUnreachableException`)

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqBinding.cs
```diff
@@ -2,89 +2,68 @@
...
+using CoreWCF.Channels.Configuration;
 using RabbitMQ.Client;
 
 namespace CoreWCF.Channels
 {
     public class RabbitMqBinding : Binding
     {
-        private TextMessageEncodingBindingElement _textMessageEncodingBindingElement;
+        public const long DefaultMaxMessageSize = 8192L;
         private RabbitMqTransportBindingElement _transport;
```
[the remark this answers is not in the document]
jonlouie: * I thought this was the recommended default message size in the docs, but I'm having trouble finding that now as well. The new value will be 512MB which is the max message size [(source)](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit_common/include/rabbit.hrl#L254-L257) as of RabbitMQ 3.8.
* Accessor will be made private

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
jonlouie: I think you're right, the extra code was adding enough delay to ensure things worked. Originally, these tests had a thread delay to ensure setup completed and wrapping host.Start() in a task got things working after removing the delay. I'll add the additional step w/ retry to ensure determinism in our tests.

jonlouie: Tests seem to run just fine after replacing the Task wrapper with just `host.Start()`. The RabbitMQ client does not have an api to enumerate queues so I'm leaning towards leaving things as they are and add retry logic within `MessageQueueHelper.SendMessageToQueue` if we see failure messages in the future.

Documentation/Testing.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Running integration tests
+
+Some integration tests are dependent on external services (e.g. `CoreWCF.RabbitMQ.Tests` require access to a RabbitMQ host), thereby requiring additional setup before those tests can run successfully. The steps below will create local containers hosting all dependent services required by the integration tests.
+
+1. [Install docker v19.03.0+](https://docs.docker.com/engine/install/) to your development environment via Docker Desktop or the docker engine
+2. From the root directory of the CoreWCF repo, run the following command to start the RabbitMQ container:
+
+        docker-compose up -d
+3. Run tests as usual, using your IDE or running `dotnet test` from the command line.
```
mconnew, line 15: Can you include the command to stop the containers and clean up in the instructions?

Documentation/Testing.md
```diff
@@ -0,0 +1,19 @@
+# Getting started with testing
+
+Most tests can be run without any special commands or processes. Running the following command from the project's root directory will restore packages, build the solution, and run the full CoreWCF test suite:
+
+    dotnet test
```
mconnew, line 5: I think you need to specify the path to the .sln file as we have 2 solution files and they are in the src folder.

Documentation/Testing.md
jonlouie: Good catch- added instructions

Documentation/Testing.md
jonlouie: Added the solution path to the command

src/CoreWCF.Primitives/src/CoreWCF/Channels/ReceiveContext.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace CoreWCF.Channels
 {
-    internal abstract class ReceiveContext
+    public abstract class ReceiveContext
     {
```
mconnew, line 12: Now this is being made public, I think we should remove the sync versions of the methods. Otherwise everyone must override both.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueMessageContext.cs
```diff
@@ -14,6 +14,7 @@ namespace CoreWCF.Queue.Common
     public class QueueMessageContext : RequestContext
     {
         public PipeReader QueueMessageReader { get; set; }
+        public ReceiveContext ReceiveContext { get; set; }
         public virtual IDictionary<string, object> Properties { get { return _properties.Value; } }
```
mconnew, line 18: This shouldn't be public otherwise there's conflicting information. What if the request message doesn't have the receive context but this property does? I'd be okay with this being a helper property to add or get the value from the request message, but it shouldn't be an independent property backed by it's own value.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
```diff
@@ -0,0 +1,145 @@
...
+
+using System;
+using System.Linq;
+using System.Net;
+using RabbitMQ.Client;
+
+namespace CoreWCF.Channels
+{
+    public class RabbitMqConnectionSettings
+    {
```
mconnew, line 12: I couldn't find this type exposed publicly anywhere. Does it need to be public?

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -3,35 +3,38 @@
...
 namespace CoreWCF.Channels
 {
-    public class RabbitMqTransportBindingElement : QueueBaseTransportBindingElement
+    public sealed class RabbitMqTransportBindingElement : QueueBaseTransportBindingElement
     {
-        /// <summary>
-        /// Creates a new instance of the RabbitMQTransportBindingElement Class using the default protocol.
-        /// </summary>
+        private const long DefaultMaxMessageSize = 512 * 2 ^ 20;  // 512 MB, max message size as of RabbitMQ v3.8
+
```
mconnew, line 17: WCF normally uses a MaxReceivedMessageSize of 64K. This is to limit memory consumption by malicious clients. 512MB is a massive amount of memory to allocate and if your messages aren't expected to be that big, this could be a problem. It's typical for users to increase the max message size if they need to transfer large amounts of data. 64K is generally enough for simple RPC calls and usually only needs to be increased if data blobs are being sent.

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -65,23 +68,32 @@ public override string Scheme
         /// <summary>
         /// The largest receivable encoded message
         /// </summary>
-        public override long MaxReceivedMessageSize { get; set; }
+        public override long MaxReceivedMessageSize { get; set; } = DefaultMaxMessageSize;
```
mconnew, line 72: The setter should have validation to make sure the value is > 0 and <=512MB

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
```diff
@@ -18,46 +21,198 @@ namespace CoreWCF.RabbitMQ.Tests
...
                 Assert.True(testService.ManualResetEvent.Wait(System.TimeSpan.FromSeconds(5)));
             }
         }
     }
 
-    public class Startup
+    public class ClassicQueueWithTLSStartup
+    {
+        public static Uri Uri = new("net.amqps://HOST:PORT/amq.direct/QUEUE_NAME#ROUTING_KEY");
+        private static readonly ICredentials s_credentials = new NetworkCredential(ConnectionFactory.DefaultUser, ConnectionFactory.DefaultPass);
```
mconnew, line 100: Did you mean to leave HOST and PORT not populated with real values?

src/CoreWCF.RabbitMQ/tests/ServiceHelper.cs
```diff
@@ -19,7 +19,7 @@ public static IWebHostBuilder CreateWebHostBuilder<TStartup>(ITestOutputHelper o
 #if DEBUG
                 .ConfigureLogging((ILoggingBuilder logging) =>
                 {
-                    logging.AddProvider(new XunitLoggerProvider(outputHelper));
+                    logging.AddProvider(new XunitLoggerProvider(outputHelper, nameof(CreateWebHostBuilder)));
                     logging.AddFilter("Default", LogLevel.Debug);
```
mconnew, line 23: This is supposed to be the name of the test. Have a look how it's used in other repos. We add a parameter to the CreateDefaultBuilder method which is attributed to be filled in by the compiler with the calling method name.

templates/CodeAnalysis.yml
```diff
@@ -15,6 +15,13 @@ stages:
     pool:
       vmImage: 'ubuntu-latest'
     steps:
+    - bash: |
+        echo 'Starting docker services...'
+        docker-compose up -d
```
mconnew, line 20: Should we also have an explicit step to stop the container or just rely on Azure tearing everything done after it's finished the test run?

templates/TestStage.yml
```diff
@@ -90,32 +104,46 @@ stages:
...
     pool:
       vmImage: $(imageName)
     steps:
+    - bash: |
+        echo 'Starting docker services...'
+        docker-compose up -d
+      name: DockerComposeLinuxServices
+      displayName: Start services hosted by linux containers
+      condition: eq(variables.platformOs, 'linux')
+      timeoutInMinutes: 5
```
mconnew, line 144: You could condition on the image name instead. It's ubuntu-latest for Linux

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqConnectionSettings.cs
jonlouie: It remained public for unit tests and under the assumption that it would be used in anything client-related

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
jonlouie: Yes, this startup class is used by a test that targets a TLS-enabled RabbitMQ host, and this test is skipped because our test service container is not configured to use that connection type.

templates/CodeAnalysis.yml
jonlouie: Added the tear-down command to match the documentation I added

src/CoreWCF.RabbitMQ/src/CoreWCF/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -65,23 +70,48 @@ public override string Scheme
...
+        {
+            get
+            {
+                return _maxReceivedMessageSize;
+            }
+
+            set
+            {
+                if (value < 0 || value >= MaxRabbitMqMessageSize)
+                {
```
mconnew, line 83: This will throw if the value is exactly MaxRabbitMqMessageSize. I think this was supposed to be `>` and not `>=`. The zero check should also be `<=` as it doesn't make sense to set a value of 0.

mconnew: I just looked at the code for MaxReceivedMessageSize on .NET Framework and when the value \<= 0, it throws `new ArgumentOutOfRangeException(nameof(value), value,
SR.ValueMustBePositive)`. You can keep the single throw if you want, but the constructor overload you are using would be better to be switched to this one as you name the argument and provide the value too. You also missed adding any text to the resource string file, you left a place holder.

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
```diff
@@ -18,46 +21,198 @@ namespace CoreWCF.RabbitMQ.Tests
...
         }
 
-        [Fact(Skip ="need rabbitmq")]
-        public async Task ReceiveMessage_ServiceCall_Success()
+        [Fact(Skip = "Requires RabbitMQ host with SSL")]
+        public void ClassicQueueWithTls_ReceiveMessage_Success()
         {
-            IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output).Build();
+            IWebHost host = ServiceHelper.CreateWebHostBuilder<ClassicQueueWithTLSStartup>(_output, nameof(ClassicQueueWithTls_ReceiveMessage_Success)).Build();
             using (host)
```
mconnew, line 34: I didn't mean to add it at the call point take a look at the implementation [here](https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs#L88) where it uses `[CallerMemberName] string callerMethodName = ""` as a parameter in the CreateWebHostBuilder method. Let the compiler do the work for you.

PR #943, opened 2022-12-13
Introduce CoreWCF.Kafka

g7ed6e:
Introduce CoreWCF.Kafka.


src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPumpBase.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        Consumer.Subscribe(_topic);
+
+        CancellationToken ct = CancellationTokenSource
+            .CreateLinkedTokenSource(_cancellationTokenSource.Token, token).Token;
+        while (!ct.IsCancellationRequested)
+        {
+            try
+            {
+                var consumeResult = Consumer.Consume(token);
```
g7ed6e, line 84:
```suggestion
                var consumeResult = Consumer.Consume(ct);
```

src/CoreWCF.Kafka/tests/Resources/rabbitmqTestMessage.bin
```diff
@@ -0,0 +1 @@
+<s:Envelope xmlns:s="http://www.w3.org/2003/05/soap-envelope" xmlns:a="http://www.w3.org/2005/08/addressing"><s:Header><a:Action s:mustUnderstand="1">http://tempuri.org/ITestContract/Create</a:Action><VsDebuggerCausalityData xmlns="http://schemas.microsoft.com/vstudio/diagnostics/servicemodelsink">uIDPo4IM/u5L8tFGizKmemV7kmEAAAAAcStFY6F8ukKPmzryw1DTrsJZH/CtAKlEjIDYla6EnwYACQAA</VsDebuggerCausalityData></s:Header><s:Body><Create xmlns="http://tempuri.org/"><name>TestMessage</name></Create></s:Body></s:Envelope>
```
g7ed6e, line 1: Insert a guid to assert.
Remove VsDebuggerCausalityData.
Extract to template string.

src/CoreWCF.Kafka/tests/IntegrationTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                var result = await producer.ProduceAsync(_kafkaFixture.TopicName, new Message<Null, byte[]> { Value = bytes });
+
+                Assert.True(result.Status == PersistenceStatus.Persisted);
+
+                producer.Flush(TimeSpan.FromSeconds(3));
+
+                var resolver = new DependencyResolverHelper(host);
+                var testService = resolver.GetService<TestService>();
+                Assert.True(testService.ManualResetEvent.Wait(TimeSpan.FromSeconds(3)));
```
g7ed6e, line 57: Assert guid sent

guillaume-chervet:
hi @g7ed6e , may you describe a little bit the architecture? and the need ?

birojnayak:
> Introduce CoreWCF.Kafka.KafkaBinding.
> 
> Two 2 consumption semantics are supported `AtLeastOnce` and `AtMostOnce`.
> 
> Kafka security support:
> 
>     * None (discouraged)
> 
>     * SaslPlainText (authentication but no encryption - discouraged)
> 
>     * Ssl (no authentication but TLS encryption)
> 
>     * SaslSsl (authentication + TLS encryption) suitable for Azure Event Hub or Confluent Cloud.
> 
>     * mTLS

@g7ed6e Let me know once it's ready to review and if there is a possibility of merging all into one (later is not mandatory though).

g7ed6e:
> @g7ed6e Let me know once it's ready to review and if there is a possibility of merging all into one (later is not mandatory though).

@birojnayak i just squashed all commits. You can start to review. I still need to rebase the fix from #1018  which is not merged yet. I need it for integration tests because i check the consumer lag and topic offsets are commited from the `ReceivedContext`.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPumpBase.cs
```diff
@@ -0,0 +1,191 @@
...
+                try
+                {
+                    var consumeResult = Consumer.Consume(ct);
+                    if (consumeResult.IsPartitionEOF)
+                    {
+                        continue;
+                    }
+
+                    _logger.LogInformation("Receiving message from kafka at {topicPartitionOffset}", consumeResult.TopicPartitionOffset);
+                    try
```
g7ed6e, line 114: remove the try/catch that surround `OnConsumerMessage` because these exceptions will be caught by the queue middlewares and the dispatch runtime

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueMessageContext.cs
```diff
@@ -76,20 +76,9 @@ public override void Abort()
...
-                {
-                    await receiveContext.AbandonAsync(CancellationToken.None);
-                }
-                else
-                {
-                    await receiveContext.CompleteAsync(CancellationToken.None);
-                }
-            }
-        }
+        public override Task ReplyAsync(Message message) => message?.IsFault == true
```
g7ed6e, line 79: to be replaced by #1018

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueFetchMessage.cs
```diff
@@ -19,17 +20,25 @@ public QueueFetchMessage(QueueMessageDispatcherDelegate next)
...
         {
-            var pipeReader = queueMessageContext.QueueMessageReader;
-            var readResult = await pipeReader.ReadAsync();
-            var memStream = new MemoryStream(readResult.Buffer.ToArray());
-            var encoder = queueMessageContext.QueueTransportContext.MessageEncoderFactory.Encoder;
-            var maxReceivedMessageSize = queueMessageContext.QueueTransportContext.QueueBindingElement.MaxReceivedMessageSize;
-            var message = await encoder.ReadMessageAsync(memStream, (int)maxReceivedMessageSize);
-            if (message.Headers.To == null)
+            try
             {
```
g7ed6e, line 24: to be replaced by #1018

g7ed6e:
@mconnew @birojnayak I think we should discuss the public API of the binding and the security for both client and server sides. To get the best throughput, Kafka client need to be finegrained tuned with a lot of properties (directly issued from librdkafka https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md). We could give user full control over configuring librdkafka or we can hide everything. please tell me what is your favorite option.

mconnew:
@g7ed6e, generally the pattern is the Binding has a few common properties exposed with sensible defaults used for everything not exposed. If you want to set anything not exposed on the Binding, then you create a custom binding and everything is available on the TransportBindingElement. Does that model make sense for Kafka?

g7ed6e:
> > @mconnew @birojnayak I think we should discuss the public API of the binding and the security for both client and server sides. To get the best throughput, Kafka client need to be finegrained tuned with a lot of properties (directly issued from librdkafka https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md). We could give user full control over configuring librdkafka or we can hide everything. please tell me what is your favorite option.
> 
> > @g7ed6e, generally the pattern is the Binding has a few common properties exposed with sensible defaults used for everything not exposed. If you want to set anything not exposed on the Binding, then you create a custom binding and everything is available on the TransportBindingElement. Does that model make sense for Kafka?
> 
> @g7ed6e I agree with @mconnew . So the way we approached RabbitMQ was that we came up with a short write-up for all properties exposed. We brainstormed around how to best fit it around the binding approach, where we will expose or make the bare minimum properties mandatory for users, and the rest can be set to default with the best interest for users and an option to override. So if you are asking for "full control/hide everything", I would suggest we find a middle ground.
> 
> So if it sounds reasonable, you can stick to the above approach, and while reviewing the code, I can go over connection properties to see which would make sense to expose or hide.

Thanks @mconnew / @birojnayak i will rework a bit the public API on both client and server side. 
The client side `KafkaOutputChannel` and the server side `KafkaTransportPumpBase` logics should remain as is except for the initialization part.. which build producer and consumer.

@birojnayak I will apply the following updates

`CoreWCF.Channels.KafkaTransportBindingElement` and `CoreWCF.Channels.KafkaTransportSecurity` will expose all `ConsumerConfig` [properties](https://github.com/confluentinc/confluent-kafka-dotnet/blob/19bd96061b4d063a0bda9a33bb20d667a829bc68/src/Confluent.Kafka/Config_gen.cs#L1213) tagged *high* or *medium* except `EnableAutoCommit` and `EnableAutoOffsetStore` that i need to control to provide either `AtLeastOnce` or `AtMostOnce` behavior. 

`CoreWCF.ServiceModel.Channels.KafkaTransportBindingElement` and `CoreWCF.ServiceModel.Channels.KafkaTransportSecurity` will expose all `ProducerConfig` [properties](https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Config_gen.cs#L994) tagged *high* or *medium*.

Then we will optout some of them as you proposed during code review.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,230 @@
...
+        (kafkaDeliverySemantics, consumerConfig) switch
+        {
+            // KafkaBinding
+            (KafkaDeliverySemantics.AtMostOnce, { EnableAutoCommit: null, EnableAutoOffsetStore: null }) => s_atMostOnceConfigValues,
+            (KafkaDeliverySemantics.AtLeastOnce, { EnableAutoCommit: null, EnableAutoOffsetStore: null } ) => s_atLeastOnceBatchCommitConfigValues,
+            // CustomBinding
+            (KafkaDeliverySemantics.AtMostOnce, { EnableAutoCommit: false, EnableAutoOffsetStore: null }) => s_atMostOnceConfigValues,
+            (KafkaDeliverySemantics.AtLeastOnce, { EnableAutoCommit: true, EnableAutoOffsetStore: false } ) => s_atLeastOnceBatchCommitConfigValues,
+            (KafkaDeliverySemantics.AtLeastOnce, { EnableAutoCommit: false, EnableAutoOffsetStore: null }) => s_atLeastOncePerMessageCommitConfigValues,
+            _ => throw new NotSupportedException(string.Format(SR.ProvidedKafkaConfigurationIsNotValid, kafkaDeliverySemantics, consumerConfig.EnableAutoCommit, consumerConfig.EnableAutoOffsetStore))
```
birojnayak, line 82: can we do this validation before we start service ?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportSecurity.cs
```diff
@@ -0,0 +1,132 @@
...
+
+    public string Username { get; }
+    public string Password { get; }
+}
+
+public class ClientCertificateCredential
+{
+    public string CertificatePem { get; set; }
+    public string KeyPem { get; set; }
+    public string KeyPassword { get; set; }
```
birojnayak, line 131: it would be nice to keep name same as ConsumerClient..

src/CoreWCF.Kafka/src/CoreWCF/Channels/SaslMechanismHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace CoreWCF.Channels;
+
+internal static class SaslMechanismHelper
+{
+    public static bool IsDefined(SaslMechanism value)
+    {
+        return value == SaslMechanism.Plain
+               || value == SaslMechanism.Gssapi
+               || value == SaslMechanism.ScramSha256
+               || value == SaslMechanism.ScramSha512
```
birojnayak, line 15: I don't think we have code for GSSAPI , SCRAM... let's throw exception

birojnayak (changes requested):
Reviewed the server and test.. pending review client. Thank you for this PR.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,230 @@
...
+                {
+                    var consumeResult = _consumer.Consume(ct);
+                    if (config.EnablePartitionEof == true && consumeResult.IsPartitionEOF)
+                    {
+                        continue;
+                    }
+
+                    _logger.LogInformation("Received message from kafka at {topicPartitionOffset}", consumeResult.TopicPartitionOffset);
+                    if (_kafkaDeliverySemantics == KafkaDeliverySemantics.AtMostOnce)
+                    {
```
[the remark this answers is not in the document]
birojnayak: @g7ed6e I also had a conversation with @mconnew to capture failure scenarios... we can probably expose a 1) DLQ in the kafkatransportbinding which if populated can be used to send failed message or 2) we can expose topic where failed message can be written... problem with (1) is we don't know what is the Queue used based on that logic would vary. Problem with (2) is we will bring Producer side dependency to the Corewcf service(consumer side)...  (3) option would be provide an extension method for user to handle via Kafka transport... let me know what you and @mconnew think.. if kafka provides out of the boox support for DLQ let me know, I couldn't find that.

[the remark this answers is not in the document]
mconnew: I don't like the idea of having an extension method, it's too high of a bar of how much code someone would need to write to use the feature. It should just be doable by configuring the binding.  
Having a very quick look over the Confluent Kafka documentation, it looks like there's no additional dependencies between consumer and producer. [They both use](https://developer.confluent.io/get-started/dotnet/#create-project) the Confluent.Kafka package.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
mconnew: It's difficult to do before service start as there are scenarios where EnableAutoOffsetStore can be null, and there are scenarios where EnableAutoCommit can be false. If you tried to validate before starting, where would you put the validation? It can't be any property setter as you can't presume order of properties being set and it's only the combination of values which is a problem. As long as the exception is thrown during startup, that's fine.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
birojnayak: @mconnew I am fine with option (2)... let's see @g7ed6e thinks..

src/CoreWCF.Kafka/tests/KafkaSemanticsTests.cs
```diff
@@ -0,0 +1,206 @@
...
+using CoreWCF.Queue.Common.Configuration;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CoreWCF.Kafka.Tests;
+
+public class KafkaSemanticsTests : IntegrationTest
```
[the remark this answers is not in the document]
g7ed6e: I did not implemented a message deduplication feature. The tests in this file relates to the way we manage storing offset and committing in the consumer.
Kafka producer implements a kind of deduplication feature through the idempotent property of a producer but it acts at the transport level. this feature is the librdkafka / Confluent.Kafka libraries responsibility I think.
Please see
https://docs.confluent.io/platform/current/clients/confluent-kafka-dotnet/_site/api/Confluent.Kafka.ProducerConfig.html#Confluent_Kafka_ProducerConfig_EnableIdempotence
https://www.conduktor.io/kafka/idempotent-kafka-producer/#Kafka-Idempotent-Producer-0

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
g7ed6e: @birojnayak / @mconnew afaik DLQ is not supported OOB by kafka. the easiest way indeed is to add a producer. For the time being i'm just discarding events to avoid blocking the consumer and trigger a rebalance.

mconnew:
> @mconnew Matt Connew FTE @birojnayak I think we should discuss the public API of the binding and the security for both client and server sides. To get the best throughput, Kafka client need to be finegrained tuned with a lot of properties (directly issued from librdkafka https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md). We could give user full control over configuring librdkafka or we can hide everything. please tell me what is your favorite option.

Lately I've been thinking a lot about integrating IConfiguration into CoreWCF. What do you think of the idea of having some common properties that you would want to modify be properties on the binding/binding element, but the advanced properties coming from configuration? I'm of two minds about having some settings only being available via config. That is a lot of options and creates a massive api surface. Some other ideas:

1. We could expose a Dictionary\<string,string\> called something like AdvancedSettings. If it was an IDictionary\<string,string\> then the implementation doesn't need to be an actual dictionary and that would enable validating the data as it's being set. E.g. max.in.flight needs to be an integer withing a set range.
2. We could add an index property on one of the classes to enable setting these things.
3. There's the IConfiguration concept I mentioned above. On .NET Framework, all the bindings have a constructor overload that takes a configuration name. We could take that base idea but have it accept an IConfiguration instead.

src/CoreWCF.Kafka/tests/AzureEventHubIntegrationTests.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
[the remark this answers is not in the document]
g7ed6e: Sure the plan is to move it to the sample repo once the feature will be available

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
g7ed6e: > shouldn't we commit once we process successfully ? (I do understand it's debatable.. if not we need to expose a way for user to act something suggested [here](https://www.confluent.io/en-gb/blog/spring-kafka-can-your-kafka-consumers-handle-a-poison-pill/#dead-letter-topic) if it fails. Else it would be hard for debugging.

@birojnayak the preferred Confluent.Kafka .NET way to implement at most once is described [here](https://docs.confluent.io/kafka-clients/dotnet/current/overview.html#consumer). Basically we need to
- disable autooffsetstore and manually call StoreOffsets
- enable autocommit

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        ConsumerConfig config = BuildConfig();
+        _consumer = new ConsumerBuilder<Null, byte[]>(config)
+            .SetKeyDeserializer(Deserializers.Null)
+            .SetValueDeserializer(Deserializers.ByteArray)
+            .SetLogHandler(OnLog)
+            .SetErrorHandler(OnError)
+            .Build();
+
+        _consumer.Subscribe(_topic);
```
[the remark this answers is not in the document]
g7ed6e: @birojnayak Indeed this is a nice feature to consume message from multiple topics which each own a specific ser/des logic. In CoreWCF, i'm wondering if we need this because we are already able to "mix" different message types in the same topic. The Dispatch logic of CoreWCF is handling the routing of the message to the correct implementation. Is it something that has been implemented in MSMQ / RabbitMQ ?

src/CoreWCF.Kafka/tests/PoisonPillTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                .SetValueSerializer(Serializers.Utf8)
+                .Build();
+
+            List<string> expected = new(messageCount - 1);
+            int poisonPillIndex = new Random().Next(1, messageCount - 1);
+            for (int i = 0; i < messageCount; i++)
+            {
+                if (i == poisonPillIndex)
+                {
+                    string poisonPill = Guid.NewGuid().ToString();
```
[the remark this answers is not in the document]
g7ed6e: I think that in case of poison-pill we want either to do nothing or to send the message in a DLQ.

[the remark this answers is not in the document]
g7ed6e: I think that in case of poison-pill we want either to do nothing or to send the message in a DLQ.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportSecurity.cs
```diff
@@ -0,0 +1,132 @@
...
+    internal BindingElement CreateBindingElement()
+    {
+        KafkaSecurityBindingElement bindingElement = new();
+
+        if (_mode == KafkaSecurityMode.Transport)
+        {
+            if (_credentialType == KafkaCredentialType.None)
+            {
+                bindingElement.SecurityProtocol ??= SecurityProtocol.Ssl;
+                bindingElement.SslCaPem ??= CaPem;
```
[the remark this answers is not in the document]
g7ed6e: We want to keep the default values but allow one to provide values using a custom binding. We cannot validate all possible configuration before starting the consumer.

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaSecurityBindingElement.Config.cs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /// <summary>
+    /// <inheritdoc cref="ClientConfig.SaslKerberosServiceName"/>
+    /// </summary>
+    public string SaslKerberosServiceName
+    {
+        get => Config.SaslKerberosServiceName;
+        set => Config.SaslKerberosServiceName = value;
+    }
```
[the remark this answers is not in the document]
g7ed6e: We do support Kerberos through CustomBinding

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaSecurityBindingElement.Config.cs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// <summary>
+    /// <inheritdoc cref="ClientConfig.SaslKerberosPrincipal"/>
+    /// </summary>
+    public string SaslKerberosPrincipal
```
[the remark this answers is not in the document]
g7ed6e: We do support all Confluent.Kafka properties and values through CustomBinding

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/SaslMechanismHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+
+internal static class SaslMechanismHelper
+{
+    public static bool IsDefined(SaslMechanism value)
+    {
+        return value == SaslMechanism.Plain
+               || value == SaslMechanism.Gssapi
+               || value == SaslMechanism.ScramSha256
+               || value == SaslMechanism.ScramSha512
+               || value == SaslMechanism.OAuthBearer;
```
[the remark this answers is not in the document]
g7ed6e: Gssapi is supported through CustomBinding. If we want to support it through the KafkaBinding we should add integration tests and thus configure authentication in the compose file

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaTransportSecurity.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    public string Username { get; }
+    public string Password { get; }
+}
+
+public class ClientCertificateCredential
+{
+    public string CertificatePem { get; set; }
+    public string KeyPem { get; set; }
+    public string KeyPassword { get; set; }
```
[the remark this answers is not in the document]
g7ed6e: I've discussed the security configuration part with @mconnew and i will update the PR accordingly.

src/CoreWCF.Kafka/src/CoreWCF/Channels/SaslMechanismHelper.cs
g7ed6e: @birojnayak i kept this helper as is to allow one configure these sasl mechanisms at tranport binding level using a CustomBinding. 
However I introduced enum values of `KafkaCredentialType` that match all the possible SaslMechanism values so we can later expose these SaslMechanism values at the binding level. For the time being, at binding level, I'm throwing if one try to use Gssapi, Scram or OAuthBearer sasl mechanism. I think this should clarify what need to be done to add more SaslMechanism support at binding level.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportSecurity.cs
g7ed6e: @birojnayak I renamed these properties so they match the ones from Confluent.Kafka

src/CoreWCF.Kafka/tests/MessageEncodingTests.cs
```diff
@@ -0,0 +1,106 @@
...
+                services.AddServiceEndpoint<TestService, ITestContract>(new KafkaBinding
+                {
+                    DeliverySemantics = KafkaDeliverySemantics.AtMostOnce,
+                    GroupId = consumerGroupAccessor.Invoke()
+                }, $"net.kafka://localhost:9092/{topicNameAccessor.Invoke()}");
+            });
+        }
+    }
+
+    private class StartupBinaryEncoding
```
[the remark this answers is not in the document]
g7ed6e: Done! Thanks for the gist.

birojnayak:
Planning to review in this week

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaBinding.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public IsolationLevel? IsolationLevel
+        {
+            get => _transportBindingElement.IsolationLevel;
+            set => _transportBindingElement.IsolationLevel = value;
+        }
+
+        public KafkaTransportSecurity Security
+        {
```
mconnew, line 84: Generally the Security property on a binding would expose a general Security type, eg KafkaSecurity. That would then have two properties, Transport and Message, exposing a Transport specific class, and a Message specific class (which also gets used for TransportWithMessageCredentials).

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaBinding.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class KafkaBinding : Binding
+    {
+        private KafkaTransportSecurity _security = new();
+        private readonly KafkaTransportBindingElement _transportBindingElement = new();
+        private readonly BinaryMessageEncodingBindingElement _binaryMessageEncodingBindingElement = new();
+        private readonly TextMessageEncodingBindingElement _textMessageEncodingBindingElement = new();
+
+        public KafkaBinding(KafkaDeliverySemantics deliverySemantics = KafkaDeliverySemantics.AtLeastOnce)
+        {
```
mconnew, line 17: You mentioned about wanting to support TransportWithMessageCredentials. As that's the case, the general convention is for the binding constructor to take a SecurityMode value. You could have multiple constructors, one which takes just a SecurityMode, another which take the delivery semantics, and one which takes both. You can probably get that down to 2 overloads with using optional parameters.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaReceiveContext.cs
```diff
@@ -0,0 +1,71 @@
...
+    private readonly ConsumeResult<Null, byte[]> _consumeResult;
+    private readonly KafkaTransportPump _kafkaTransportPump;
+
+    public KafkaReceiveContext(ConsumeResult<Null, byte[]> consumeResult, KafkaTransportPump kafkaTransportPump)
+    {
+        _consumeResult = consumeResult;
+        _kafkaTransportPump = kafkaTransportPump;
+    }
+    protected override async Task OnAbandonAsync(CancellationToken token)
+    {
```
mconnew, line 22: nit: spacing

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportBindingElement.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        public override QueueTransportPump BuildQueueTransportPump(BindingContext context)
+        {
+            var serviceProvider = context.BindingParameters.Find<IServiceProvider>();
+            var serviceDispatcher = context.BindingParameters.Find<IServiceDispatcher>();
+            var logger = serviceProvider.GetRequiredService<ILogger<KafkaTransportPump>>();
+
+            return new KafkaTransportPump(this, logger, serviceDispatcher, DeliverySemantics);
+        }
```
mconnew, line 106: I'm not saying this is how you should do things, especially as only one parameter comes from DI, but an alternative would be to add KafkaTransportPump to DI then use `ActivatorUtilities.CreateInstance(serviceProvider, typeof(KafkaTransportPump), serviceDispatcher, DeliverySemantics)`.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
mconnew: My preference would be make sure we don't block ourselves in with the design and add this if people ask for it. In other words, make sure we wouldn't require anyone to change an existing usage if it was added later.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+    private CountdownEvent _receiveContextCountdownEvent;
+
+    private readonly Uri _baseAddress;
+    private CancellationTokenSource _cts;
+    private AsyncManualResetEvent _mres;
+    private static readonly (bool? EnableAutoCommit, bool? EnableAutoOffsetStore) s_atMostOnceConfigValues = (false, null);
+    private static readonly (bool? EnableAutoCommit, bool? EnableAutoOffsetStore) s_atLeastOncePerMessageCommitConfigValues = (false, null);
+    private static readonly (bool? EnableAutoCommit, bool? EnableAutoOffsetStore) s_atLeastOnceBatchCommitConfigValues = (true, false);
+    private static readonly Regex s_topicNameRegex =
+        new(@"^[a-zA-Z0-9\.\-_]{1,255}$", RegexOptions.Compiled, TimeSpan.FromMilliseconds(100));
```
mconnew, line 36: This can be converted to a source generated regex. See [https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators) for more info.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+                throw new NotSupportedException(SR.InvalidDeadLetterQueueTopicName);
+            }
+
+            if (!s_topicNameRegex.IsMatch(transportBindingElement.DeadLetterQueueTopic))
+            {
+                throw new NotSupportedException(string.Format(SR.InvalidTopicName, Topic));
+            }
+        }
+        TransportBindingElement = transportBindingElement;
+        _baseAddress = serviceDispatcher.BaseAddress;
```
mconnew, line 63: If you are going to keep this around, you could have problems if the passed in binding element is modified by poorly written code. You should clone the binding element that was passed in to protect against this.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+            Producer = new ProducerBuilder<Null, byte[]>(producerConfig)
+                .SetKeySerializer(Serializers.Null)
+                .SetValueSerializer(Serializers.ByteArray)
+                .Build();
+        }
+
+        Task.Run(async () =>
+        {
+            while (!ct.IsCancellationRequested)
+            {
```
mconnew, line 103: The lifetime validity of the CancellationToken passed in to StartPunpAsync is until the Task returned from StartPumpAsync completes. Here you are creating a background task using Task.Run which captures ct in the delegate, and ct is a token linked to the token passed to StartPumpAsync. After the Task returned from the method StartPumpAsync completes, cancelling the CancellationToken shouldn't have side effects, but it does.  
I don't think you need to link them. If you do link them, you should keep a reference to the CancellationTokenSource returned from calling CreateLinkedTokenSource as you need to Dispose it when you're done. Otherwise you can get a memory leak where it keeps itself alive. A bit like you can get when you add an EventHandler to an event and you don't remove it.

mconnew, line 103: This only has a concurrency of 1. This can become a bottleneck if the messages can be large. Is this intentional? A limitation of the Kafka consumer client?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+            _mres.Set();
+        }, ct);
+        return Task.CompletedTask;
+    }
+
+    public override async Task StopPumpAsync(CancellationToken token)
+    {
+        _cts.Cancel();
+        await _mres.WaitAsync(CancellationToken.None);
```
mconnew, line 149: This will throw a NRE if you call StopPumpAsync before StartPumpAsync.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+            _cts.Cancel();
+        }
+    }
+
+    private Task OnConsumeMessage(ConsumeResult<Null, byte[]> consumeResult,
+        QueueTransportContext queueTransportContext)
+    {
+        var receiveContext = new KafkaReceiveContext(consumeResult, this);
+        IncrementReceiveContextCount();
+        var context = new QueueMessageContext
```
mconnew, line 215: You are calling IncrementReceiveContextCount in KafkaTransportPump, but DecrementReceiveContextCount in KafkaReceiveContext. I prefer having the responsibility of increment and decrement be owned by a single class. What do you think about having the constructor of KafkaReceiveContext call the increment instead?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    public override async Task StopPumpAsync(CancellationToken token)
+    {
+        _cts.Cancel();
+        await _mres.WaitAsync(CancellationToken.None);
+        _cts.Dispose();
+        _receiveContextCountdownEvent.Signal();
+        _receiveContextCountdownEvent.Wait(CancellationToken.None);
+        _receiveContextCountdownEvent.Dispose();
```
mconnew, line 153: You should pass in the CancellationToken passed to StopPumpAsync otherwise you can block indefinitely. This will need a try/catch to catch OperationCanceledException, If the token gets cancelled before you are done, you need to do some kind of unclean shutdown. This means deciding what to do with in flight requests. Do you potentially put them in a DLQ? Do you keep track of in-flight requests and inform Kafka that they weren't completed, then if the executing thread does complete, the OnCompleteAsync/OnAbandonAsync know things were shut down and are a no-op? That's the approach used in multiple places in WCF.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        _receiveContextCountdownEvent.Signal();
+    }
+
+    public void Dispose()
+    {
+        if (TransportBindingElement.ErrorHandlingStrategy == KafkaErrorHandlingStrategy.DeadLetterQueue)
+        {
+            Producer.Dispose();
+        }
```
mconnew, line 241: If Dispose is called before StartPumpAsync, this will throw NRE. It might just be simpler and safer to call `Producer?.Dispose(); Producer = null;` You need not protect against multiple Dispose calls throwing. What happens if `Consumer.Dispose()` is called multiple times? Does it throw or silently ignore it? 
What happens if Dispose() is called before StopPumpAsync?

src/CoreWCF.Kafka/src/CoreWCF/Channels/Helpers/KafkaCredentialTypeHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Channels;
+
+internal static class KafkaCredentialTypeHelper
+{
+    public static bool IsDefined(KafkaCredentialType value)
+    {
+        return value == KafkaCredentialType.None
+               || value == KafkaCredentialType.SslKeyPairCertificate
```
mconnew, line 11: Are the values for KafkaCredentialType strictly incrementing by 1? E.g. the values are 0, 1, 2 etc. If so, a faster check is to do a bounds check on the integer value:  
```c#
return (int)value >= (int)KafkaCredentialType.None &&
       (int)value <= (int)KafkaCredentialType.SaslOAuthBearer;
```
Some of those casts might not be needed, but I put them in just to make sure.

g7ed6e: Resharper is suggesting this pattern matching saving one cast
```c#
(int)value is >= (int)KafkaCredentialType.None and <= (int)KafkaCredentialType.SaslOAuthBearer
```

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: CoreWCF.Kafka is targeting netstandard2.0 but it looks like this feature is only available starting from dotnet7.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaReceiveContext.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportBindingElement.cs
g7ed6e: Thanks @mconnew  i did not know ActivatorUtilities.CreateInstance (i however know Activator.CreateInstance).
Having a look to what is done for RabbitMQ it seems that RabbitMqTransportPump is added to DI through an extension method named AddServiceModelRabbitMqSupport which is never called and RabbitMqTransportBindingElement is manually newing the pump instance which mean that the service registration (and thus the extension method) is useless.
I think we should align both implementations what is your view about this ?

src/CoreWCF.Kafka/src/CoreWCF/Channels/Helpers/SaslMechanismHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+using Confluent.Kafka;
+
+namespace CoreWCF.Channels;
+
+internal static class SaslMechanismHelper
+{
+    public static bool IsDefined(SaslMechanism value)
+    {
+        return value == SaslMechanism.Plain
+               || value == SaslMechanism.Gssapi
```
birojnayak, line 13: Gssapi not supported ?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportBindingElement.cs
birojnayak: I personally like the extension method as this explicitly signifies you are using RabbitMQ... I will look at the RabbitMQ piece.. thanks for bring it up

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaSecurityBindingElement.Config.cs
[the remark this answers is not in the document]
birojnayak: I couldn't find test case, so was raising...

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/SaslMechanismHelper.cs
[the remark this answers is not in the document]
birojnayak: ok make sense... I don't want to block PR because we don't have gssapi test, so please create a back log issue to add test when you feel like.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
birojnayak: @g7ed6e the diff between RabbitMQ (or any Queue) vs notification(Kafka) is Kafka provides out of the box subscription to listen multiple topics while RabbitMQ client doesn't. I looked at your recent changes, I think below way we can extend the feature for multiple subscription.... We can expose another property with List<String> topics, and under startpump logic where background task is run can be spawned per topic. Or let me know what's your thought on extending ?  Bdw I am not asking to implement now (can be a future Feature/Back log issue), I am trying to answer @mconnew question about future extensibility without impacting previous version.

Also , can the background task to consume message in StartPumpAsync can be moved to another async method?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: Yes. Under the hood multiple threads are concurrently retrieving the messages. Please see https://docs.confluent.io/kafka-clients/dotnet/current/overview.html#the-consume-loop

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: fixed

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: done

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaSecurityBindingElement.Config.cs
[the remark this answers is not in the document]
g7ed6e: @birojnayak Indeed Kerberos is not tested at CoreWCF side as we are here just delegating to Confluent.Kafka.
The idea is :
- All Confluent properties settings are exposed through KafkaTransportBindingElement so a user can set them with a CustomBinding. In that case we do not have unit tests because these tests would be testing mappings. Basically this means all the features supported by Confluent.Kafka are supported by CoreWCF.Kafka.
- There are common security scenario (TLS + SASL authentication, mutual TLS..) for which we want to bring premium support by allowing configuration directly through KafkaBinding. For these scenario we do have integration tests.
This strategy should allow users that already know Kafka works to find what they need depending of their scenario.
For users who do not know how Kafka works are familiar with WCF security concepts, the KafkaBinding will take care to setup the correct security mechanism in Kafka world.
Adding @mconnew as i think the above approach should be shared.

src/CoreWCF.Kafka/tests/SecurityModeTests.cs
```diff
@@ -0,0 +1,389 @@
...
+using CoreWCF.Queue.Common.Configuration;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CoreWCF.Kafka.Tests;
+
+public class SecurityModeTests : IntegrationTest
```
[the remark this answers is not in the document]
g7ed6e: According to the approach explained above adding negative tests here would be adding all possible configuration combinations.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
```diff
@@ -0,0 +1,230 @@
...
+    private static readonly (bool? EnableAutoCommit, bool? EnableAutoOffsetStore) s_atLeastOnceBatchCommitConfigValues = (true, false);
+
+    public KafkaTransportPump(KafkaTransportBindingElement transportBindingElement,
+        KafkaSecurityBindingElement securityBindingElement,
+        ILogger<KafkaTransportPump> logger,
+        IServiceDispatcher serviceDispatcher, KafkaDeliverySemantics kafkaDeliverySemantics)
+    {
+        _logger = logger;
+        _kafkaDeliverySemantics = kafkaDeliverySemantics;
+        _topic = serviceDispatcher.BaseAddress.PathAndQuery.TrimStart('/');
```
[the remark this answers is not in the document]
g7ed6e: Thanks! i added topic name validation.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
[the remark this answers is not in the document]
g7ed6e: @birojnayak  Confluent.Kafka .net allows to subscribe to multiples topics by passing a regex to the `Consumer.Subscribe()` method. It allows to have one thread processing the consumer loop (under the hood librdkafka is using multiple threads) and dispatch messages (multithreaded if needed).
https://docs.confluent.io/platform/current/clients/confluent-kafka-dotnet/_site/api/Confluent.Kafka.IConsumer-2.html#Confluent_Kafka_IConsumer_2_Subscribe_System_String_
Please notice:
- in latest changes I made I added a regex to ensure we fail fast if topic name has special character or length>255. 
- the topic name is passed within the address
```c#
services.AddServiceEndpoint<TestService, ITestContract>(new KafkaBinding
                {
                    AutoOffsetReset = AutoOffsetReset.Earliest,
                    DeliverySemantics = KafkaDeliverySemantics.AtMostOnce,
                    GroupId = consumerGroupAccessor.Invoke(),
                    Security = new KafkaSecurity
                    {
                        Mode = KafkaSecurityMode.Transport,
                        Transport = new KafkaTransportSecurity
                        {
                            CredentialType = KafkaCredentialType.None,
                            CaPem = SslHelper.GetSslCaPem()
                        }
                    }
                }, $"net.kafka://localhost:9093/my-topic");
```
I think we can later add an enum property to the binding to let user choose between either a plain text or a regex.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaBinding.cs
g7ed6e: @mconnew I introduced `KafkaSecurity` and followed what is done in `BasicHttpBinding`. Please notice that despite I introduced `KafkaMessageSecurity` type and `KafkaSecurityMode.TransportWithMessageCredentials` and `KafkaSecurityMode.Message` I'm planning to add MessageSecurity support in a follow up PR.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaBinding.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportBindingElement.cs
g7ed6e: @birojnayak please see https://github.com/CoreWCF/CoreWCF/pull/1127

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: it should be ok now i added null guards and a lock

src/CoreWCF.Kafka/tests/SecurityModeTests.cs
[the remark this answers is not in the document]
birojnayak: non-blocker.. so resolving

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
birojnayak: thank you for this link, understood the background processing to fetch messages... but once messages are fetched aren't we consuming/processing one by one ? (via await OnConsumeMessage). I don't have the expertise on how kafka messages are processed around service, but if this a normal trend, I am fine. Let me know what you think.

birojnayak (approved):
https://github.com/CoreWCF/CoreWCF/pull/943/files#r1251195239 <= only concern from me after reviewing latest changes, otherwise looks ok. Thank you for working on this.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: Indeed we are processing the message one by one. With Kafka all consumers from a given consumer group are given offset of a topic. Each consumer is committing its offset (i.e something like ack in rabbitmq) so messages have to be consumed sequentially by each consumer. If one need to scale out because the consumer loop is too slow then that's need to add a consumer to the consumer group. In the dotnet sdk the exposed abstraction is consumer that's why the consume loop is always single threaded. (i've never seen a kafka dotnet consumer sample multithreaded to be honest). When scaling out the consumer group by adding a new consumer to the group please notice that the topic need to have at least one partition to distribute by consumer.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaTransportPump.cs
g7ed6e: @mconnew I wired the KafkaBinding.CloseTimeout to a CancellationTokenSource that I'm using to gracefully shutdown (the DLQ producer have to be flushed and the consumer have to be closed properly to exit the consumer group so the broker know how to distribute the remaining offsets).
The AsyncManualResetEvent _mres is responsible to ensure we properly stop the consumption and the CountdownEvent _receiveContextCountdownEvent is responsible to wait for already fetched messages and apply the appropriate handling.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaSecurity.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private KafkaSecurity(KafkaSecurityMode mode, KafkaTransportSecurity transportSecurity, KafkaMessageSecurity messageSecurity)
+    {
+        Fx.Assert(KafkaSecurityModeHelper.IsDefined(mode), $"Invalid SecurityMode value: {mode.ToString()}.");
+
+        if (!KafkaSecurityModeHelper.IsSupported(mode))
+        {
+            throw new NotSupportedException("Security mode not supported yet.");
+        }
```
mconnew, line 29: This should be in a string resource

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaSecurity.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            _mode = value;
+        }
+    }
+
+    public KafkaTransportSecurity Transport
+    {
+        get => _transportSecurity;
+        set => _transportSecurity = value ?? new KafkaTransportSecurity();
+    }
```
mconnew, line 60: This has the semantics of setting null sets it to defaults. This means the getter can return something different than was set. You could either throw for setting null, or in ApplySecurity do something like (_transportSecurity ?? new KafkaTransportSecurity()).ConfigureTransport....

src/CoreWCF.Kafka/src/CoreWCF/Channels/Helpers/SecurityProtocolHelper.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Confluent.Kafka;
+
+namespace CoreWCF.Channels;
+internal static class SecurityProtocolHelper
+{
```
mconnew, line 8: Nit: blank line between namespace and class

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
```diff
@@ -0,0 +1,180 @@
...
+        public void Send(Message message, TimeSpan timeout)
+        {
+            using CancellationTokenSource cts = new(timeout);
+            SendAsync(message, cts.Token).GetAwaiter().GetResult();
+        }
+
+        public IAsyncResult BeginSend(Message message, AsyncCallback callback, object state)
+        {
+            return SendAsync(message, CancellationToken.None).ToApm(callback, state);
+        }
```
mconnew, line 148: Don't use CancellationToken.None, create a token based off DefaultSendTimeout.

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public IAsyncResult BeginSend(Message message, TimeSpan timeout, AsyncCallback callback, object state)
+        {
+            return BeginSend(message, callback, state);
+        }
```
mconnew, line 153: You should use the passed in timeout. I think you should change your method signature on SendAsync to take a TimeSpan instead of CancellationToken. The reason is you need to create a CancellationTokenSource for the timeout, but as that creates a background timer, you need to dispose the CTS when the work is done. Passing that around APM methods is a pain and a lot of code. Instead you can create the CTS in SendAsync and dispose it in the finally block.

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaSecurity.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            if (!KafkaSecurityModeHelper.IsDefined(value))
+            {
+                throw new ArgumentOutOfRangeException(nameof(value));
+            }
+
+            if (!KafkaSecurityModeHelper.IsSupported(value))
+            {
+                throw new NotSupportedException("Security mode not supported yet.");
+            }
```
mconnew, line 45: String resource

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/Helpers/KafkaChannelHelpers.cs
```diff
@@ -0,0 +1,27 @@
...
+using Confluent.Kafka;
+
+namespace CoreWCF.ServiceModel.Channels;
+
+internal static class KafkaChannelHelpers
+{
+    public static Exception ConvertProduceException(ProduceException<Null, byte[]> produceException)
+    {
+        return new CommunicationException(
+            string.Format(CultureInfo.CurrentCulture,
```
mconnew, line 16: Is there any way to detect a timeout scenario. This would happen if the CancellationToken is cancelled. In that scenario, we would normally throw TimeoutException. It's the only exception we throw regularly which isn't derived from CommunicationException

src/CoreWCF.Queue/tests/InMemoryQueue/InMemoryQueueTransportPump.cs
```diff
@@ -70,7 +70,7 @@ public override Task StopPumpAsync(CancellationToken token)
     {
         _cts.Cancel();
         _cts.Dispose();
-        _mres.Wait(token);
+        _mres.Wait(CancellationToken.None);
         _mres.Dispose();
```
mconnew, line 74: Curious why you are passing in CancellationToken.None? This is also equivalent of calling Wait() with no parameters.

Issue #668, opened 2022-05-12
ServiceBehaviorAttribute -> UseSynchronizationContext missing


skurth:
In .NET-Framework WCF, there is a property on ServiceBehaviorAttribute named UseSynchronizationContext. When you set it to false, there is less blocking with multiple (long running) requests:

https://social.msdn.microsoft.com/Forums/vstudio/en-US/1873eb17-a711-4512-84be-463b74b087b5/wcf-long-running-task-blocks-all-other-calls?forum=wcf

https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.servicebehaviorattribute.usesynchronizationcontext?redirectedfrom=MSDN&view=netframework-4.8#System_ServiceModel_ServiceBehaviorAttribute_UseSynchronizationContext

Is this not an issue anymore in CoreWCF or can/will this feature be added?

mconnew:
We do have support for this in the implementation, but it looks like we didn't add it to this behavior to modify it. It's only an issue if you have a sync context set when you start the service. This generally only happens with a UI app. We'll use this issue to track adding that API back in.

bjornhellander:
I can have a go at this. Do you have any thoughts about what kind of tests you would like to see?

PR #983, opened 2023-01-14
Add property UseSynchronizationContext to class ServiceBehaviorAttribute

bjornhellander:
An attempt to handle #668.

First off, I am new to this project, so I am prepared for some re-work. Don't feel bad if you have a lot of comments :-)

I hope that I have understood the actual change needed, which in that case was trivial, but I am feeling unsure of how to best test this. I have provided a first attempt using nuget package _Xunit.StaFact_ to simulate a UI synchronization context, both for synchronous and asynchronous operations.

If the general test idea seems ok, please look extra carefully at _UseSynchronizationContextFalseWithSynchronizationContextTestAsync_ in both test classes, since the behavior differs between the synchronous and asynchronous cases. At that level of detail, I don't know what the expected behavior actually is. I have written the asserts based on a combination of what my gut says and what actually happens on my machine, so some asserts might be checking unwanted behavior. I can try to replicate the tests using the old WCF if needed, but have not done so yet.

I also want to say that I have seen some old tests failing occasionally in Visual Studio due to "address already in use"-problems when running all tests. Since I am new here I don't know if that happens or if I have done something bad in this PR.

### DIFF
--- a/src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
+++ b/src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
@@ -1,0 +1,211 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.ServiceModel.Channels;
+using System.Threading;
+using System.Threading.Tasks;
+using CoreWCF.Configuration;
+using Helpers;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CoreWCF.NetTcp.Tests
+{
+    /// <summary>
+    /// Verifies the effect of <see cref="ServiceBehaviorAttribute.UseSynchronizationContext"/> on asynchronous operations.
+    /// </summary>
+    public class ServiceBehaviorUseSynchronizationContextAsyncTests
+    {
+        public const string RelativeAddress = "/nettcp.svc/UseSynchronizationContextAsync";
+        private readonly ITestOutputHelper _output;
+
+        public ServiceBehaviorUseSynchronizationContextAsyncTests(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        /// <summary>
+        /// This test exists only to demonstrate what the UIFact attribute does.
+        /// </summary>
+        [UIFact]
+        public async Task UIFactTestAsync()
+        {
+            int originalThread = Environment.CurrentManagedThreadId;
+            await Task.Yield();
+            Assert.Equal(originalThread, Environment.CurrentManagedThreadId);
+        }
+
+        [UIFact]
+        public async Task UseSynchronizationContextTrueWithSynchronizationContextTestAsync()
+        {
+            SynchronizationContext orgSynchronizationContext = SynchronizationContext.Current;
+            Assert.NotNull(orgSynchronizationContext); // Set by the UIFact attribute
+            int orgManagedThreadId = Environment.CurrentManagedThreadId;
+
+            AsyncServiceWithUseSynchronizationContextTrue serviceInstance =
+                await RunAsync<AsyncServiceWithUseSynchronizationContextTrue>();
+
+            Assert.Same(orgSynchronizationContext, serviceInstance.SynchronizationContextWhenCreated);
+            Assert.Same(orgSynchronizationContext, serviceInstance.SynchronizationContextWhenCalled);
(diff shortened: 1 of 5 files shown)


src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        public const string RelativeAddress = "/nettcp.svc/UseSynchronizationContextAsync";
+        private readonly ITestOutputHelper _output;
+
+        public ServiceBehaviorUseSynchronizationContextAsyncTests(ITestOutputHelper output)
+        {
+            _output = output;
+
+            TaskService.s_synchronizationContextWhenCreated = null;
+            TaskService.s_synchronizationContextWhenCalled = null;
```
mconnew, line 33: Passing data around via statics isn't a good idea as you can have problems with concurrent test execution. You could solve this by using a singleton in DI.  
1. Change the method signature of RunAsync to `private async Task<TService> RunAsync<TService>()`  
2. Add a singleton TService to DI in `Startup<TService>.ConfigureServices` : `services.AddSingleton<TService>();`  
3. Request the TService instance from DI in RunAsync after the test has run : `TService serviceInstance = host.Services.GetRequiredService<TService>();` and return it as the last statement in the using block.  

Other than the use of statics, the tests in this class look pretty thorough.

src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        [Fact]
+        public async Task UseSynchronizationContextTrueWithoutSynchronizationContextTestAsync()
+        {
+            await RunWithoutSynchronizationContextAsync(async () =>
+            {
+                Assert.Null(SynchronizationContext.Current);
+                int orgManagedThreadId = Environment.CurrentManagedThreadId;
+
```
mconnew, line 72: You aren't using orgManagerThreadId in the test validation. Is there missing validation or was this line left in place by accident?

src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        [Fact]
+        public async Task UseSynchronizationContextFalseWithoutSynchronizationContextTestAsync()
+        {
+            await RunWithoutSynchronizationContextAsync(async () =>
+            {
+                Assert.Null(SynchronizationContext.Current);
+                int orgManagedThreadId = Environment.CurrentManagedThreadId;
+
```
mconnew, line 98: Same comment as above about getting the current thread id.

mconnew:
@bjornhellander, if you can fix the use of statics and resolve the managed thread id being fetched in those two tests, I'll merge this.

src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
bjornhellander: Yes, I see what you mean. Didn't think of that. But this would also mean that the service implementations would need to access the singleton and that would require registering the singleton in all tests that use EchoService and ChatService, right? These two existed before and I just updated them a bit to store the context and thread id to make it possible to add the asserts in the new tests. I would prefer to add dedicated service implementations just for these tests, to not clutter all other tests with irrelevant registrations. Does that feel right?

mconnew: It has a default constructor so the non-DI mechanism will still work elsewhere. WCF on its own can instantiate a type with a default constructor with no DI injection. You would only break other tests if you added a required construction parameter. But I do agree with adding a custom type just for this test aa it's cleaner.

src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
bjornhellander: I had a validation there before and forgot to remove the variable. I added a comment now on why it wasn't ok, in case it helps someone in the future.

src/CoreWCF.NetTcp/tests/ServiceBehaviorUseSynchronizationContextAsyncTests.cs
bjornhellander: Same mistake here

bjornhellander:
Updated, @mconnew

PR #1006, opened 2023-02-22
Fix test that's timing out on .NET Framework

mconnew:
AsyncNetAdoptionTests.OneWayPatternTest_Parallel is timing out on .NET Framework. Under the hood on .NET Framework, HttpWebRequest uses ServicePointManager to manage connection pooling. The default number of simultaneous connections to a single endpoint is 2, which means the test which is executing 1000 requests in parallel can only be concurrently sending 2 at a time. This causes the test to fail on .NET Framework due to some of the requests taking too long to obtain a free connection in the pool. This should make the test stable on .NET Framework now.

### DIFF
--- a/src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
+++ b/src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
@@ -45,6 +45,9 @@ namespace BasicHttp
         [InlineData(1000)]
         public async Task OneWayPatternTest_Parallel(int callCount)
         {
+            // Does nothing on .NET [Core], but enables .NET Framework to have sufficient concurrency
+            // to make all the requests
+            System.Net.ServicePointManager.DefaultConnectionLimit = callCount;
             var inputs = Enumerable.Range(0, callCount).Select(x => x.ToString()).ToArray();
             var host = ServiceHelper.CreateWebHostBuilder<AsyncNetAdoptionOneWayServiceStartup>(_output).Build();
             using (host)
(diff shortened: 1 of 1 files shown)


mconnew:
@g7ed6e, can you review. Let me know if there's more context that you need.

src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
```diff
@@ -45,6 +45,9 @@ public async Task OneWayPatternTest()
         [InlineData(1000)]
         public async Task OneWayPatternTest_Parallel(int callCount)
         {
+            // Does nothing on .NET [Core], but enables .NET Framework to have sufficient concurrency
+            // to make all the requests
+            System.Net.ServicePointManager.DefaultConnectionLimit = callCount;
```
MeikTranel, line 50: Should probably reset this after the test otherwise you might run into problems where you implicitly fix something in another test with a largely unreasonable value for allowed concurrent connections.

g7ed6e (approved):
LGTM. As @MeikTranel mentioned the initial property value could be restored.

src/CoreWCF.Http/tests/AsyncNetAdoptionTests.cs
mconnew: You're correct, that would be best practice, I'll fix it. I don't think it's something worth being concerned about in the CoreWCF case though as this only applies for .NET Framework and not .NET [Core]. On .NET the SocketsHttpHandler doesn't limit the connections to 2 like HttpWebRequest does so the WCF client won't hit this problem when running tests on .NET 6/.NET 7 so an implicitly fixed test won't see any failures on .NET 6/.NET 7 anyway.

mconnew: I just realized there is a risk if there's another test running concurrently which changes this value between this test changing it and before the first request (when the value is snapshotted by ServicePointManager when it creates the ServicePoint). We don't currently run our tests concurrently, but that's something I do wish to fix at some point to speed up the test run. As this is the only test doing this, it's not a problem, but if we add other tests doing this and start executing tests concurrently, we could hit a problem. Still going to revert it here though as it's a potential future problem and not a current one.

PR #948, opened 2022-12-18
Add scoped AuthorizationHandler<T> support

g7ed6e:
Fixes #947

PR #1012, opened 2023-02-26
Drop useless null check added in #948

g7ed6e:


mconnew (approved):

mconnew:
Can you squash these changes before merging?

PR #1018, opened 2023-03-01
Fix ReceiveContext AbandonAsync / CompleteAsync never called

g7ed6e:


g7ed6e:
@mconnew Can you review this please ?
I still have concern with handling a message that is dispatched successfully but service implementation throws.
In this case, `ImmutableDispatchRuntime` is calling `RequestContext.ReplayAsync(null)` which causes `ReceiveContext.CompleteAsync` to be called before the operation is invoked. This behavior leads to a situation where the actual `QueueTransportPump` implementation is not able to know if invocation threw or not.

src/CoreWCF.Queue/tests/InMemoryQueue/InMemoryQueueTransportPump.cs
```diff
@@ -0,0 +1,75 @@
...
+        _queue = queue;
+        _receiveContextInterceptor = receiveContextInterceptor;
+    }
+
+    public override Task StartPumpAsync(QueueTransportContext queueTransportContext, CancellationToken token)
+    {
+        _cts = new CancellationTokenSource();
+        _mres = new ManualResetEventSlim();
+        CancellationToken ct = CancellationTokenSource.CreateLinkedTokenSource(token, _cts.Token).Token;
+
```
mconnew, line 32: Linked cancelltion token sources should be disposed as they call token.Register on each of the passed tokens. This can create a memory leak.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs
```diff
@@ -478,8 +478,25 @@ internal DispatchOperationRuntime GetOperation(ref Message message)
             return _demuxer.GetOperation(ref message);
         }
 
+        private void ReceiveContextRPCFacet_CreatIfRequired_Shim(MessageRpc rpc)
+        {
+            // see https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/Dispatcher/ReceiveContextRPCFacet.cs,25
+            ReceiveContext receiveContext = null;
+            if (!ReceiveContext.TryGet(rpc.Request, out receiveContext))
+            {
```
mconnew, line 486: We shouldn't be throwing if the receive context is missing. That just means the transport for this request doesn't use it. This is what's causing the tests to fail. If there isn't one, then anything involving the receive context becomes a no-op

g7ed6e: done. Thanks!

src/CoreWCF.Queue/tests/InMemoryQueue/InMemoryQueueTransportPump.cs
g7ed6e: @mconnew I made the QueuePollingService (which is an IHostedService thus disposed by the host) disposable so we can support disposable QueueTransportPump implementation.

mconnew:
> @mconnew Can you review this please ? I still have concern with handling a message that is dispatched successfully but service implementation throws. In this case, `ImmutableDispatchRuntime` is calling `RequestContext.ReplayAsync(null)` which causes `ReceiveContext.CompleteAsync` to be called before the operation is invoked. This behavior leads to a situation where the actual `QueueTransportPump` implementation is not able to know if invocation threw or not.

Does CoreWCF throw if you don't have an operation marked as OneWay? If it's limited to only using OneWay methods, then ReplyAsync should do nothing if it gets passed a null.  

Other than that being resolved, everything else looks good. I feel like that can be deferred for now to unblock you.

mconnew (approved):

PR #1022, opened 2023-03-05
Add WCF client for RabbitMQ implementation

jonlouie:
* Add RabbitMQ client similar to implementations for other bindings in dotnet/WCF
* In RabbitMQ Integration tests, replace the calls to RabbitMQ library with RabbitMQ WCF client


src/CoreWCF.RabbitMQ.Client/src/CoreWCF.RabbitMQ.Client.csproj
```diff
@@ -0,0 +1,20 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>CoreWCF.RabbitMQ.Client</AssemblyName>
+    <PackageId>CoreWCF.RabbitMQ.Client</PackageId>
+    <RootNamespace />
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <ItemGroup>
```
g7ed6e, line 11: @jonlouie I think we can avoid taking a dependency on the server side part and the common CoreWCF code.
Please have a look there to exclude common code.
https://github.com/CoreWCF/CoreWCF/pull/943/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16e

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            get => _transport.VirtualHost;
+            set => _transport.VirtualHost = value;
+        }
+
+        /// <summary>
+        /// Credentials to access the RabbitMQ host
+        /// </summary>
+        public ICredentials Credentials
+        {
```
mconnew, line 79: Credentials should probably come form the ClientCredentials. In the relevant place in the ChannelFactory code I'll provide a link to some sample code showing how to pull the credentials from the ClientCredentials class

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,113 @@
...
+            set => _transport.Credentials = value;
+        }
+
+        /// <summary>
+        /// Gets the scheme used by the binding
+        /// </summary>
+        public override string Scheme => _transport.Scheme;
+
+        public RabbitMqMessageEncoding MessageEncoding { get; set; } = RabbitMqMessageEncoding.Text;
+
```
mconnew, line 90: Need to validate on the setter that a valid value is being set. Any int can be cast to an enum. I could do `rabbitMqBinding.MessageEncoding = (RabbitMqMessageEncoding)27;` and the value would be set to something which doesn't make sense. Take a look at the code [here](https://source.dot.net/#System.ServiceModel.Http/System/ServiceModel/BasicHttpMessageSecurity.cs,25) to see a typical pattern.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+        private RabbitMqTransportBindingElement _transport;
+
+        internal RabbitMqChannelFactory(RabbitMqTransportBindingElement transport, BindingContext context)
+            : base(context.Binding)
+        {
+            _transport = transport;
+            _bufferManager = BufferManager.CreateBufferManager(transport.MaxBufferPoolSize, int.MaxValue);
+
+            var messageEncoderBindingElements = context.BindingParameters.OfType<MessageEncodingBindingElement>().ToList();
+            if (messageEncoderBindingElements.Count > 1)
```
mconnew, line 21: Did you copy this code from somewhere as I want to check if it needs to be fixed where you copied it from? The more efficient way to do this is to call `context.BindingParameters.Find<MessageEncodingBindingElement>()` and do a null check. You could also use FindAll which returns a collection and do the count check.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                throw new InvalidOperationException("More than one MessageEncodingBindingElement was found in the BindingParameters of the BindingContext");
+            }
+
+            if (messageEncoderBindingElements.Count == 1)
+            {
+                _messageEncoderFactory = messageEncoderBindingElements.First().CreateMessageEncoderFactory();
+            }
```
mconnew, line 29: Although this will go away with my previous comment, you did call `ToList()` so you don't need to use `First()`, you know there's a single item in there so could use `messageEncoderBindingElements[0]` instead.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System.Linq;
+using System;
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+
+namespace CoreWCF.ServiceModel.Channels
+{
+    public class RabbitMqChannelFactory : ChannelFactoryBase<IOutputChannel>
+    {
```
mconnew, line 9: Make this class internal

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        protected override void OnOpen(TimeSpan timeout)
+        {
+        }
+
+        protected override IAsyncResult OnBeginOpen(TimeSpan timeout, AsyncCallback callback, object state)
+        {
+            return Task.CompletedTask;
+        }
```
mconnew, line 65: `Task.CompletedSynchronously` always returns false, which tells the caller that the callback and state will get called by the completion of OnBeginOpen, which has already happened as `Task.IsCompleted` returns true. You would need to do a null check on the callback and call it passing the `IAsyncResult` that you are returning. The problem with that is the `IAsyncResult` you return needs to wrap the `state` object you passed in, and `Task.CompeltedTask` doesn't do that. All this is to say there are no shortcuts when it comes to implementing APM methods (which is why it was thankfully replaced with async/await) so you're best off copying in the `TaskHelpers.ToApm` and `TaskHelpers.ToApmEnd` helper methods and using them to make sure everything is done right.  
I noticed you didn't override OnBeginClose/OnEndClose, so I think ChannelFactoryBase handles them. Do you need to override the Open methods too? I think you might be able to just remove these methods.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+namespace CoreWCF.ServiceModel.Channels
+{
+    public class RabbitMqChannelFactory : ChannelFactoryBase<IOutputChannel>
+    {
+        private BufferManager _bufferManager;
+        private MessageEncoderFactory _messageEncoderFactory;
+        private RabbitMqTransportBindingElement _transport;
+
+        internal RabbitMqChannelFactory(RabbitMqTransportBindingElement transport, BindingContext context)
+            : base(context.Binding)
```
mconnew, line 15: You should get the client credentials in the constructor of this ChannelFactory. You can see the code to do this [here](https://source.dot.net/#System.ServiceModel.Http/System/ServiceModel/Channels/HttpChannelFactory.cs,129) from HttpChannelFactory. You then get the SecurityTokenManager like [here](https://source.dot.net/#System.ServiceModel.Http/System/ServiceModel/Channels/HttpChannelFactory.cs,557) and then get the token provider for the credentials. I suggest using ClientCredentials.UserName as the property to get the credentials from. If you do that, you can see the call to a helper method that Http uses for the UserName credentials [here](https://source.dot.net/#System.ServiceModel.Http/System/ServiceModel/Channels/HttpChannelFactory.cs,461). The implementation is [here](https://source.dot.net/#System.ServiceModel.Primitives/System/ServiceModel/Channels/TransportSecurityHelpers.cs,170). You generally make the call to SecurityTokenManager.CreateSecurityTokenProvider during the call to opening the channel as there a ChannelParameterCollection that can be passed like this:
```c#
var channel = factory.CreateChannel();
var channelParameters = ((IChannel)channel).GetProperty<ChannelParameterCollection>();
channelParameters.Add(new Foo());
((IChannel)channel).Open(); // Causes the security token provider to be created using the channel parameters
```
This means the actual IOutputChannel is the one that needs to get the token provider, and the security token on the open call path..

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+        protected override void OnEndOpen(IAsyncResult result)
+        {
+        }
+
+        protected override IOutputChannel OnCreateChannel(System.ServiceModel.EndpointAddress queueUrl, Uri via)
+        {
+            var connectionSettings = _transport.GetConnectionSettings();
+            var factory = connectionSettings.GetConnectionFactory();
+            var connection = factory.CreateConnection();
+            var rabbitMqClient = connection.CreateModel();
```
mconnew, line 76: This initialization code needs to happen in RabbitMqOutputChannel.OnBeginOpen and OnOpen (make a private OnOpenAsync and use ToApm/ToApm end in OnBeginOpen/OnEndOpen) as you need to get the SecurityTokenProvider and optionally open it.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        public void Send(System.ServiceModel.Channels.Message message, TimeSpan timeout)
+        {
+            Send(message);
+        }
+
+        public IAsyncResult BeginSend(System.ServiceModel.Channels.Message message, AsyncCallback callback, object state)
+        { return Task.CompletedTask; }
```
mconnew, line 125: BeginSend is what gets called if your contract is async (eg returns `Task<Foo>`) and this is a no-op. Make a private SendAsync and call that from BeginSend/EndSend using ToApm and ToApmEnd.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        public void Send(System.ServiceModel.Channels.Message message)
+        {
+            var messageBuffer = EncodeMessage(message);
+
+            try
+            {
+                _rabbitMqClient.BasicPublish(
+                    exchange: _connectionSettings.Exchange,
```
mconnew, line 108: Check IModel.IsOpen before trying to send. What conditions can the RabbitMqClient end up not open? If it's something that can only be manually done, then you should Abort() the channel and throw something that derives from CommunicationException (or that exception itself).

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+            finally
+            {
+                // Make sure buffers are always returned to the BufferManager
+                _parent.BufferManager.ReturnBuffer(messageBuffer.Array);
+            }
+        }
+
+        public void Send(System.ServiceModel.Channels.Message message, TimeSpan timeout)
+        {
```
mconnew, line 120: Should we optionally (if queue supports it) be using WaitForConfirms with this timeout value?

mconnew:
>     <value>No IPEndpoints were found for host {0}.</value>

Can you limit this to just the string resources you need?

---
Refers to: src/CoreWCF.RabbitMQ.Client/src/Resources/Strings.resx:121 in a15d92a. [](commit_id = a15d92a9ae71aacdedf12ddb782a89216828d2c7, deletion_comment = False)

src/CoreWCF.RabbitMQ.Client/src/CoreWCF.RabbitMQ.Client.csproj
jonlouie: The client takes a dependency on the server side project because they share the RabbitMqConnectionSettings class (both server and client use it to handle endpoint URIs and create connections to a RabbitMQ host).

I also wanted to keep client separate from server, but besides duplicating the class, no other immediate solutions came to mind.

g7ed6e: I see. But this dependency leads to:
CoreWCF.RabbitMQ.Client => CoreWCF.RabbitMQ => CoreWCF.Primitives => ASP.NET Core 2.1
You could either duplicate code (this is the solution i've chosen for the Kafka implementation) or move the common RabbitMQ code to the client package and make the server take a dependency on the client. this would avoid clients to pull ASP.NET Core.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
jonlouie: I no longer have the source, but it was from another WCF Client, outside of dotnet/wcf. BindingParameters here is of type `System.ServiceModel.Channels.BindingParameterCollection` which does not seem to have the `Find()` or `FindAll()` methods.

mconnew: [BindingParameterCollection](https://github.com/dotnet/wcf/blob/main/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs#L1013) derives from KeyedByTypeCollection\<object\> which has a [Find\<T\>() method](https://github.com/dotnet/wcf/blob/main/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs#L54).

jonlouie: I might be missing something, but it looks like the change to have `BindingParameterCollection` derive from `KeyedByTypeCollection` wasn't released yet. We're referencing the latest System.ServiceModel.Primitives pkg (v4.10, released Aug 17, 2022), and BindingParameterCollection still derives from `KeyedCollection`. Looks closer to [this version](https://github.com/dotnet/wcf/blob/3f393620ec22b553dff47f999068d64db8ab39f1/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs#L992).

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
```diff
@@ -28,13 +28,31 @@ public IntegrationTests(ITestOutputHelper output)
         }
 
         [Fact(Skip = "Requires RabbitMQ host with SSL")]
-        public void ClassicQueueWithTls_ReceiveMessage_Success()
+        public void ClassicQueueWithTls_SendReceiveMessage_Success()
         {
```
mconnew, line 32: You should probably add the Trait for LinuxOnly on here too. Is the reason for lack of SSL because the container doesn't come with SSL configured? If so, can you open an issue to enable this scenario?

mconnew: This got me thinking, is there a mechanism to provide a certificate validation callback to the RabbitMQ client library? We can easily generate a certificate for the server, but if we don't install the CA cert on the test machine, the certificate will be rejected. If there is a certificate validation callback mechanism, we should get the certificate validator from the ClientCredentials and hook it up.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,117 @@
...
+        private RabbitMqTransportBindingElement _transport;
+        private TextMessageEncodingBindingElement _textMessageEncodingBindingElement;
+        private BinaryMessageEncodingBindingElement _binaryMessageEncodingBindingElement;
+        private RabbitMqMessageEncoding _messageEncoding = RabbitMqMessageEncoding.Text;
+
+        public RabbitMqBinding(string uri) : this(new Uri(uri))
+        { }
+
+        public RabbitMqBinding(Uri uri)
+        {
```
mconnew, line 22: You don't normally pass the Uri to the binding. You pass it to the ChannelFactory constructor and you get told the Uri in RabbitMqChannelFactory.OnCreateChannel

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        private void InitializeSecurityTokenManager(BindingContext context)
+        {
+            _channelCredentials = context.BindingParameters.OfType<SecurityCredentialsManager>().FirstOrDefault();
+            _channelCredentials ??= new ClientCredentials();
+            _securityTokenManager = _channelCredentials.CreateSecurityTokenManager();
+        }
+
+        public RabbitMqTransportBindingElement Transport => _transport;
+
```
mconnew, line 41: You don't need to define a field. If you have a getter only property, you are allowed to assign it in a constructor.

mconnew: The same applies to BufferManager and MessageEncoderFactory

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqChannelFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+        protected override void OnEndOpen(IAsyncResult result)
+        {
+            result.ToApmEnd();
+        }
+
+        protected override IOutputChannel OnCreateChannel(System.ServiceModel.EndpointAddress queueUrl, Uri via)
+        {
+            return new RabbitMqOutputChannel(
+                this,
```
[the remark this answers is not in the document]
mconnew: You need to pass the remote address (EndpointAddress) and via. If someone has used ChannelFactory.CreateChannel() or ChannelFactory.CreateChannel(endpointAddress), then the remote address and the via Uri will be the same. You should use the via address for the actual communication, and the EndpointAddress is used the for To header in the Message.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqDefaults.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace CoreWCF.ServiceModel.Channels
+{
+    internal static class RabbitMqDefaults
+    {
+        internal const string Scheme = "amqp";
+
+        private static readonly System.ServiceModel.Channels.MessageEncoderFactory _messageEncoderFactory;
+        static RabbitMqDefaults()
+        {
```
mconnew, line 9: Nit: blank line before the static constructor

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,284 @@
...
+            return base.GetProperty<T>();
+        }
+
+        /// <summary>
+        /// Open the channel for use. We do not have any blocking work to perform so this is a no-op
+        /// </summary>
+        protected override void OnOpen(TimeSpan timeout)
+        {
+            OnOpenAsync(timeout).Wait();
+        }
```
mconnew, line 84: If OnOpenAsync throws, calling Wait() will throw an AggregateException with the real exception inside of it. You need to use GetAwaiter().GetResult() as that will throw the real exception.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,284 @@
...
+            var token = await securityTokenProvider?.GetTokenAsync(timeout);
+            return token;
+        }
+
+        private SecurityTokenProvider GetSecurityTokenProvider()
+        {
+            if (_securityTokenManager == null)
+            {
+                return null;
+            }
```
mconnew, line 256: Throw an exception here as it should always not be null.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        /// <summary>
+        /// Published a Message to RabbitMQ.
+        /// </summary>
+        /// <exception cref="TimeoutException"></exception>
+        private Task SendAsync(Message message)
+        {
+            return SendAsync(message, TimeSpan.Zero);
+        }
```
mconnew, line 194: TimeSpan.Zero generally means execute the functionality you are asking for, but don't wait if it doesn't complete immediately. With many api's, passing Zero will throw. The value you want to pass is `System.Threading.Timeout.InfiniteTimeSpan`

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,284 @@
...
+                    Abort();
+                    throw new CommunicationException(SR.RabbitMqClientNotOpen);
+                }
+
+                _rabbitMqClient.BasicPublish(
+                    exchange: _connectionSettings.Exchange,
+                    routingKey: _connectionSettings.RoutingKey,
+                    body: messageBuffer);
+                if (timeout.Ticks > 0)
+                {
```
mconnew, line 221: You don't need to get the ticks. You can just do `timeout > TimeSpan.Zero`

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
mconnew: The other uses of Wait() in Send, Close etc also need changing.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqTransportBindingElement.cs
```diff
@@ -0,0 +1,114 @@
...
+        public override string Scheme
+        {
+            get { return RabbitMqDefaults.Scheme; }
+        }
+
+        /// <summary>
+        /// The largest receivable encoded message
+        /// </summary>
+        public override long MaxReceivedMessageSize
+        {
```
mconnew, line 42: You don't need this property here. The property exists on the service side so you can limit resource allocation to prevent malicious clients. You don't need to worry about that on the client side. You can still have a check when serializing as you know it will fail if it's bigger than 512MB as RabbitMQ can't accept messages bigger than that, but you don't need to allow it to be configurable on the client side. If you want to allow it to be configurable, then you should drop Received from the name as you aren't receiving anything.

src/CoreWCF.RabbitMQ.Client/src/Resources/Strings.resx
```diff
@@ -0,0 +1,177 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="MaxReceivedMessageSizeExceeded" xml:space="preserve">
+    <value>The maximum message size quota for incoming messages ({0}) has been exceeded. To increase the quota, use the MaxReceivedMessageSize property on the appropriate binding element.</value>
+  </data>
+  <data name="MaxSentMessageSizeExceeded" xml:space="preserve">
+    <value>The maximum message size quota for outgoing messages ({0}) has been exceeded.</value>
```
mconnew, line 124: I'm not finding any reference to this string resource

templates/TestStage.yml
```diff
@@ -79,19 +85,39 @@ stages:
...
         projects: ${{ parameters.testProjects}}
         arguments: '--configuration Release --framework $(targetFramework)'
 
+    - task: DotNetCoreCLI@2
+      displayName: Build test projects targeting .NET Framework
+      condition: eq(variables.targetFramework, 'net472')
+      inputs:
+        command: 'build'
+        projects: ${{ parameters.testProjectsTargetingNetFramework}}
+        arguments: '--configuration Release --framework $(targetFramework)'
```
mconnew, line 100: Instead of having this split in the build system, could you instead have the test project still have all the same target frameworks but conditionally exclude all source files if targetting .NET Framework, and have a condition on the project references too. So basically it becomes an empty project on .NET Framework with no tests.

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
jonlouie: 1. Added the LinuxOnly trait
2. Yes, my understanding is that we will need to provide a cert and mount the volume containing the cert ([example](https://gitlab.com/hito20/rabbitmq-with-tls-in-docker/-/blob/393d06135c6ed09b6d457b207fe34b9de0e157e2/docker-compose.yml)). 
3. Yes, the RabbitMQ client does have the validation callback mechanism defined [here](https://rabbitmq.github.io/rabbitmq-dotnet-client/api/RabbitMQ.Client.SslOption.html#RabbitMQ_Client_SslOption_CertificateSelectionCallback).

jonlouie: Issue created here: https://github.com/CoreWCF/CoreWCF/issues/1047

templates/TestStage.yml
jonlouie: Made the changes and tests pass in CI/CD, but I'm seeing strange behavior when building locally with Visual Studio. There are no build errors reported, but it still shows the `CoreWCF.RabbitMQ.Client` as failing to build (see printout below). However, when building the project from the command line, it succeeds for each Target Framework.

```
Build started...
1>------ Build started: Project: CoreWCF.RabbitMQ.Client, Configuration: Debug Any CPU ------
1>Reading resource file D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src/Resources/Strings.resx
1>Writing resources source file D:\Workplace\jonlouie\CoreWCF\bin/obj//CoreWCF.RabbitMQ.Client/CoreWCF.RabbitMQ.Client/SR.cs
1>Writing 19 resource strings
1>D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src\CoreWCF\ServiceModel\Channels\RabbitMqOutputChannel.cs(227,50,227,51): warning CS0168: The variable 'e' is declared but never used
1>D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src\CoreWCF\ServiceModel\Channels\RabbitMqTransportBindingElement.cs(11,22,11,45): warning CS0414: The field 'RabbitMqTransportBindingElement._maxReceivedMessageSize' is assigned but its value is never used
1>CoreWCF.RabbitMQ.Client -> D:\Workplace\jonlouie\CoreWCF\bin\Debug\CoreWCF.RabbitMQ.Client\net7.0\CoreWCF.RabbitMQ.Client.dll
1>Done building project "CoreWCF.RabbitMQ.Client.csproj".
1>Reading resource file D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src/Resources/Strings.resx
1>Writing resources source file D:\Workplace\jonlouie\CoreWCF\bin/obj//CoreWCF.RabbitMQ.Client/CoreWCF.RabbitMQ.Client/SR.cs
1>Writing 19 resource strings
1>Reading resource file D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src/Resources/Strings.resx
1>Writing resources source file D:\Workplace\jonlouie\CoreWCF\bin/obj//CoreWCF.RabbitMQ.Client/CoreWCF.RabbitMQ.Client/SR.cs
1>Writing 19 resource strings
1>D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src\CoreWCF\ServiceModel\Channels\RabbitMqOutputChannel.cs(227,50,227,51): warning CS0168: The variable 'e' is declared but never used
1>D:\Workplace\jonlouie\CoreWCF\src\CoreWCF.RabbitMQ.Client\src\CoreWCF\ServiceModel\Channels\RabbitMqTransportBindingElement.cs(11,22,11,45): warning CS0414: The field 'RabbitMqTransportBindingElement._maxReceivedMessageSize' is assigned but its value is never used
1>CoreWCF.RabbitMQ.Client -> D:\Workplace\jonlouie\CoreWCF\bin\Debug\CoreWCF.RabbitMQ.Client\net472\CoreWCF.RabbitMQ.Client.dll
1>Done building project "CoreWCF.RabbitMQ.Client.csproj".
1>CoreWCF.RabbitMQ.Client -> D:\Workplace\jonlouie\CoreWCF\bin\Debug\CoreWCF.RabbitMQ.Client\net6.0\CoreWCF.RabbitMQ.Client.dll
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
========== Elapsed 00:05.264 ==========
```

When building the project from the command line, it succeeds for each Target Framework.

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
mconnew: The remote certificate validation configuration from the ClientCredentials needs to be wired up. The code for [getting the validator is here](BufferManagerOutputStream). The code to convert this to a delegate which matches RemoteCertificateValidationCallback can be found [here](AddServerCertIdentityValidation). You can just copy that method removing some of the tracing code which you don't need.

templates/TestStage.yml
mconnew: I think we might have some build warnings configured to be treated as errors in the .editorconfig file which don't apply when compiling at the command line.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF.RabbitMQ.Client.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <AssemblyName>CoreWCF.RabbitMQ.Client</AssemblyName>
```
mconnew, line 6: The client project should only declare .net6.0, otherwise it will generate a lib folder for each of the frameworks, include net472, which will result in it claiming to support net472. You will want to make the project references in the test project conditional otherwise when compiling the empty projects for .net framework, it will complain that this project only supports .NET 6.0.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqBinding.cs
```diff
@@ -0,0 +1,100 @@
...
+            set
+            {
+                if (!Enum.IsDefined(typeof(RabbitMqMessageEncoding), value))
+                {
+                    throw new ArgumentOutOfRangeException(SR.Format(SR.InvalidEnumValue, value));
+                }
+                _messageEncoding = value;
+            }
+
+        }
```
mconnew, line 76: Nit: Extra blank line

src/CoreWCF.RabbitMQ.Client/src/CoreWCF/ServiceModel/Channels/RabbitMqOutputChannel.cs
```diff
@@ -0,0 +1,286 @@
...
+            if (token != null)
+            {
+                // When UserNameSecurityToken is made public, cast token to UserNameSecurityToken
+                // and read UserName and Password properties directly instead of using reflection.
+                userName = GetPropertyValueFromReflection(token, "UserName");
+                password = GetPropertyValueFromReflection(token, "Password");
+            }
+
+            _connectionSettings = GetConnectionSettings(userName, password);
+            var factory = _connectionSettings.GetConnectionFactory();
```
mconnew, line 176: How is the Via address getting to the connection settings?

src/CoreWCF.RabbitMQ/tests/IntegrationTests.cs
jonlouie: The 2 pull request links are returning 404 errors for me. Is there PR # associated with them that I could try using instead?

mconnew: I don't know what happened but I think it's fixed now

jonlouie: Links work now, thanks!

jonlouie: Is there an alternate method to [get the cert validator](https://github.com/dotnet/wcf/blob/main/src/System.ServiceModel.Http/src/System/ServiceModel/Channels/HttpsChannelFactory.cs#L32)? It appears to be [internal](https://github.com/dotnet/wcf/blob/main/src/System.ServiceModel.Primitives/src/System/ServiceModel/Security/X509ServiceCertificateAuthentication.cs#L151).

src/CoreWCF.RabbitMQ/tests/CoreWCF.RabbitMQ.Tests.csproj
```diff
@@ -22,14 +27,9 @@
     <PackageReference Include="RabbitMQ.Client" Version="6.4.0" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+    <ProjectReference Include="..\..\CoreWCF.RabbitMQ.Client\src\CoreWCF.RabbitMQ.Client.csproj" />
```
mconnew, line 31: Your conditions only allow for net6.0. You should check against net472 otherwise you don't get anything executed on .net7.0. So this condition would be != 'net472', and the one above would be == 'net472'.

src/CoreWCF.RabbitMQ.Client/src/CoreWCF.RabbitMQ.Client.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
g7ed6e, line 4: isn't netstandard2.0 a possible target ?

jonlouie: This project references some apis made available in `System.ServiceModel.Primitives, version 6.0.0-preview1.23060.3`, which only supports net6.0 and above

mconnew: Even if System.ServiceModel.Primitives still supported netstandard2.0, it still couldn't. The problem is when targeting netstandard, you can end up running on .NET Framework. On Framework the Binding and BindingElement classes have references to IChannelListener for the service side of things, but these don't exist in the client. This would cause runtime problems as the implementation classes won't have the correct overrides of these methods. You won't even be able to instantiate an instance.  
The only way to do it would be to not target netstandard but multi-target .NET Framework and .NET 6.0. If there's demand, we can do that and just have throwing stubs for the service side methods and use conditional compilation. I don't think this is a big issue as the migration path would be .NET Framework WCF with MSMQ -> .NET Framework CoreWCF with MSMQ -> .NET CoreWCF with MSMQ -> .NET CoreWCF with RabbitMQ. If you have a substantial service, the migration will want to be done more gradually.

PR #1030, opened 2023-03-16
Use only parameter type to compare user provided impl and operation contract in source gen

### LINKED ISSUE #1029: `[FromServices]` codegen does not work if implementation parameter names do not match interface parameter names
Example with CoreWCF 1.0.0 on net7.0-windows on VS 17.4.4:

    [ServiceContract]
    interface IService 
    { 
        [OperationContract] void DoThings(int aaa);
    }

    // CS0535 ServiceImpl does not implement DoThings from IService
    public partial class ServiceImpl : IService
    {
        void DoThings(int bbb, [FromServices] HttpContext hc) { }
    }

Expected behavior: codegen occurs for `[FromServices]` regardless of implementing method argument names.
Actual behavior: CS0535 error occurs
Workaround: Update service or interface to have matching argument names

g7ed6e:
Fixes #1029


mconnew (approved):
Please squash your commits when you merge.

PR #1033, opened 2023-03-17
Run unit tests concurrently

g7ed6e:


src/Common/tests/Helpers/WebHostExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    public static class WebHostExtensions
+    {
+        public static int GetHttpPort(this IWebHost webHost)
```
mconnew, line 9: Could this be implemented as
```c#
public static int GetHttpPort(this IWebHost webHost) => GetHttpBaseAddressUri(webHost).Port;
```
Similar idea for https

src/CoreWCF.BuildTools/tests/.cs
```diff
@@ -0,0 +1,4 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
mconnew, line 1: Should this file have a filename and not just an extension? It feels wrong to have a file without a name.

src/CoreWCF.ConfigurationManager/tests/WSHttpBindingTests.cs
```diff
@@ -63,8 +63,10 @@ public void WSHttpBinding_WithCustomSetting()
             }
         }
 
+#if !NETFRAMEWORK
```
mconnew, line 66: Why did you add this condition? Was it to stop the IDE from trying to run the test? The WCF client repo has a custom XUnit test runner with a lot of nice capabilities like forcing a crash dump with the test freezes. But I think that's overkill here. I found a nice approach in the [blog post here](https://josephwoodward.co.uk/2019/01/skipping-xunit-tests-based-on-runtime-conditions) where a custom attribute that derives from FactAttribute can be used. We could replace the Trait (and the command line arguments in the pipeline yml files) with a custom attribute which sets the Skip value.

src/CoreWCF.BuildTools/tests/.cs
mconnew: I've had an idea, could we have a single source file with this in and have a project Property which triggers adding the common test file to the Compile item group? The Directory.Build.Targets file could add the source file to the Compile item group based on the property value.

src/Common/tests/UseDefaultConnectionLimitsAttribute.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Net;
+using System.Reflection;
+using Xunit.Sdk;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+public class UseEndpointConnectionLimitAttribute : BeforeAfterTestAttribute
```
mconnew, line 7: This attribute can't be used on more than one test in the project. Think of the scenario where TestA uses this attribute and is running. Just before TestA ends, TestB also using this attribute starts. Then TestA ends and resets the connection limit just before any client connections are created in TestB, causing TestB to fail.  
This can be solved by adding a static synchronization primitive, eg a SemaphoreSlim with a capacity of 1. In the Before method, call Wait() with a high enough timeout to allow for another test (or even multiple) to complete, and in the After method, call Release().

src/CoreWCF.Http/tests/HttpSysTests.cs
```diff
@@ -13,6 +14,7 @@
 
 namespace CoreWCF.Http.Tests
 {
+    [Collection(HttpSysTestCollectionDefinition.HttpSysTestCollection)]
```
mconnew, line 17: What does the Collection here do to enable concurrent execution?

src/CoreWCF.Http/tests/ServiceContractNameTests.cs
```diff
@@ -16,6 +16,8 @@ namespace CoreWCF.Http.Tests
     public class ServiceContractNameTests
     {
         private readonly ITestOutputHelper _output;
+        private IWebHost _host;
```
mconnew, line 19: Why is this now a field and not a local variable?

src/CoreWCF.Metadata/tests/.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Xunit;
+
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly, DisableTestParallelization = true)]
```
mconnew, line 6: Okay, I see why now. You could also use this mechanism on the test which required modifying the connection limit for .NET Framework.

mconnew:
This is fantastic. We should look to resolve any pending issues as quickly as possible as this PR will quickly become very painful to keep in sync with main if more tests are added.

src/CoreWCF.Http/tests/HttpSysTests.cs
g7ed6e: Default Xunit behavior is to run tests from a single TestCollection sequentially. I'm using this behavior to prevent HttpSys tests to run concurrently because when passing port 0 it asks for a port in an admin range. As only a few UT use HttpSys I thought it's acceptable to have this collection run sequentially but the other TestCollection run concurrently.

src/CoreWCF.Http/tests/ServiceContractNameTests.cs
g7ed6e: Having the host being a field has no impact on the tests of this file but does allow not having to update all the private methods in the file to pass the host as argument. I propose to open a follow up PR to refactor this test pattern globally as many "old" UT are written this way.

src/CoreWCF.ConfigurationManager/tests/WSHttpBindingTests.cs
g7ed6e: I've created an issue to open a follow up PR to implement the custom Fact attribute + pipeline modifications

src/Common/tests/UseDefaultConnectionLimitsAttribute.cs
g7ed6e: done

src/Common/tests/Helpers/WebHostExtensions.cs
g7ed6e: done

src/CoreWCF.BuildTools/tests/.cs
g7ed6e: renamed to XunitTestAssemblyAttribute.xs and included using conditional constructs from msbuild

src/CoreWCF.Metadata/tests/.cs
g7ed6e: I introduced XunitTestAssemblyAttribute.NoConcurrency.cs to disable concurrency. Inclusion in this project is controlled by Directory.Build.* files

g7ed6e:
> This is fantastic. We should look to resolve any pending issues as quickly as possible as this PR will quickly become very painful to keep in sync with main if more tests are added.

@mconnew i think pending issues are resolved and created issues to open follow up PR

mconnew (approved):

PR #1035, opened 2023-03-17
Fixes handling of IsResponseFormatSetExplicitly and IsRequestFormatSetExplicitly

arontsang:
fixes https://github.com/CoreWCF/CoreWCF/issues/1034

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using CoreWCF.Collections.Generic;
 using CoreWCF.Web;
 
@@ -19,7 +20,7 @@ namespace CoreWCF.Description;
 /// This allows endpoints to be defined using WCF classic attributes as well as CoreWCF
 /// attributes, for a simpler migration path.
 /// </summary>
-internal static class WebHttpServiceModelCompat
+public static class WebHttpServiceModelCompat
 {
     public static void ServiceModelAttributeFixup(ServiceEndpoint endpoint)
     {
@@ -81,7 +82,7 @@ internal static class WebHttpServiceModelCompat
         }
     }
 
-
+    private static readonly Regex IsXXXSetExplicitly = new Regex("Is.*SetExplicitly", RegexOptions.Compiled);
     private static class AttributeConverter<TOut> where TOut : class, IOperationBehavior
     {
         private static readonly ConcurrentDictionary<Type, Func<Attribute, TOut?>> s_converterCache = new();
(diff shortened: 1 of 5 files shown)


src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -19,7 +20,7 @@ namespace CoreWCF.Description;
 /// This allows endpoints to be defined using WCF classic attributes as well as CoreWCF
 /// attributes, for a simpler migration path.
 /// </summary>
-internal static class WebHttpServiceModelCompat
+public static class WebHttpServiceModelCompat
 {
```
mconnew, line 24: This class can't be made public

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -133,6 +136,9 @@ where outputProp.CanWrite
                 // strings, numbers, and enums.
                 // Therefore a simple cast expression should convert the enums simply.
                 let value = Expression.Convert(Expression.Property(inputExpression, inputProp), outputProp.PropertyType)
+                // We need to copy the IsXXXSetExplicitly properties last
+                // as they can be overwritten by the other setters.
+                orderby IsXXXSetExplicitly.IsMatch(inputProp.Name)
                 select (MemberBinding)Expression.Bind(outputProp, value);
```
mconnew, line 142: I think a RegEx is overkill. There's a fixed set of properties and I think you can just use `inputProp.Name.Contains("SetExplicitly")` as there are no other uses of "SetExplicitly" than the properties causing problems.

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
```diff
@@ -118,6 +119,8 @@ private static class AttributeConverter<TOut> where TOut : class, IOperationBeha
                 .Invoke(null, Array.Empty<object>());
         }
 
+
+
```
mconnew, line 123: Nit, you accidentally added extra blank lines.

src/CoreWCF.WebHttp/src/CoreWCF/Description/WebHttpServiceModelCompat.cs
mconnew: I tried making this change and it works. The problem I saw was that without making WebHttpServiceModelCompat public, the tests as written won't compile. There are two solutions.
####Test indirectly
If you set OpenApiContractInfo.ResponseFormat one way, and set the attribute a different way, then get the swagger file, you can check what content type it advertises.

####Include WebHttpServiceModelCompat in test project
I believe you should be able to explicitly add the WebHttpServiceModelCompat source file to the test project and then call it. I don't think it has any dependencies on other internal types.

arontsang: @mconnew I've switch to using an indirect test, by checking the response header instead.

PR #1040, opened 2023-03-26
Add net8.0 UT and keyed by service type ServiceBehavior injection support

g7ed6e:
- Add .net8.0 tfm to unit tests
- Move building the unit tests projects in the build stage
- ~~Introduce a flag in test matrix to control whether build fail or not (only used for net8.0, we should drop it soon)~~
- update S.SM 4.x references to 4.10.3
- update S.SM 6.x references to ~~6.1.0~~ 6.2.0
- Add keyed by service ServiceBehavior support starting from net8

### DIFF
--- a/src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
+++ b/src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
@@ -1,8 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>$(TestTargetFrameworks)</TargetFrameworks>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
@@ -23,20 +22,15 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'netcoreapp2.1' ">
+  <ItemGroup Condition="'$(TargetFramework)'=='net472'">
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.1.3" />
     <PackageReference Include="Microsoft.AspNetCore.Server.HttpSys" Version="2.1.12" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+  <ItemGroup Condition="'$(TargetFramework)'!='net472'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.24" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.13" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(DotNetVersion).*-*" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
+++ b/src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
@@ -30,7 +30,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)'!='net472'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(DotNetVersion).*-*" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(DotNetVersion).*" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
+++ b/src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
@@ -1,12 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>$(TestTargetFrameworks)</TargetFrameworks>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
-
     <!-- Required for multi-framework support for in-memory integration testing-->
     <OutputType>Exe</OutputType>
     <GenerateProgramFile>false</GenerateProgramFile>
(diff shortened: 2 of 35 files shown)


mconnew:
This is a good idea, I'd like to get ahead of any potential breaks coming in .NET 8

g7ed6e:
> e to get ahead of any potential breaks coming in .NET

it looks like we are good with preview5

mconnew:
@g7ed6e, what do you think about introducing a property in the top level Directory.Build.Props which specifies the test tfm's and then just reference that property in each of the test projects? That way in the future we can update all tests with a smaller change.

src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
```diff
@@ -39,6 +39,11 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.5" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
mconnew, line 42: Can we do something smarter using wildcards and string slicing so we don't need to add a new conditional ItemGroup each time? Could we use string substrings on the target framework to extract out the version number of the tfm and then specify the version as `$(DotNetVersion).*` to pull in the latest version?

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
```diff
@@ -97,6 +97,10 @@
       "description": "The target framework for the project.",
       "datatype": "choice",
       "choices": [
+        {
+          "choice": "net8.0",
+          "description": "Target net7.0"
```
mconnew, line 102: Need to update the description to net8.0

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
```diff
@@ -118,8 +122,8 @@
           "description": "Target net462"
         }
       ],
-      "replaces": "net6.0",
-      "defaultValue": "net6.0"
+      "replaces": "net8.0",
```
mconnew, line 125: I don't think the default should be net8.0 until after net8.0 is released GA

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
g7ed6e: done

src/CoreWCF.Templates/src/templates/CoreWCFService/.template.config/template.json
g7ed6e: @mconnew i restored default to be net6.0

src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
g7ed6e: @mconnew I did not succeeded  yet to "template" this or implement a kind of for loop in ItemGroup based of a PropertyGroup value.

g7ed6e:
> @g7ed6e, what do you think about introducing a property in the top level Directory.Build.Props which specifies the test tfm's and then just reference that property in each of the test projects? That way in the future we can update all tests with a smaller change.

done

src/CoreWCF.ConfigurationManager/tests/CoreWCF.ConfigurationManager.Tests.csproj
g7ed6e: done!

g7ed6e:
> @g7ed6e, what do you think about introducing a property in the top level Directory.Build.Props which specifies the test tfm's and then just reference that property in each of the test projects? That way in the future we can update all tests with a smaller change.

done!

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
```diff
@@ -333,5 +351,20 @@ public ServiceDescription(IEnumerable<IServiceBehavior> injectedBehaviors, IServ
             AddBehaviors<TService>(this, behaviors);
             SetupSingleton<TService>(this, services);
         }
+
+        private static Lazy<Func<IServiceProvider, IEnumerable<IServiceBehavior>>> s_getKeyedServiceBehaviorsDelegate = new(() =>
+        {
+            Assembly assembly = Assembly.Load("Microsoft.Extensions.DependencyInjection.Abstractions");
```
mconnew, line 357: Is this assembly guaranteed to be available at runtime? Or do we depend on the developer having added a package?

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
```diff
@@ -322,6 +333,13 @@ public ServiceDescription(IEnumerable<IServiceBehavior> injectedBehaviors, IServ
             ServiceProvider = services;
             // Clone IServiceBehavior DI services implementing ICloneable to allow per service override.
             var behaviors = injectedBehaviors.ToList();
+
+            // TODO: Remove this check and the dynamic assembly load once we target net8
+            if (Environment.Version.Major >= 8 && s_isDynamicCodeSupported.Value)
```
mconnew, line 338: I don't think we need to worry about if dynamic code is supported. The reason we had that switch in the operation invoker is because the Expression implementation will be slower with AOT and that's important on a hot code path. This is at service startup so the performance isn't critical. We can accept the slower interpreted version of Expressions in that scenario. My concern with disabling this feature with aot is some behaviors would be used with the regular coreclr, but when using aot, they aren't applied so the service will behave differently.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
```diff
@@ -333,5 +351,20 @@ public ServiceDescription(IEnumerable<IServiceBehavior> injectedBehaviors, IServ
             AddBehaviors<TService>(this, behaviors);
             SetupSingleton<TService>(this, services);
         }
+
+        private static Lazy<Func<IServiceProvider, IEnumerable<IServiceBehavior>>> s_getKeyedServiceBehaviorsDelegate = new(() =>
+        {
+            Assembly assembly = Assembly.Load("Microsoft.Extensions.DependencyInjection.Abstractions");
+            Type type = assembly.GetType("Microsoft.Extensions.DependencyInjection.IKeyedServiceProvider");
+            ParameterExpression parameter = Expression.Parameter(typeof(IServiceProvider));
+            UnaryExpression cast = Expression.TypeAs(parameter, type);
```
mconnew, line 360: While the asp.net core built in service provider implements IKeyedServiceProvider, a 3rd party implementation might not. Need to have a check if the service provider implements it to avoid either a case exception or a null ref exception.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
g7ed6e: > I don't think we need to worry about if dynamic code is supported. The reason we had that switch in the operation invoker is because the Expression implementation will be slower with AOT and that's important on a hot code path. This is at service startup so the performance isn't critical. We can accept the slower interpreted version of Expressions in that scenario. 

Here i disabled the feature with AOT because my understanding is that we cannot dynamically load an assembly in AOT context but i may be wrong.

> My concern with disabling this feature with aot is some behaviors would be used with the regular coreclr, but when using aot, they aren't applied so the service will behave differently.

My idea is to bring the feature in CoreWCF 1.x targeting netstandard2.0 without AOT support so we can start to get usage. Then once we target net8.0, replace that code with AOT compatible code as we get not reflected access to the `IKeyedServiceProvider` type.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
g7ed6e: good catch thanks!

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
g7ed6e: `CoreWCF.Primitives` pulls it and it also comes with `Microsoft.NET.Sdk.Web`.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceDescription.cs
g7ed6e: @mconnew as discussed I dropped the Assembly.Load and the AOT flag.

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -34,22 +32,16 @@
...
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.Negotiate" Version="6.0.24" />
-  </ItemGroup>
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net7.0'">
+  <ItemGroup Condition="'$(TargetFramework)'!='net472'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.13" />
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.Negotiate" Version="7.0.13" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(DotNetVersion).*-*" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Negotiate" Version="$(DotNetVersion).*-*" />
```
mconnew, line 44: Is this going to cause only the preview/rc builds to be used as it requires a hyphen in the version string?

src/CoreWCF.NetNamedPipe/tests/CoreWCF.NetNamedPipe.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net6.0-windows;net7.0;net7.0-windows</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>$(TestTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(TestTargetFrameworks);net6.0-windows;net7.0-windows;net8.0-windows</TargetFrameworks>
```
mconnew, line 4: You don't need this first TargetFrameworks property as it always gets overwritten by the second one now you removed the condition.

templates/CodeAnalysis.yml
```diff
@@ -55,16 +63,16 @@ stages:
...
 
     - task: DotNetCoreCLI@2
       displayName: Run tests with Coverage
       timeoutInMinutes: 20
       inputs:
         command: test
-        projects: ${{ parameters.testProjects}}
-        arguments: '--no-restore --no-build -c Release -f net7.0 --filter Category!=WindowsOnly --collect:"XPlat Code Coverage" --settings coverlet.runsettings'
+        projects: ${{ parameters.testProjects }}
+        arguments: '--no-restore --no-build -c Release -f net6.0 --collect:"XPlat Code Coverage" --settings coverlet.runsettings'
```
mconnew, line 76: With some of the net8.0 features like keyed dependency injection, should we run the test coverage on net8.0?

mconnew:
Just a couple of minor things, but other than those, looks good.

mconnew (approved):

templates/CodeAnalysis.yml
g7ed6e: done

src/CoreWCF.NetNamedPipe/tests/CoreWCF.NetNamedPipe.Tests.csproj
g7ed6e: good catch Thanks

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
g7ed6e: done

PR #1049, opened 2023-04-03
Porting Mtom fixes from WCF Client

mconnew:
The original code incorrectly added the Mime-Version header and some other headers to the response body (the MTOM XmlWriter does this for you based on writeMessageHeaders) when using HTTP. This change added the Mime-Version header to the HTTP headers and tells the MTOM XmlWriter to not emit them, but only when using HTTP.

See dotnet/wcf#4831 for details

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/HttpChannelHelpers.cs
@@ -898,6 +898,13 @@ namespace CoreWCF.Channels
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


birojnayak:
are there any test to be updated ? Or we don't have any Mtom tests ?

mconnew:
We have tests. I think CoreWCF/WCF is tolerant to this problem as this bug existed in the WCF client project and was passing against .NET Framework WCF service which didn't complain about this. We also have the .NET Framework WCF client in our net472 tests running against the CoreWCF service and not complaining. This does make some sense to me as WCF doesn't care about the Mime-Version header existing and the other extra info (with the headers) is redundant information, so WCF doesn't care if the info also exists somewhere where it isn't checking.  
I believe this was reported because other SOAP stacks did need this to be correct.

birojnayak (approved):

PR #1055, opened 2023-04-07
Implementation of NetNamedPipe

mconnew:


MeikTranel (commented):
I was under the impression that `dotnet test Solution.sln` was already selecting the correct TFM - this seems an awfully complicated solution for something that looks to me like a solved problem?

Out of curiosity - is there a special problem aside from run projects referencing the test.sdk and therefore "IsTestProject=True" and call their internal VSTest Implementation from the vstest host with all TFMs listed in the project?

mconnew:
@MeikTranel, I've been trying to resolve an issue of needing to run the NetNamedPipe tests using the TFM of net6.0-windows. This is because the current WCF client package for that targets net6.0-windows. When the test project targets net6.0-windows, running the tests specifying net6.0 doesn't work. I've added the net6.0 and net6.0-windows tfm's to the test project and conditionally excluded the package reference and the test source code when building for net6.0. This means the net6.0 tests contains zero tests. I've given up trying to get the pipeline to build for net6.0-windows. This whole many days long struggle to solve this has made me realize that targeting net6.0-windows is a bad idea so I'm going to modify the WCF Client package for NetNamedPipe to target net6.0 and use the SupportedOSPlatform attribute instead to prevent use on non-windows, like I've now done here. Then we can get our tests running on net6.0 and net7.0 in this repo. This will be released as Preview to begin with and the tests run in VS locally as well as they should run in NetFx too in the pipeline so it's a problem which will be resolved before going GA.

MeikTranel:
I see - there is some ways to make it work i think but i guess you've got it going for now. If you want some help getting net6.0-windows tests to work i have some tricks up my sleeve - just shout and i can help (assuming me tinkering on azure-pipelines config even works)

mconnew:
I appreciate the offer, but I have a strategy that will work now once the client packages are on net6.0. The -windows tfm's are a bit odd though as there isn't a Linux equivalent so you can't make a package Linux only using the tfm. That was pretty much the deciding factor for me to go plain net6.0 as I would need to take the SupportedOSPlatform attribute approach for anything that might be Linux only in the future and I'm not a big fan of special casing how to do things for a particular platform.

mconnew:
Oh, the IsTestProject property has functionality beyond what we're using it for? I just made up that property to modify including common code or not. We have things like template test projects which are in our .sln file but are excluded from the file glob for the set of projects. Could we disable them by doing things like passing a parameter to the build like `/p:SkipTemplates=true` and then have `<IsTestProject Condition=" '$(SkipTemplates)' == 'true' ">false</IsTestProject>` to stop those tests from being ran when passing the .sln file? This sounds like it would be a lot cleaner way of managing our different subset of tests.  
The reason why we include the framework as part of the matrix is for speed of running the tests. It allows netfx, net7.0 and net6.0 tests to be ran concurrently. I believe not specifying it would cause them to be run sequentially, tripling the time it takes to run our tests.

PR #1060, opened 2023-04-15
Run Templates tests concurrently

g7ed6e:


src/CoreWCF.Templates/tests/BasicTests.cs
```diff
@@ -66,10 +79,20 @@ public TestVariation Framework(string framework)
...
 
-        public static implicit operator object[](TestVariation testVariation) => new object[] { testVariation };
+        public void Deserialize(IXunitSerializationInfo info)
+        {
+
+        }
+
+        public void Serialize(IXunitSerializationInfo info)
+        {
+            info.AddValue(nameof(Arguments), JsonConvert.SerializeObject(Arguments));
```
mconnew, line 89: JSON.NET is being semi-deprecated with the introduction of [System.Text.Json](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/overview). If it doesn't introduce too much extra code, I would prefer to use that. It has first class support for AOT and in the far future we probably want to support AOT scenarios and this would be one more thing needing to be fixed if we stick with JSON.NET.

src/CoreWCF.Templates/tests/BasicTests.cs
```diff
@@ -66,10 +79,20 @@ public TestVariation Framework(string framework)
             return this;
         }
 
-        public static implicit operator object[](TestVariation testVariation) => new object[] { testVariation };
+        public void Deserialize(IXunitSerializationInfo info)
+        {
```
mconnew, line 83: Do you need to implement the deserialization code too? Based on my experience with some custom test runners in the WCF client codebase, this is needed when running the tests in VS as it does the test discovery (including enumerating the variations) in one process, serializes the test variations and passes it to another process to run them. Without the deserialization, the tests didn't run correctly or sometimes at all in the VS test runner. At least one issue I had to solve was isolated to debugging the tests.

src/CoreWCF.Templates/tests/CoreWCF.Templates.Tests.csproj
```diff
@@ -11,8 +11,9 @@
   </ItemGroup>
    <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+    <PackageReference Include="Meziantou.Xunit.ParallelTestFramework" Version="2.1.0" />
```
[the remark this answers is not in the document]
g7ed6e: Yes under the hood it creates a TestCollection per Fact and per Theory (if theory data is serializable). Please see https://www.meziantou.net/parallelize-test-cases-execution-in-xunit.htm

src/CoreWCF.Templates/tests/BasicTests.cs
g7ed6e: It was needed indeed for VS. This is fixed now. Thanks!

src/CoreWCF.Templates/tests/BasicTests.cs
g7ed6e: Done, I drop Json.NET usage in favor of STJ.

mconnew (approved):

PR #1062, opened 2023-04-17
Add custom xunit attributes

### LINKED ISSUE #1052: Add a custom Fact attribute that set the Skip value when needed
> Why did you add this condition? Was it to stop the IDE from trying to run the test? The WCF client repo has a custom XUnit test runner with a lot of nice capabilities like forcing a crash dump with the test freezes. But I think that's overkill here. I found a nice approach in the [blog post here](https://josephwoodward.co.uk/2019/01/skipping-xunit-tests-based-on-runtime-conditions) where a custom attribute that derives from FactAttribute can be used. We could replace the Trait (and the command line arguments in the pipeline yml files) with a custom attribute which sets the Skip value.

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/1033#discussion_r1159158385_

g7ed6e:
Resolves #1052

### DIFF
--- a/src/CoreWCF.Http/tests/HttpSysTests.cs
+++ b/src/CoreWCF.Http/tests/HttpSysTests.cs
@@ -24,7 +24,7 @@ namespace CoreWCF.Http.Tests
             _output = output;
         }
 
-        [WindowsNetCoreOnlyFact]
+        [WindowsOnlyFact]
 #if NET5_0_OR_GREATER
         [System.Runtime.Versioning.SupportedOSPlatform("windows")]
 #endif
--- a/src/CoreWCF.Http/tests/HttpSysTests.cs
+++ b/src/CoreWCF.Http/tests/HttpSysTests.cs
@@ -24,8 +24,7 @@ namespace CoreWCF.Http.Tests
             _output = output;
         }
 
-        [Fact]
-        [Trait("Category", "WindowsOnly")]
+        [WindowsNetCoreOnlyFact]
 #if NET5_0_OR_GREATER
         [System.Runtime.Versioning.SupportedOSPlatform("windows")]
 #endif
--- a/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
+++ b/src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
@@ -228,8 +228,11 @@ namespace WSHttp
             }
         }
 
-        [WindowsNetCoreOnlyFact]
+        [WindowsOnlyFact]
         [Description("transport-security-with-windows-authentication-httpsys")]
+#if NET5_0_OR_GREATER
+        [System.Runtime.Versioning.SupportedOSPlatform("windows")]
+#endif
         internal void WSHttpRequestImpersonateWithHttpSys()
         {
             string testString = new string('a', 3000);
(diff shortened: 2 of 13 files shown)


src/CoreWCF.Http/tests/HttpSysTests.cs
```diff
@@ -24,8 +24,7 @@ public HttpSysTests(ITestOutputHelper output)
             _output = output;
         }
 
-        [Fact]
-        [Trait("Category", "WindowsOnly")]
+        [WindowsNetCoreOnlyFact]
 #if NET5_0_OR_GREATER
```
mconnew, line 28: I believe this one should be WindowsOnly. The #if around NET5_0 was for the benefit of the SupportedOSPlatform attribute.. The UseHttpSys api has this attribute, which then means it's needed to be added to CreateHttpSysBuilder. Without these attributes, you get compiler warnings as the test builds target net6.0 and not net6.0-windows. These attributes aren't available for .NET Framework which is why there's the conditional compilation around the attribute. The test runs just fine on .NET Framework though.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
```diff
@@ -230,14 +227,9 @@ internal void WSHttpRequestImpersonateWithKestrel()
                 ((IChannel)channel).Close();
             }
         }
-#endif
 
-        [Fact]
+        [WindowsNetCoreOnlyFact]
         [Description("transport-security-with-windows-authentication-httpsys")]
```
mconnew, line 232: I believe this should be WindowsOnly for the same reasons as previously commented.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
```diff
@@ -286,12 +277,8 @@ public void WSHttpRequestImpersonateFailsWithoutAuthentication()
             }
         }
 
-        [Fact]
+        [WindowsNetCoreOnlyFact]
         [Description("no-security-with-an-anonymous-client-using-impersonation-httpsys")]
```
mconnew, line 281: WindowsOnly, should work on .NET Framework

src/CoreWCF.NetTcp/tests/NetTcpHostedServiceTests.cs
```diff
@@ -21,8 +21,7 @@ public NetTcpHostedServiceTests(ITestOutputHelper output)
             _output = output;
         }
 
-        [Fact]
-        [Trait("Category", "WindowsOnly")]  // HttpSys not supported on Linux
+        [WindowsNetCoreOnlyFact]  // HttpSys not supported on Linux
 #if NET5_0_OR_GREATER
```
mconnew, line 25: WindowsOnly

mconnew:
Once you've fixed the misattributed tests, everything else looks good so feel free to merge.

src/CoreWCF.Http/tests/HttpSysTests.cs
g7ed6e: done, thank you for the explanation.

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
g7ed6e: done

src/CoreWCF.Http/tests/SimpleWSHTTPTest.cs
g7ed6e: done

src/CoreWCF.NetTcp/tests/NetTcpHostedServiceTests.cs
g7ed6e: done

PR #1074, opened 2023-04-21
making the API public for SamlSecurityToken 

birojnayak:
Fixing #1073 

The Saml2SecurityToken is already public and checked also [here](https://referencesource.microsoft.com/#System.IdentityModel/System/IdentityModel/Tokens/SamlSecurityToken.cs,60)

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/SamlSecurityToken.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/SamlSecurityToken.cs
@@ -46,7 +46,7 @@ namespace CoreWCF.IdentityModel.Tokens
 
         internal MSIdentitySAML.SamlSecurityToken WrappedSamlSecurityToken { get; }
 
-        internal MSIdentitySAML.SamlAssertion Assertion { get; private set; }
+        public MSIdentitySAML.SamlAssertion Assertion { get; private set; }
 
         internal SecurityToken SigningToken { get; set; }
 


src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/SamlNameIdentifierClaimResource.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace CoreWCF.IdentityModel.Tokens
 {
-    internal class SamlNameIdentifierClaimResource
+    public class SamlNameIdentifierClaimResource
```
mconnew, line 8: If we're making this class public, it needs to match .NET Framework for serialization. There's a missing `[DataContract]` attribute, and the fields marked with `[DataMember]` need to either revert to having the same names as .NET Framework, or set the `Name` property on the attribute. Otherwise it will serialize differently and won't be compatible if someone is serializing it with DataContractSerializer.

src/CoreWCF.Primitives/src/CoreWCF/IdentityModel/Tokens/SamlSecurityToken.cs
```diff
@@ -46,7 +46,7 @@ protected void Initialize(MSIdentitySAML.SamlAssertion assertion)
 
         internal MSIdentitySAML.SamlSecurityToken WrappedSamlSecurityToken { get; }
 
-        internal MSIdentitySAML.SamlAssertion Assertion { get; private set; }
+        public MSIdentitySAML.SamlAssertion Assertion { get; private set; }
```
mconnew, line 49: I don't think this is the correct approach. We shouldn't expose a type from the Microsoft.IdentityModel namespace on a CoreWCF.IdentityModel class where it would have normally been a CoreWCF.IdentityModel type if we weren't wrapping it. My preferred approach would be to add an explicit cast operator enable casting CoreWCF.IdentityModel.Tokens.SamlSecurityToken to Microsoft.IdentityModel.Tokens.Saml.SamlSecurityToken. In the explicit cast operator, we just return the instance stored in WrappedSamlSecurityToken. Looking at the issue where this was asked for, they can then cast the CoreWCF instance to a MS.IM instance and work with those api's directly. Otherwise we're going to end up having to expose everything in some way on SamlSecurityToken, and some of the properties we would end up needing to write a whole load of new code for to support it.

mconnew:
LGTM

mconnew (approved):

PR #1082, opened 2023-04-27
Create stale.yml

g7ed6e:


mconnew:
Will this mark an issue as stale if the original author has posted a reply and I don't get to quickly?

mconnew:
Btw, as long as debug-only is set to true, don't wait for me to approve any changes to stale.yml to merge it. There will be some experimenting and you shouldn't be blocked on me to try things.

mconnew (approved):
Approving as this is debug-only so won't break anything.

g7ed6e:
> Will this mark an issue as stale if the original author has posted a reply and I don't get to quickly?

@mconnew The counter starts (or is reset) according to the updated_at value of an issue. I'm understanding that any activity on the issue will reset the counter.

PR #1100, opened 2023-05-31
NetNamedPipe System.ArgumentOutOfRangeException fix

chrjstophoros:
`Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)` is now advancing the buffer at the end of the method, just like `int Read(byte[] buffer, int offset, int count)` does it.

This helped fix a ArgumentOutOfRangeException. I don't know this project good enough however, so I can't really tell if this change makes sense or not.

### DIFF
--- a/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/Framing/ServerSingletonConnectionReaderMiddleware.cs
+++ b/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/Framing/ServerSingletonConnectionReaderMiddleware.cs
@@ -429,13 +429,7 @@ namespace CoreWCF.Channels.Framing
 
                         // Consume those bytes from our buffer
                         _buffer = _buffer.Slice(bytesToCopy);
-
-                        // If the buffer has been exhausted, advance the input pipe to consume them and release the buffer
-                        if (_buffer.Length == 0)
-                        {
-                            _connection.Input.AdvanceTo(_buffer.End);
-                        }
-
+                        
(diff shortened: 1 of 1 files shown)


chrjstophoros:
@dotnet-policy-service agree

mconnew:
@chrjstophoros, how were you triggering this bug? Having a regression test which fails without this fix would be really useful so understanding the failure scenario would really help.

mconnew:
This one's interesting (if you find these sorts of things interesting 😃). I understand why this tripped up on NetNamedPipe and not NetTcp. This only affects streamed mode transfer. When using streamed mode, data is sent in chunks up to (I think?) 16KB. There's a variable length encoded integer sent stating the size of the chunk, then the chunk is actually sent. With NetTcp, we send the size and specify not to send immediately. We basically turn off TCP_NODELAY. This means the OS will buffer the bytes representing the length because it's not efficient to send a few bytes on its own. Then we turn on TCP_NODELAY and send the chunk data. On the receiving side, the data is received in a single buffer with the length attached to the chunk. This means calling DecodeSize will always have some buffer left as the chunk payload will always be received with the chunk size.  

For NetNamedPipe, there's no concept of turning on/off TCP_NODELAY, so everything just gets sent immediately. This means the receiver might receive the chunk size without the chunk data in a call to Input.ReceiveAsync. This results in `_buffer` being empty after DecodeSize is called and EnsureBufferAsync then calls Input.ReceiveAsync again without AdvanceTo having been called.  

This would also have affected the upcoming UnixDomainSockets support. @birojnayak, make sure you rebase with this change before you test streamed mode transfer.

Issue #659, opened 2022-05-10
Could not find a base address that matches scheme https for the endpoint with binding WebHttpBinding. Registered base address schemes are [http].


alexey-kostylev:
Hi guys,
I'm runningCoreWcf service in azure as a WebApp and getting this exception in azure console:
Unhandled exception. System.InvalidOperationException: **Could not find a base address that matches scheme https for the endpoint with binding WebHttpBinding. Registered base address schemes are [http]**.
   at CoreWCF.ServiceHostBase.MakeAbsoluteUri(Uri relativeOrAbsoluteUri, Binding binding, UriSchemeKeyedCollection baseAddresses)
   at CoreWCF.Description.DispatcherBuilder.BuildDispatcher[TService](ServiceConfiguration`1 serviceConfig, IServiceProvider services)
   at CoreWCF.Configuration.ServiceConfiguration`1.GetDispatchers()
   at CoreWCF.Configuration.DispatcherBuilderImpl.BuildDispatchers(Type serviceType)
   at CoreWCF.Channels.ServiceModelHttpMiddleware.BuildBranch()
   at CoreWCF.Channels.ServiceModelHttpMiddleware.EnsureBranchBuilt()
   at CoreWCF.Channels.ServiceModelHttpMiddleware.ServiceBuilderOpenedCallback(Object sender, EventArgs e)
   at System.EventHandler.Invoke(Object sender, EventArgs e)
   at CoreWCF.Channels.CommunicationObject.OnOpened()
--- End of stack trace from previous location ---
   at CoreWCF.Configuration.WrappingIServer.StartAsync[TContext](IHttpApplication`1 application, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)

I use the following configuration:
`public static void ConfigureCoreWcfApp(this IApplicationBuilder app)
        {
            app.UseServiceModel(builder =>
            {
[...]

PR #1101, opened 2023-05-31
add support on IpV6Any addresses in soap:address location, add suppor…

vait:
Hey! I have the same [problem](https://github.com/CoreWCF/CoreWCF/discussions/853) and fixed it by add using LocalPath from URI which generated in UseRequestHeadersForMetadataAddressBehavior

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
@@ -94,17 +94,20 @@ namespace CoreWCF.Description
 
         public IMetadataEndpointAddressProvider DynamicMetadataEndpointAddressProvider { get; internal set; }
 
-        internal bool TryGetHttpHostAndPort(Uri listenUri, HttpRequest httpRequest, out string host, out int port, out string scheme)
+        internal bool TryGetHttpHostAndPort(Uri listenUri, HttpRequest httpRequest, out string host, out int port, out string scheme, out string path)
         {
             host = null;
             port = 0;
             scheme = null;
+            path = null;
 
             Uri customUri = DynamicMetadataEndpointAddressProvider?.GetEndpointAddress(httpRequest);
             if (customUri != null)
             {
+
                 host = customUri.Host;
                 port = customUri.Port;
+                path = customUri.LocalPath != "/" ? customUri.LocalPath : null;
                 scheme = DynamicMetadataEndpointAddressProvider is UseRequestHeadersForMetadataAddressBehavior.UseHostHeaderMetadataEndpointAddressProvider
                     ? listenUri.Scheme
                     : customUri.Scheme;
@@ -275,7 +278,7 @@ namespace CoreWCF.Description
 
         private DynamicAddressUpdateWriter GetDynamicAddressWriter(HttpRequest httpRequest, Uri listenUri, bool removeBaseAddress)
         {
-            if (!TryGetHttpHostAndPort(listenUri, httpRequest, out string requestHost, out int requestPort, out string requestScheme))
+            if (!TryGetHttpHostAndPort(listenUri, httpRequest, out string requestHost, out int requestPort, out string requestScheme, out string path))
             {
                 return null;
             }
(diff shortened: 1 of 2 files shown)


vait:
@dotnet-policy-service agree

g7ed6e:
Thanks @vait for the PR. Do you think you can add unit test ?

vait:
> Thanks @vait for the PR. Do you think you can add unit test ?

I think Yes, I'll try next week

mconnew:
I can see how the path would need capturing, but my worry is about relative paths in endpoints. E.g. if your service has a base path of "http://localhost/MyService.svc" and you add a service endpoint with `AddServiceEndpoint<MyService, IMyService>(new BasicHttpBinding(), "basicHttp")`, then the endpoint address will be "http://localhost/MyService.svc/basicHttp", but the wsdl might be configured to be at "http://localhost/MyService.svc?wsdl". The wsdl should state the service will be found at "http://localhost/MyService.svc/basicHttp" but I suspect this change will break that and make it claim to be at "http://localhost/MyService.svc". The wsdl generation tests purposefully nuke the urls and set them to something explicit before comparing with the expected wsdl output so they won't fail on any bug introduced in this area.

mconnew:
I don't understand why you need to fix the url's containing the ipv6Any address. We don't fix 127.0.0.1. Can you explain what problem you are trying to fix with that?

vait:
Hi, Guys!


I'm trying to solve two problems.

Our services work after reverse proxy/load balancer.

1st problem:

Outside the service has address with virtual directory and https shceme : https://name.domain.com/aspsoap/CallManager.svc.
And in a response we get wrong shceme (http) and empty virtual directory. But our load balancer pass X-Forwarded-header. But in current solution/sample (UseXForwardedHeadersForMetadataAddressProvider) I can't pass virtual directory "asmsoap".

Some guys have a same problem (https://github.com/CoreWCF/CoreWCF/discussions/853).

Screenshot 2023-06-27 094825

2nd problem:
In our instance Kestrel has a following configuration

Screenshot 2023-06-27 095043

Because it work in environment with dynamic IP, we can't specify IP address, so we use wildchar. And this configuration generates service address like this: http://[::]:port/service.svc

Screenshot 2023-06-27 094846

I think I'ii try to change solution with path (if it will be empty in UseXForwardedHeadersForMetadataAddressProvider implementation). But https://github.com/CoreWCF/CoreWCF/pull/1101/files#diff-b51ef3213db6fc1cf9900503a9ca011317d68a0fa2c1b9c1e3b3945c323486ebR257-R264 is correct.

vait:
I understood what do you mean. So a fix will take more time than  I thought )

vait:
Hello, @mconnew!

>The wsdl generation tests purposefully nuke the urls and set them to something explicit before comparing with the expected wsdl output so they won't fail on any bug introduced in this area.

How do you suggest to using a CoreWCF with load balancer?

In my case, the load balancer passes a X-Forwarded-Path header with a "local path" URI of our service . I need the ability to override URI in WSDL with correct adresses and path in my infrastructure.

So, in this PR I added ability to override URI when it's in full (with localpath part) and it works only with UseXForwardedHeadersForMetadataAddressProvider.

VitahaPo:
Hello @mconnew.
Could you check this PR?
Especially I am interested in fix for second problem when CoreWCF generates SOAP address like this http://[::]:port/service.svc.
I am  experiencing this issue in a containerized environment and I see similar problems already exists https://github.com/CoreWCF/CoreWCF/issues/700
https://github.com/CoreWCF/CoreWCF/issues/659
https://github.com/CoreWCF/CoreWCF/issues/1291

Since this PR contains fixes multiple issues I have applied partially this fix for my fork https://github.com/VitahaPo/CoreWCF/commit/d12cf97f91b9a7c30375ae9998015efa04c3e6a3
and it works well so far

Let me know if I can help somehow to move this forward, I can create separate PR with fix only for second issue if needed

leanalcabasa:
Hi @VitahaPo ,

I am also encountering the same issue. Can I ask if we have the same problem?

After implementing the a pproach of https://github.com/CoreWCF/CoreWCF/discussions/853, it fixed the soap:address location, but now xsd:import schemaLocation="" and wsdl:import namespace="http://tempuri.org/" location="" are empty.

leanalcabasa:
Thanks @VitahaPo 

I saw you commented another fix in your fork : https://github.com/VitahaPo/CoreWCF/commit/d12cf97f91b9a7c30375ae9998015efa04c3e6a3

Is it for a different problem?

Hopefully, this PR gets approved so it is supported.

VitahaPo:
> Thanks @VitahaPo
> 
> I saw you commented another fix in your fork : [VitahaPo@d12cf97](https://github.com/VitahaPo/CoreWCF/commit/d12cf97f91b9a7c30375ae9998015efa04c3e6a3)
> 
> Is it for a different problem?
> 
> Hopefully, this PR gets approved so it is supported.

Yeah that was for another problem when  SOAP address in wsdl looks like this http://[::]:port/service.svc. I see it happens for apps deployed in containers

leanalcabasa:
I see, thanks @VitahaPo 

Hi @mconnew , can we revisit this PR? It would be helpful for us that hosts it within AKS or behind a load balancer. Or maybe there is another approach you can suggest?

Thank you

mconnew:
For the wsdl:Import statements with blank locations, try using ?SingleWsdl instead. That merges all the WSDL documents into one so there's no more imports.

VitahaPo:
> For the wsdl:Import statements with blank locations, try using ?SingleWsdl instead. That merges all the WSDL documents into one so there's no more imports.

Unfortunately, it is hard to ask all API users to switch to singleWsdl

I was able to fix it by doing this https://github.com/CoreWCF/CoreWCF/commit/644408fb4aa5af2612c6faf4626ff96be2d06687
which is also fixed in the current PR. Do you know why originally there were functionalities to RemoveBaseAddress?

PR #1112, opened 2023-06-09
Addressing Version WSAddressingAugust2004 is not supported

ioda:
I did the fix for WSAddressingAugust2004 
But didn't sure about the unit test

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/EndpointAddress.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/EndpointAddress.cs
@@ -1013,6 +1013,10 @@ namespace CoreWCF
             {
                 WriteContentsTo10(writer);
             }
+            else if (addressingVersion == AddressingVersion.WSAddressingAugust2004)
+            {
+                WriteContentsTo200408(writer);
+            }
             else if (addressingVersion == AddressingVersion.None)
             {
                 WriteContentsToNone(writer);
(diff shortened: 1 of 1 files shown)


ioda:
@dotnet-policy-service agree company="Wolters Kluwer"

g7ed6e:
Thanks @ioda for the PR. Do you think you can add unit test ?

mconnew:
@ioda, I found a simpler way to test the output without dealing with Encoding bytes from a MemoryStream to get the string. A unit test would look something like this (not tested code against compiler so might be minor errors, but hopefully you get the idea):
```c#
string expectedEndpointXml = @".... the xml that woud be output on .NET Framework"; // Don't forget to double up the quotes in the xml
var endpointAddress = new EndpointAddress("http://example.org/service.svc");
StringBulder sb = new StringBuilder();
using(var xmlWriter = XmlWriter.Create(sb))
{
    endpointAddress.WriteContentsTo(AddressingVersion.WSAddressingAugust2004, xmlWriter);
}
Assert.Equal(sb.ToString(), expectedEndpointXml;
```
You'll need to run the code on .NET Framework first to get the expected string which you put in on the first line.

src/CoreWCF.Primitives/src/CoreWCF/EndpointAddress.cs
```diff
@@ -1085,6 +1089,77 @@ private void WriteContentsTo10(XmlDictionaryWriter writer)
                 }
             }
         }
+        void WriteContentsTo200408(XmlDictionaryWriter writer)
```
mconnew, line 1092: Needs a blank line between methods. Also missing a private accessibility modifier.

mconnew:
I wrote the unit tests so we can merge this change.

PR #1119, opened 2023-06-19
Fix ServiceAuthorizationBehavior lifetime to enable override per registered Service 

### LINKED ISSUE #1114: ServiceAuthenticationBehavior throws NullReferenceException in GetAuthorizationService if new:ed up to be applied to one specific service only
I'm trying to apply a ServiceAuthenticationBehavior on a specific `ServiceHostBase` only. I do this by:

```c#
serviceBuilder.ConfigureAllServiceHostBase(serviceHost =>
{
    if (serviceHost.Description.ServiceType == typeof(MyCustomService))
    {
        serviceHost.Description.Behaviors.Remove<ServiceAuthorizationBehavior>(); // The shared Singleton instance
        serviceHost.Description.Behaviors.Add(new ServiceAuthorizationBehavior
        {
            ExternalAuthorizationPolicies = new ReadOnlyCollection<IAuthorizationPolicy>(new List<IAuthorizationPolicy> { new MyCustomAuthorizationPolicy() }),
            PrincipalPermissionMode = PrincipalPermissionMode.Custom,
        });
    }
}
```

However, when I do so, a NullReferenceException is thrown:

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=CoreWCF.Primitives
  StackTrace:
   at CoreWCF.Description.ServiceAuthorizationBehavior.CoreWCF.Description.IServiceBehavior.ApplyDispatchBehavior(ServiceDescription description, ServiceHostBase serviceHostBase) in /_/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs:line 216
```

This code is called from `DispatchBuilder.InitializeServiceHost`.

It appears to be because the internal setter property `ServiceScopeFactory` is never assigned a value, so it remains `null`.

Only workaround I've found is to set it through reflection, i.e:
```c#
typeof(ServiceAuthorizationBehavior)
    .GetProperty("ServiceScopeFactory", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance)
    .SetValue(tempServiceAuthorizationBehavior, app.ApplicationServices.GetService<IServiceScopeFactory>());
```

But this can't be right. Is there a better way of doing this or have I stumbled upon a bug or unsupported scenario?

g7ed6e:
fixes #1114


src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
```diff
@@ -38,6 +38,20 @@ public ServiceAuthorizationBehavior()
             _principalPermissionMode = DefaultPrincipalPermissionMode;
         }
 
+        private ServiceAuthorizationBehavior(ServiceAuthorizationBehavior other)
```
mconnew, line 41: I really like this approach better than my original idea. You don't need to do anything different on the DI injection piece with this change, you can leave it as a simple singleton as the service description is cloning it.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
```diff
@@ -224,6 +241,11 @@ internal void MakeReadOnly()
             _isReadOnly = true;
         }
 
+        internal ServiceAuthorizationBehavior Clone()
```
mconnew, line 244: Make this class implement ICloneable and then ServiceDescription can be general cased to if it implements ICloneable, case to that and call Clone.  Only differences is ICloneable returns `object` so you would need to cast, but that's not a big deal.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
g7ed6e: The idea is the configuration one would apply when pulling the service from DI is a kind of template that allow for further override when using `ServiceBuilder.ConfigureServiceHost`.

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
g7ed6e: done. It looks better this way. Thanks!

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceAuthorizationBehavior.cs
mconnew: You would still get that with a straight Singleton. You pull it from DI and customize it, then when populating the description, it would get cloned. Then in ConfigureServiceHostBase, you would pull it from the behaviors and modify it, which would affect other instances as it's cloned.

g7ed6e: @mconnew then maybe that's better to drop the factory introduced by this PR. What do you think?

g7ed6e: @mconnew Done! Thanks

mconnew (approved):
I love it when an elegant simple solution can be found to a potentially difficult problem like this!!

PR #1127, opened 2023-06-28
Drop RabbitMQ ServiceCollectionExtension method

### LINKED ISSUE #1125: AddServiceModelRabbitMqSupport is useless because RabbitMqTransportPump is newed by the RabbitMqTransportBindingElement
Having a look to what is done for RabbitMQ it seems that RabbitMqTransportPump is added to DI through an extension method named AddServiceModelRabbitMqSupport which is never called and RabbitMqTransportBindingElement is manually newing the pump instance which mean that the service registration (and thus the extension method) is useless.
I think we should align both implementations what is your view about this ?

_Originally posted by @g7ed6e in https://github.com/CoreWCF/CoreWCF/pull/943#discussion_r1244251891_

g7ed6e:
Resolves #1125


g7ed6e:
@mconnew / @birojnayak  Also removed the `AddServiceModelMsmqSupport` from `CoreWCF.MSMQ` and dropped some public types that can be kept internal as we are still in preview.
The idea is to align MSMQ, RabbitMQ and Kafka implementation avoiding the unnecessary AddXXX for the time being.
If we decide that on have to call a queue specific AddXXX we should align implementation.

mconnew (approved):

PR #1129, opened 2023-06-29
Cleanup test code

afifi-ins:


g7ed6e:
@afifi-ins we may also add https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.CodeStyle/ to enforce code style at builtime.. but i'm not a big fan of breaking build for code style reason.
Another option would be to decorate PR with sonarcloud.

g7ed6e (approved):

mconnew:
@g7ed6e, I'd like to fail the PR for some code style failures, but I agree failing the actual build probably isn't the best thing as it gets in the way of seeing if the actual tests themselves are passing. It would make draft PR's less useful.

PR #1148, opened 2023-07-14
Expose an new API with HttpTransportBindingElement and HttpTransportSecurity to enable AuthorizationPolicy support

### LINKED ISSUE #1143: Cannot specify AuthenticationScheme when using the new [Authorize] support.
It's great to be able to integrate with ASP.NET Core's auth system, but I need to use HTTP Basic and output `<http:BasicAuthentication/>` in my WSDL. In order to use [Authorize], you currently have to have ClientCredentialType.InheritedFromHost instead of ClientCredentialType.Basic; this doesn't use a policy-provided AuthenticationScheme, only the DefaultScheme, so it doesn't call ChallengeAsync() on my custom basic auth handler. 

It also doesn't provide any authentication scheme in the service metadata. This can be done with a CustomBinding, but then there's no way to trigger the [Authorize] support - it does a type check for BasicHttpBinding, WebHttpBinding and WSHttpBinding specifically.

g7ed6e:
resolves #1143


mconnew:
CustomBinding is the wrong place to do this. We should leave it as purely a modifiable collection of BindingElement instances with zero implementation knowledge. Add a property on HttpTransportBindingElement for this. There are potentially 3 options, use it, don't use it, or use the default. There's a few ways we could do this.
1. Use an enum with 3 options
2. Use a `Nullable<bool>`
3. Keep track of if the property is explicitly set. See OpenApiSchemaBuilder for an example. We would want to keep this extra property internal though.
4. Phrase the property as always using it. E.g. AlwaysUseAuthorizationPolicySupport. A value of true always uses it, a value of false means it might be used if implicitly turned on (InheritFromHost)

My preference is 4, but I'm not an expert on good api's so leaving it open. If we go with option 4, it could also be exposed on some of the Http bindings, e.g. BasicHttpBinding.Security.Transport.AlwaysUseAuthorizationPolicySupport

src/CoreWCF.Http/src/CoreWCF/HttpBindingBase.cs
```diff
@@ -137,25 +137,24 @@ internal TransportBindingElement GetTransport()
...
-            {
-                basicHttpSecurity.EnableTransportAuthentication(_httpTransport);
-                return _httpTransport;
-            }
-            else
-            {
-                // ensure that there is no transport security
-                basicHttpSecurity.DisableTransportAuthentication(_httpTransport);
-                return _httpTransport;
+            switch (basicHttpSecurity.Mode)
```
mconnew, line 140: Nice cleanup

mconnew (approved):

PR #1149, opened 2023-07-14
Enable the stale workflow

g7ed6e:


g7ed6e:
@mconnew i think we can enable this

Issue #1126, opened 2023-06-28
Calling OperationContext.Current before the host started causes later exceptions and bugs


DxCK:
### Background
Each time the application in question write to log, it accesses `OperationContext.Current` to try and gather some details when available.
This fact causes the `OperationContext.Current` to be accessed as early as before the service host even created.
As I'm migrating the said application into CoreWCF, I have encountered very weird bugs and exceptions, which I traced back to the above early `OperationContext.Current` access.

### Outcomes
The most disturbing outcome is that `OperationContext.Current` may be replaced in mid-operation if there is more than 1 client that invokes operations at a given time. Meaning:

```
        // this is an [OperationContract] implementation
        public void DoSomething()
        {
            if (OperationContext.Current != OperationContext.Current)
            {
                // sometimes it gets here...
            }
        }
```

Some other outcomes I have observed are:
- ObjectDisposedException when accessing OperationContext.Current.IncomingMessageHeaders.
- NullReferenceException sent as a fault back to the client (Visible only when `IncludeExceptionDetailInFaults = true`)
- Some other exceptions thrown internally within CoreWCF infra.

### How to Reproduce
To reproduce, just add the following line before the aspnetcore host creation:
`var context = CoreWCF.OperationContext.Current;`
And the executable project must target .NET Framework.

I wrote a simple project to reproduce it:
[CoreWcfOperationContextBugRepro.zip](https://github.com/CoreWCF/CoreWCF/files/11892471/CoreWcfOperationContextBugRepro.zip)

I have verified the bug to exist in versions: 1.3.1, 1.3.2, and 1.4.0-preview1 .
It's possible more versions are affected, but I didn't test others.

### Probable Cause
There is a `class Holder` which actually points to the `OperationContext`, and there is an `AsyncLocal<Holder> s_currentContext` which is carrying a different reference of `Holder` for each asynchronous control flow.
Accessing `OperationContext.Current` for the first time causes the instantiation of `Holder` class for the current asynchronous control flow, and it sets it into `AsyncLocal`. Once a `Holder` reference is set into `AsyncLocal`, it will be reused forever to point to a given `OperationContext` including future ones that relevant to the current asynchronous control flow.
[...]

PR #1157, opened 2023-07-18
Fixed: Calling OperationContext.Current before the host started

DxCK:
- Fixed: Calling OperationContext.Current before the host started causes later exceptions and bugs.
- Implemented a test that validates the above fixed issue.

#1126

### DIFF
--- a/src/CoreWCF.Primitives/tests/OperationContextTests.cs
+++ b/src/CoreWCF.Primitives/tests/OperationContextTests.cs
@@ -58,7 +58,7 @@ namespace CoreWCF.Primitives.Tests
                     try
                     {
                         // Wait until a second request is running.
-                        if (!SecondRequestInMidOperationEvent.Wait(TimeSpan.FromSeconds(10)))
+                        if (!SecondRequestInMidOperationEvent.Wait(TimeSpan.FromSeconds(30)))
                         {
                             throw new FaultException("An expected event from the second request did not set in time.");
                         }
@@ -82,7 +82,7 @@ namespace CoreWCF.Primitives.Tests
                     SecondRequestInMidOperationEvent.Set();
 
                     // Wait until OperationContext assertion happens in the first request
-                    if (!OperationContextAssertedEvent.Wait(TimeSpan.FromSeconds(10)))
+                    if (!OperationContextAssertedEvent.Wait(TimeSpan.FromSeconds(30)))
                     {
                         throw new FaultException("An expected assertion event from the first request did not set in time.");
                     }
@@ -139,7 +139,7 @@ namespace CoreWCF.Primitives.Tests
                     Task firstRequest = RequestAndAssert(cancellationSource);
 
                     // Wait until the first request stored initial OperationContext.Current .
-                    if (!service.FirstRequestStoredOperationContextEvent.Wait(TimeSpan.FromSeconds(10), cancellationToken))
+                    if (!service.FirstRequestStoredOperationContextEvent.Wait(TimeSpan.FromSeconds(30), cancellationToken))
                     {
                         // In case we don't get the event on time, we want first and foremost to see any exception thrown
                         // from the request.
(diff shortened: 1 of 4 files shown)


DxCK:
@dotnet-policy-service agree company="Quali"

src/CoreWCF.Primitives/tests/OperationContextTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        [ServiceBehavior(IncludeExceptionDetailInFaults = true)]
+        private class Service : IContract
+        {
+            public void Operation()
+            {
+                var context = OperationContext.Current;
+
+                // This sleep is not needed to reproduce the original bug in a real environment,
+                // but is essential to reproduce it in a test environment.
+                Thread.Sleep(100);
```
mconnew, line 42: I believe this test is going to be fragile (if the issue wasn't fixed). My understanding of why you sleep is to allow time for another thread to modify the same Holder and the delay is to allow time for that to happen. I believe there's an easier more deterministic way to achieve this and you should be able to do it with a single request. Modify the service to look like this:
```c#
        [ServiceBehavior(IncludeExceptionDetailInFaults = true)]
        public class Service : IContract
        {
            public ManualResetEventSlim RequestStarted { get; } = new ManualResetEventSlim(false);
            public ManualResetEventSlim ContextModified { get; } = new ManualResetEventSlim(false);
            public void Operation()
            {
                var context = OperationContext.Current;
                RequestStarted.Set();
                ContextModified.Wait(TimeSpan.FromSeconds(10));

                if (context != OperationContext.Current)
                {
                    throw new FaultException("OperationContext.Current was replaced in mid-operation.");
                }
            }
        }
```
In the `Startup.ConfigureServicesMethod` add the `Service` to DI as a singleton.
```c#
        private class Startup
        {
            public void ConfigureServices(IServiceCollection services)
            {
                services.AddSingleton<Service>();
            }
            /// Rest of class
         }
```
In the test method, you start the request but don't await it yet. You then wait on the `RequestStarted` event, modify the `OperationContext.Current`, then set the `ContextModified` event, then await the request.
```c#
    var responseTask = client.SendAsync(request);
    var service = host.Services.GetRequiredService<Service>();
    service.RequestStarted.Wait(TimeSpan.FromSeconds(10));
    OperationContext.Current = null;
    service.ContextModified.Set();
    var response = await responseTask;
    var responseBody = await response.Content.ReadAsStringAsync();
```

You only need to make one request so remove the Enumerable.Range which kicks off multiple parallel requests and just make a single request. Basically this works by kicking off the http request but not waiting for the response yet, then waiting on a `ManualResetEventSlim` to know that the request has started and the initial `OperationContext.Current` value has been captured. The service implementation is now waiting to be signaled that the `OperationContext.Current` has been modified. So the test sets it to null, which will be reflected in the service call, then signals the service to continue with the operation where it then picks up the null value.  

As a general rule don't use Sleep calls to try to trigger a bug in a probabilistic way. The CI test hosts tend to only have 2 CPU's configured so tests like yours are likely to fail there as there's a good chance that with 100 runs there will be at least one where not two calls are executing concurrently (which is what's required to hit this bug).

DxCK: I have pushed my implementation a perfectly synchronized approach for the test, based on your comments.
However beware that the timeouts make the test fragile in the CI environment. It was failed with a timeout of 10 seconds, but passed with a timeout of 30 seconds. I don't really know what is the correct number to put there.

mconnew: Being fragile due to taking too long is different than being fragile due to racing code. Put the timeout at 30 seconds and if that turns out to be too low, I can increase it later.

DxCK: The current pushed code already has the timeout at 30 seconds.

mconnew: Great, I'll merge this. Thanks for doing this and contributing a repro test. I plan to do a release including this fix next week unless anything else comes up which derails things.

Issue #1003, opened 2023-02-14
XML signature validation error if signature element has namespace prefix


gekiss:
In case when XML Signature has namespace prefix then throws the `System.ArgumentNullException: Value cannot be null. (Parameter 'value')`

Below a part of XML
```
<s:Header>
<wsse:Security xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd">
...
<ds:Signature Id="SIG-49D16B5DDDE3A09A9E1676366233345155" xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
<ds:SignedInfo>
 ...
</ds:Signature>
</wsse:Security>
...
</s:Header>
```

This happens because signature element retrieved using the method GetElementsByTagName without namespace specified.
https://github.com/CoreWCF/CoreWCF/blob/3503aa81ff3e71d2ce2d67f0562bcca9056f3893/src/CoreWCF.Primitives/src/CoreWCF/Security/WSSecurityOneDotZeroReceiveSecurityHeader.cs#L128

I think it should be:
```
 XmlNodeList nodeList = doc.GetElementsByTagName("Signature", SignedXml.XmlDsigNamespaceUrl); 
```

jacek450:
@gekiss I encountered the same issue and applying your fix worked.

PR #1180, opened 2023-08-11
Fixed: Xml signature validation

gekiss:
- Fixed: XML signature validation error if signature element has namespace prefix.
- Fixed: XML signature validation error if signature contain `InclusiveNamespaces` element.
- Implemented a test that validates the above fixed issue.
#1003


gekiss:
@dotnet-policy-service agree

src/CoreWCF.Primitives/src/CoreWCF/Security/XmlHelper.cs
```diff
@@ -411,5 +411,40 @@ public static UniqueId ReadTextElementAsUniqueId(XmlElement element)
...
             return new UniqueId(ReadTextElementAsTrimmedString(element));
         }
+
+        internal static string[] TokenizeInclusiveNamespacesPrefixList(string inclusiveNamespacesPrefixList)
+        {
+            if (inclusiveNamespacesPrefixList == null)
+                return null;
+
+            string[] prefixes = inclusiveNamespacesPrefixList.Split(null);
+            int count = 0;
```
mconnew, line 421: You can simplify this code by using the Split overload which takes StringSplitOptions and pass RemoveEmptyEntries. This way the returned array will always be the correct length and you only need to replace "#default" with String.Empty. That allows you to drop the prefix.Length check and the resize code at the end.

src/CoreWCF.Primitives/src/CoreWCF/Security/WSSecurityOneDotZeroReceiveSecurityHeader.cs
```diff
@@ -125,8 +125,21 @@ protected override SignedXml ReadSignatureCore(XmlDictionaryReader signatureRead
                 writer.WriteEndDocument();
             }
             SignedXMLInternal signedXml = new SignedXMLInternal(doc);
-            XmlNodeList nodeList = doc.GetElementsByTagName("Signature");
+            XmlNodeList nodeList = doc.GetElementsByTagName("Signature", SignedXml.XmlDsigNamespaceUrl);
             signedXml.LoadXml((XmlElement)nodeList[0]);
+            if (signedXml.SignedInfo.CanonicalizationMethodObject is XmlDsigExcC14NTransform xmlDsigExcC14NTransform)
+            {
```
mconnew, line 131: I tried to do some validation looking at the behavior with our existing tests and it looks like we usually have XmlDsigC14NTransform as our transform so any problems here won't be picked up by our existing tests. The main thing I'm concerned with here is that signatureReader might not have a namespace matching the prefix when you do the lookup. Can you null check the value returned from LookupNamespace and only add if it isn't null.

mconnew:
Just a couple of small changes and we can accept this PR.

src/CoreWCF.Primitives/src/CoreWCF/Security/XmlHelper.cs
```diff
@@ -411,5 +411,27 @@ public static UniqueId ReadTextElementAsUniqueId(XmlElement element)
         {
             return new UniqueId(ReadTextElementAsTrimmedString(element));
         }
+
+        internal static string[] TokenizeInclusiveNamespacesPrefixList(string inclusiveNamespacesPrefixList)
+        {
+            if (inclusiveNamespacesPrefixList == null)
+                return null;
+
+            string[] prefixes = inclusiveNamespacesPrefixList.Split(new char[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
mconnew, line 420: You can pass `null` as the first parameter. The "The separator array" section of the documentation says:

> If the separator parameter is null or contains no characters, white-space characters are assumed to be the delimiters.

You could also pass `Array.Empty<char>` if there's problems with ambiguous overloads.
```suggestion
            string[] prefixes = inclusiveNamespacesPrefixList.Split(null, StringSplitOptions.RemoveEmptyEntries);
```

mconnew:
I should be releasing this next week if no blockers come up.

PR #1190, opened 2023-08-24
Add IISExpress launch profile in templates

g7ed6e:


mconnew:
I don't want to expand the scope of this PR so this comment is about potential follow-up work. Is there a way to have tests for the template? I don't know if there's any existing pattern for testing templates, but in the wcf dotnet-svcutil project (which isn't a template, but is a command line tool), we run the commandline tool to generate the code and have a checked in folder containing the expected output from running it. We have utility methods which mostly do a straight file comparison, but have some smarts which can understand the parts which might vary from run to run and ignores those differences.

mconnew (approved):

g7ed6e:
> I don't want to expand the scope of this PR so this comment is about potential follow-up work. Is there a way to have tests for the template? I don't know if there's any existing pattern for testing templates, but in the wcf dotnet-svcutil project (which isn't a template, but is a command line tool), we run the commandline tool to generate the code and have a checked in folder containing the expected output from running it. We have utility methods which mostly do a straight file comparison, but have some smarts which can understand the parts which might vary from run to run and ignores those differences.

- That's a good idea and it could ease the testing part. The existing UT we have for the templates are using dotnet new to get the templates code and then build, publish and host the app with a dotnet run. This new flag requires to orchestrate IIS express in CI to achieve the same result.

mconnew (approved):

PR #1195, opened 2023-08-28
Cleanup the share memory object for named pipe

### LINKED ISSUE #1117: Stopping a host that's using NetNamedPipe doesn't clean up the shared memory
Reported in https://github.com/CoreWCF/CoreWCF/issues/6#issuecomment-1594193331 by #716 
This blocks starting a new instance within the same process. I believe the following code in NamedPipeListener should fix it, but I need to verify with tests.
```c#
        internal Task StopAsync(CancellationToken cancellationToken)
        {
            // TODO: Signal the main dispatcher we're closing down and give it some time to cleanly
            // close the channels. If it doesn't close them in time, we'll close the connection from
            // underneath it to avoid leaking the connection.
            // For now we just close the connection from underneath it.
            _sharedMemory?.Dispose();
            _sharedMemory = null;
            return _transportConnectionManager.CloseAllConnectionsAsync(cancellationToken);
        }
```

mconnew:
Fixes #1117


mconnew:
Failure was in an unrelated Kafka test. Running the test again.

PR #1218, opened 2023-09-11
Use expressions to build InvokeDelegate

g7ed6e:


mconnew:
I have a query out asking about Expression support in aot. In the old UWP .NET Native days it was implemented with a slow tree walking interpreter. Presuming that's still the case, then you probably want to keep the old implementation switched with an if block around `RuntimeFeature.IsDynamicCodeSupported`. The JIT/aot compiler recognizes that check and will eliminate  the unnecessary implementation on that platform.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -104,54 +106,123 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
-                    }
+                var targetParam = Expression.Parameter(typeof(object), "target");
+                var inputsParam = Expression.Parameter(typeof(object[]), "inputs");
+                var outputsParam = Expression.Parameter(typeof(object[]), "outputs");
+                var paramsLocal = Expression.Variable(typeof(object[]), "paramsLocal");
+                var result = Expression.Variable(typeof(object), "result");
+
+                var methodParam = Expression.Constant(method);
+
+                var assignParamsLocal = Expression.Assign(paramsLocal, Expression.Condition(
```
mconnew, line 170: You don't need to have the condition on the paramCount be part of the expression. You can conditionally emit a different expression based on the paramCount > 0. This is part of the performance improvement moving to Expression's, you can emit less code based on conditions.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +168,74 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                    Expression.GreaterThan(Expression.Constant(paramCount), Expression.Constant(0)),
+                    Expression.NewArrayBounds(typeof(object), Expression.Constant(paramCount)),
+                    Expression.Constant(null, typeof(object[]))));
+
+                var assignParamsLocalValues = inputPos.Length == 0
+                    ? Expression.Block(Expression.Empty())
+                    : Expression.Block(
+                        Enumerable.Range(0, inputPos.Length).Select(i =>
```
mconnew, line 191: Nice loop unrolling!!

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +168,74 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                            Expression.Assign(
+                                Expression.ArrayAccess(paramsLocal, Expression.Constant(inputPos[i])),
+                                Expression.ArrayIndex(inputsParam, Expression.Constant(i)))));
+
+                var invokeMethod = Expression.Block(
+                    Expression.Assign(result, Expression.Condition(
```
mconnew, line 197: You can move the returnsValue check outside of the Expression creation and conditionally generate a different Expression based on it. No need to include in the execution of the Expression itself.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +168,74 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                        Expression.Equal(Expression.Constant(returnsValue), Expression.Constant(true)),
+                        Expression.Call(methodParam, nameof(MethodInfo.Invoke), null, targetParam, paramsLocal),
```
mconnew, line 199: It looks like you are calling "Invoke" on the methodInfo instance. So you are still calling `method.Invoke(target, paramsLocal)`. With Expression.Call, you can instead do `Expression.Call(targetParam, methodInfo, arguments)` where `arguments` is an `Expression[]` containing an array populated with Expression instances populated from inputsParam. Basically the loop you unrolled earlier will end up instantiating an array of Expression instead that you will pass here. There is a possibility you will need to emit an `Expression.Cast`, but I hope that's not the case. If it is, then the type you cast to will come from `methodInfo.GetParameters()`. If you do need to do that, make sure you get the type from the parameter array using the position from `inputPos[i]` and not position `i`.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +168,74 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                        Expression.Block(
+                            Expression.Call(methodParam, nameof(MethodInfo.Invoke), null, targetParam, paramsLocal),
+                            Expression.Constant(null, typeof(object))))),
+                    Expression.Empty());
+
+                var catchParameterExpr = Expression.Parameter(typeof(TargetInvocationException), "tie");
+                var throwCapturedExceptionDispatchInfoExpr = Expression.Call(Expression.Call(typeof(ExceptionDispatchInfo), nameof(ExceptionDispatchInfo.Capture), null,Expression.Property(catchParameterExpr, nameof(Exception.InnerException))), nameof(ExceptionDispatchInfo.Throw), null);
+
+                var tryCatch = Expression.TryCatch(invokeMethod,
```
mconnew, line 208: The try/catch should now disappear. It's only needed because calling `MethodInfo.Invoke` wraps the exception that's throw in in a `TargetInvocationException` and we need to unwrap it. If you use `Expression.Call` directly on the `methodInfo` without calling the `Invoke` method, it will directly throw the correct exception and we don't need to unwrap it.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +168,74 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                    Expression.Catch(catchParameterExpr,
+                        Expression.Block(
+                            throwCapturedExceptionDispatchInfoExpr
+                        )));
+
+                var assignOutputs =
```
mconnew, line 214: For the outputs, you would instantiate a local variable for each of the outs and wrap that in an Expression. You then add that in the relevant position in the `Expression[]` array that is passed to `Expression.Call`. After the method call, you use the same `Expression` instance representing the local variable and assign it into the output array.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done!

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: no longer there but inputPos array disappeared too in the rewritten version

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done. Please notice that we have to cast in/out around the method invocation with `Expression.Convert`

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: great!, done!

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +166,81 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+
+                    variables.Add(variable);
+                    invocationParameters.Add(variable);
+                }
+
+                var castTargetParam = Expression.Convert(targetParam, method.DeclaringType);
+
+                expressions.Add(returnsValue
+                    ? Expression.Block(Expression.Assign(result, Expression.Convert(Expression.Call(castTargetParam, method, invocationParameters), typeof(object))))
+                    : Expression.Block(Expression.Call(castTargetParam, method, invocationParameters),
```
mconnew, line 217: Do you need the Expression.Convert? Is this needed in the case where the return value is a value type so it needs to be boxed? If the return type is a value type, then I would expect it to assign to an object variable without any problem.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +166,81 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                foreach (var outputVariable in outputVariables)
+                {
+                    expressions.Add(Expression.Assign(
+                        Expression.ArrayAccess(outputsParam, Expression.Constant(j)),
+                        Expression.Convert(outputVariable, typeof(object))));
+                    j++;
+                }
+
+                expressions.Add(Expression.Block(
+                    Expression.Empty(),
```
mconnew, line 230: I think you can change this to just
```c#
expressions.Add(result);
```
As the finalBlock will return the value of the last expression. This is nesting a block inside a block and I don't think that's needed.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +166,81 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+
+                Expression<InvokeDelegate> lambda = Expression.Lambda<InvokeDelegate>(
+                    finalBlock,
+                    targetParam,
+                    inputsParam,
+                    outputsParam);
 
-                return lambda;
+                return lambda.Compile();
             }
```
mconnew, line 244: Can we add an ILogger with a special category to output the equivalent code to help diagnose any problems people might have. Log at the verbose level so most won't see it. See the [example code here](https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions.blockexpression?view=net-7.0#examples) to see what you should expect. I think you should be able to log what the equivalent code would be.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: I just retried without the Expression.Convert. I got the following exception (sorry french) System.ArgumentException: 'Impossible d'utiliser une expression de type 'System.Int32' pour une assignation au type 'System.Object''.
The contract was
```c#
    [System.ServiceModel.ServiceContract]
    public interface ICalculatorService
    {
        [System.ServiceModel.OperationContract]
        [System.ServiceModel.FaultContract(typeof(MathFault))]

        int Divide(int a, int b);
    }
```

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -84,12 +96,17 @@ internal static object CreateInstanceOfClass<T>() where T : class, new()
                 return default(T);
             }
 
-            internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int inputParameterCount, out int outputParameterCount)
+            internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int inputParameterCount, out int outputParameterCount) =>
+                s_isDynamicCodeSupported.Value
+                    ? GenerateInvokeDelegateInternalWithExpressions(method, out inputParameterCount, out outputParameterCount)
```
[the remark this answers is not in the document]
g7ed6e: done

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
g7ed6e: I need to figure out how to get an `ILogger` there as this code is not DI aware.

g7ed6e: @mconnew done! I'm bringing the ILogger through DispatchOperation

g7ed6e: Please see https://dev.azure.com/dotnet/CoreWCF/_build/results?buildId=98641&view=logs&j=d68afe5c-acae-5905-6bc4-d7e9ecfea404&t=2c035bf3-2a1f-5fc5-9550-cc6f70ae3aed&l=10893

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperation.cs
```diff
@@ -193,6 +193,8 @@ internal bool IsSessionOpenNotificationEnabled
 
         internal Lazy<AuthorizationPolicy> AuthorizationPolicy { get; set; }
 
+        internal IServiceProvider ServiceProvider { get; set; }
+
```
mconnew, line 197: I'm not liking this approach as DispatchOperation is publicly instantiable. If someone creates an instance and replaces the one CoreWCF created, then this can lead to a break. This is only being used to carry the IServiceProvider to OperationInvokerBehavior, but that doesn't need it. You can capture the IServiceProvider from OperationInvokerBehavior.AddBindingParameters from the BindingParameters. It's added to the binding parameters [here](https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Primitives/src/CoreWCF/Description/DispatcherBuilder.cs#L172)

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/TaskMethodInvoker.cs
```diff
@@ -270,7 +274,9 @@ private void EnsureIsInitialized()
             {
                 // Only pass locals byref because InvokerUtil may store temporary results in the byref.
                 // If two threads both reference this.count, temporary results may interact.
-                InvokeDelegate invokeDelegate = InvokerUtil.GenerateInvokeDelegate(TaskMethod, out int inputParameterCount, out int outputParameterCount);
+                using var scope = _serviceProvider.CreateScope();
+                ILogger<InvokeDelegate> logger = scope.ServiceProvider.GetRequiredService<ILogger<InvokeDelegate>>();
```
mconnew, line 278: I'm curious why you are creating a DI scope here? I wouldn't expect a logger to require a scope. Is it to make sure Dispose() is called on the logger if needed?

src/CoreWCF.Primitives/src/CoreWCF/Configuration/ServiceModelApplicationBuilderExtensions.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using CoreWCF.Channels;
+using CoreWCF.Dispatcher;
 using Microsoft.AspNetCore.Builder;
```
mconnew, line 9: You've added a using without code needed it.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntime.cs
```diff
@@ -112,6 +113,8 @@ internal bool HasDefaultUnhandledActionInvoker
 
         internal Lazy<AuthorizationPolicy> AuthorizationPolicy { get; }
 
+        internal IServiceProvider ServiceProvider { get; }
+
```
mconnew, line 117: I'm not seeing this being used anywhere.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
mconnew: The French is okay, I used to be reasonably fluent, and I understand what it's saying. I think the issue is because it's a value type needing to be boxed. Is the convert needed for a non-value type?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
mconnew: Can you add methodInfo.ToString() as well to the log output so we can see what the interface method is declared as?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
mconnew: I tracked down the source code and found the logic. If both are the same type, it's fine (doesn't matter if reference type of value type). If they types are different and either is a value type, it fails. If they are both reference types, then is uses Type.IsAssignableFrom. So it looks like as we're assigning to an object, you don't need to convert if it's a reference type, and only if it's a value type. [Code is here](https://github.com/dotnet/runtime/blob/a54ccf39b17586e7c2a38b882e80a3cc62d83d7d/src/libraries/System.Linq.Expressions/src/System/Dynamic/Utils/TypeUtils.cs#L919).

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/InvokerUtil.cs
```diff
@@ -150,8 +173,113 @@ internal static InvokeDelegate GenerateInvokeDelegate(MethodInfo method, out int
...
+                    expressions.Add(Expression.Call(castTargetParam, method, invocationParameters));
+                    expressions.Add(Expression.Assign(result, Expression.Constant(null, typeof(object))));
+                }
+
+                int j = 0;
+                foreach (var outputVariable in outputVariables)
+                {
+                    expressions.Add(Expression.Assign(
+                        Expression.ArrayAccess(outputsParam, Expression.Constant(j)),
+                        Expression.Convert(outputVariable, typeof(object))));
```
mconnew, line 245: You can avoid the Convert call here as well when the outputVariable is not a value type.

g7ed6e: good catch! i updated

mconnew (approved):

PR #1227, opened 2023-09-20
Add OperationInvoker source generator

g7ed6e:
1. Add a source generator to generate an `IOperationInvoker` implementation per method marked as `[OperationContract]`. The implementation are registered within a static `Dictionary<string, IOperationInvoker>`
2. Add the ability to enable the source generator in consumer projects by adding
```xml
<PropertyGroup>
  <EnableCoreWCFOperationInvokerGenerator>true</EnableCoreWCFOperationInvokerGenerator>
</PropertyGroup>
```
3. Update `OperationInvokerBehavior` to pull the `IOperationInvoker` from the static dictionary when the source generator feature is active.
4. Add unit tests of the source gen
5. Add integration tests by referencing the source gen in CoreWCF.Http.Tests and set the project property to true. ~by duplicating the CoreWCF.Http.Tests project and enabling the source gen to ensure correct lookup in the dictionary~


g7ed6e:
todo: check that package props and targets are executed when CoreWCF.Primitives is a transitive dependency.

g7ed6e:
TODO: Use same approach as #1240 to avoid PathTooLongException

src/CoreWCF.Http/tests/Authorization/AuthorizationIntegrationTests.Service.cs
```diff
@@ -10,7 +10,7 @@ namespace CoreWCF.Http.Tests.Authorization;
 public partial class AuthorizationIntegrationTests
 {
     [ServiceContract]
-    private interface ISecuredService
+    internal interface ISecuredService
     {
```
mconnew, line 14: I presume that private interfaces don't work with a source generated invoker (which makes sense). What happens in that case? What do you think we should do? Turning on source generation of the invoker is an opt-in feature, so we could fail the build, or we could emit a warning that we couldn't generate the invoker for this type, or we could just not attempt to emit anything. I don't like that last option though as it makes discovery difficult.

src/CoreWCF.BuildTools/src/OperationInvokerGenerator.Emitter.cs
```diff
@@ -0,0 +1,233 @@
...
+                    builder.Append(operationContractSpec.Method!.ToDisplayString());
+                    builder.Replace(".", "_")
+                        .Replace("<", "_")
+                        .Replace(">", "_")
+                        .Replace("(", "_")
+                        .Replace(")", "_")
+                        .Replace("[", "_")
+                        .Replace("]", "_")
+                        .Replace(", ", "_")
+                        .Replace(" ", "_");
```
mconnew, line 206: I think it would be worth using something else to either replace comma or a space otherwise you'll get a collision between these two methods:
```
System.String Echo(System.String input, MyNameSpace.Ref refParam,Boolean b, out System.Int32 i);
System.String Echo(System.String input, ref System.Boolean b, out System.Int32 i);
```

src/CoreWCF.Http/tests/CoreWCF.Http.SourceGeneration.Tests.csproj
```diff
@@ -0,0 +1,79 @@
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
```
mconnew, line 3: Would it be worth importing the regular CoreWCF.Http.Tests.csproj file instead of duplicating everything here? I don't want things to get out of sync

src/CoreWCF.BuildTools/src/OperationInvokerGenerator.Emitter.cs
```diff
@@ -0,0 +1,233 @@
...
+                    _builder.AppendLine("Array.Empty<object>();");
+                }
+                else
+                {
+                    _builder.AppendLine($"new object[{outputParameterCount}];");
+                }
+                _builder.AppendLine();
+
+                _builder.AppendLine($"{indentor}[System.Runtime.CompilerServices.ModuleInitializer]");
+                _builder.Append($"{indentor}internal static void RegisterOperationInvoker() => ");
```
mconnew, line 185: I'm concerned about having a separate static method for each and every registered operation. I've seen contracts with thousands of operations per interface. We should at least only have one module initializer per contract, but ideally just a single one for the entire project. It's all generated into a single source file so the `file` accessor means all of the generated types can be referenced from a single method.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntimeHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    internal static string GetKey(MethodInfo method)
+    {
+        StringBuilder stringBuilder = new StringBuilder($"{method.DeclaringType.FullName}.{method.Name}(");
+        stringBuilder.Append(string.Join(", ", method.GetParameters().Select(GetParameterString)));
+        foreach (var keyValuePair in s_IntegratedTypesMap)
+        {
+            stringBuilder.Replace(keyValuePair.Key, keyValuePair.Value);
+        }
```
mconnew, line 28: This is potentially very expensive. It would be cheaper to do this in GetParameterFullName. Then you could lookup the type in the dictionary instead of the type name. If this is going to be aot only, then you can use Dictionary<type,string>, but if it's going to be used in regular jit scenarios too, Type might not be stable when using hot reload. I know we had a problem with MethodInfo in the WCF client when using hot reload. In that case, use Type.TypeHandle.Value to get the type handle value and use Dictionary<IntPtr, string>

src/CoreWCF.BuildTools/src/OperationInvokerGenerator.Emitter.cs
mconnew: I found another collision. You will get a collision between these two I believe:
```c#
void Echo_bool(string value);
void Echo(bool isit, string value);
```

mconnew: Should we autogenerate the invoker class name to not correspond to the actual operation method to avoid this problem. Add a comment in the generated code to say which method it's for, although you can see that in the implementation method anyway, but it's easier if it's in a comment.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntimeHelpers.cs
g7ed6e: done here https://github.com/CoreWCF/CoreWCF/pull/1227/commits/75c082b1d373ce0322eb9eff810f58aa45ef1180

src/CoreWCF.Http/tests/CoreWCF.Http.SourceGeneration.Tests.csproj
g7ed6e: done

src/CoreWCF.BuildTools/src/OperationInvokerGenerator.Emitter.cs
g7ed6e: I'm now generating class name like `OperationInvoker0`, `OperationInvoker2`, `OperationInvoker3`.. and a comment describes the method which is called.

src/CoreWCF.BuildTools/src/OperationInvokerGenerator.Emitter.cs
g7ed6e: I kept a `RegisterOperationInvoker` static method per `OperationInvoker` implementation but this method is no longer attributed with `ModuleInitializer`. 
Instead An `OperationInvokerModuleInitializer` class is generated at the end of the file and this class exposes a unique `ModuleInitializer` attributed method that call all the `RegisterOperationInvoker` method of the generated `OperationInvoker` types.

src/CoreWCF.Http/tests/Authorization/AuthorizationIntegrationTests.Service.cs
g7ed6e: Current behavior is that we got a CS0122 build error because we do generate the code.
image

I see 3 potential options:
1. do nothing, the build is broken but the compiler error is not explicit about the source generator.
2. detect the case in the source gen (and not emit the OperationInvoker implementation) and emit a build error more meaningful for user.
3. detect the case in the source gen (and not emit the OperationInvoker implementation) and emit a build warning more meaningful for user. (this case will lead to runtime exceptions). 

As this feature mainly target AOT scenario where the difficulty is to track potential runtime issue at build time I would go for either 1. or 2.

Note: The actual behavior of the `EnableCoreWCFOperationInvokerGenerator` project property is that we add a check in `src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs` so we trigger a PlatformNotSupportedException if the feature is enabled and we cannot find the implementation in the dictionary that contains registered implementations. We may also have more finegrained behavior and introduce a fallback towards the current non generated OperationInvoker implemenation. This finegrained behavior could be controlled by having values like "Always" / "IfPossible" / "Never" in `EnableCoreWCFOperationInvokerGenerator`. But i'm not sure we actually need this.

mconnew: I believe the original code will work with aot, it just has bad performance characteristics. The DataContractSerializer has to use MethodInfo.Invoke. I don't like the idea of a full break unless there's a way to override it. There are a few reasons why that might be okay. It could be a contract interface only used by a WCF client so doesn't need an invoker generated. It could be an interface in a shared library which isn't used to expose an endpoint in the current project.  

Could we make it an analyzer which creates an error at the point of use of the contract interface? That allows the developer to put in an ignore override in the code with a disable pragma, or even at the project level. This means if the interface isn't used anywhere, no error and it's silently ignored. We can't just worry about use with AddServiceEndpoint though as I've seen developers use a generic class to host multiple services/contracts so AddServiceEndpoint might not have the contract type at the call point.

g7ed6e: > I believe the original code will work with aot, it just has bad performance characteristics. The DataContractSerializer has to use MethodInfo.Invoke. I don't like the idea of a full break unless there's a way to override it. There are a few reasons why that might be okay. It could be a contract interface only used by a WCF client so doesn't need an invoker generated. It could be an interface in a shared library which isn't used to expose an endpoint in the current project.
> 
> Could we make it an analyzer which creates an error at the point of use of the contract interface? That allows the developer to put in an ignore override in the code with a disable pragma, or even at the project level. This means if the interface isn't used anywhere, no error and it's silently ignored. We can't just worry about use with AddServiceEndpoint though as I've seen developers use a generic class to host multiple services/contracts so AddServiceEndpoint might not have the contract type at the call point.

I updated the analyzer according to this requirement. We nos output a build error at the location the interface is used if the interface is private.

src/CoreWCF.BuildTools/tests/OperationInvokerGeneratorTests.cs
```diff
@@ -0,0 +1,1046 @@
...
+                    {
+@$"
+namespace MyProject
+{{
+    [{attributeNamespace}.ServiceContract]
+    public interface IIdentityService
+    {{
+        [{attributeNamespace}.OperationContract]
+        string Echo(params string[] inputs);
+    }}
```
mconnew, line 692: I didn't know you could use a params array with WCF!!

src/CoreWCF.Http/tests/ServiceWithSSMFaultContractTest.cs
```diff
@@ -19,13 +19,6 @@
 using Xunit;
 using Xunit.Abstractions;
 
-// Needed to use WebApplicationFactory on Net472
-[assembly: WebApplicationFactoryContentRoot(
```
mconnew, line 23: Is this not needed?

src/CoreWCF.Primitives/src/CoreWCF.Primitives.props
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <ItemGroup>
+    <!-- Exposes EnableCoreWCFOperationInvokerGenerator at build time -->
+    <CompilerVisibleProperty Include="EnableCoreWCFOperationInvokerGenerator" />
```
mconnew, line 5: >CompilerVisibleProperty [](http://example.com/codeflow?start=5&length=23)

We're generating C# code. What happens to both this new implementation and the existing injection generator when using with a vb project?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntimeHelpers.cs
```diff
@@ -0,0 +1,123 @@
...
+        stringBuilder.Append(")");
+        var result = stringBuilder.ToString();
+        return result;
+    }
+
+    private static string GetParameterString(ParameterInfo p)
+    {
+        StringBuilder sb = new();
+        //bool removeLastChar = false;
+        Type parameterType = p.ParameterType;
```
mconnew, line 35: remove?

templates/TestStage.yml
```diff
@@ -87,7 +87,7 @@ stages:
 
       - task: DotNetCoreCLI@2
         displayName: Run Tests
-        timeoutInMinutes: 20
+        timeoutInMinutes: 25
         inputs:
```
mconnew, line 91: Does this still need increasing with the new xunit version? NetFx should be the worst case as we disable parallelism and in my PR this completed in 12 mins 31 seconds

mconnew:
This looks almost ready, just a couple of Qs.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/DispatchOperationRuntimeHelpers.cs
g7ed6e: done

src/CoreWCF.BuildTools/tests/OperationInvokerGeneratorTests.cs
g7ed6e: I dontt know if it's supported at runtime either. If it is, the source gen will support it too.

src/CoreWCF.BuildTools/src/OperationParameterInjectionGenerator.SourceGenerationSpec.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace CoreWCF.BuildTools
+{
```
mconnew, line 8: Inconsistent use of file-scoped namespaces. Some new files have braces like this one, and some have no braces like PrivateServiceContractAnalyzer. What are you thoughts on consistency through the project? I like adopting newer cleaner ways of writing code, but I'd really like everything at least within a single project to be consistent. Should we update the entire project (not a massive project so not a big effort)? Or have new files use this older braces style? Or do you think it's not a big deal to be mixed?

g7ed6e: > Inconsistent use of file-scoped namespaces. Some new files have braces like this one, and some have no braces like PrivateServiceContractAnalyzer. What are you thoughts on consistency through the project? I like adopting newer cleaner ways of writing code, but I'd really like everything at least within a single project to be consistent. Should we update the entire project (not a massive project so not a big effort)? Or have new files use this older braces style? Or do you think it's not a big deal to be mixed?

I think the solution does contain old style c# syntax (I'm thinking to code that has been ported to modern .NET) that would deserve a refresh but can live as is as it's easing comparison to .NetFX or SSM repo. 
However for new code like the one from BuildTools we should be consistent across the project and aligned with today's standards. I will fix that.

src/CoreWCF.Http/tests/CoreWCF.Http.Tests.csproj
```diff
@@ -49,4 +49,10 @@
     <ProjectReference Include="..\src\CoreWCF.Http.csproj" />
     <ProjectReference Include="..\..\CoreWCF.BuildTools\src\CoreWCF.BuildTools.Roslyn4.0.csproj" OutputItemType="Analyzer" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'net8.0'">
```
mconnew, line 53: When doing this, I like to include the files with a \<None\> item so they still show up in VS solution explorer.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs
```diff
@@ -12,8 +13,24 @@ public class OperationInvokerBehavior : IOperationBehavior
     {
         private IServiceProvider _serviceProvider;
 
+        private static readonly Lazy<bool> s_isOperationInvokerGeneratorEnabled = new Lazy<bool>(() =>
+        {
+            var assemblies = AppDomain.CurrentDomain.GetAssemblies();
+            foreach (var assembly in assemblies)
+            {
+                EnableCoreWCFOperationInvokerGeneratorAttribute attribute = assembly.GetCustomAttribute<EnableCoreWCFOperationInvokerGeneratorAttribute>();
```
mconnew, line 21: Use GetCustomAttributesData and not GetCustomAttribute as GetCustomAttribute instantiates an instance of the attribute. I'm also a little concerned about this approach of searching all assemblies currently loaded. You could reference a library which had this turned on and you aren't using any contracts from that library, and the app you compiled, you didn't turn it on. There would be no way to use that library without being forced to enable this in your main app. The only assembly you have definitely have control of is the main app assembly. What do you think about only paying attention to the attribute on main executing assembly? If everything is in a library, enable it there, but you also need to turn it on in your main app?

mconnew: I realized this won't work for xunit tests as the executing assembly will be the xunit test runner. We can't just hard code in something for xunit though as there are multiple different unit test frameworks. Need to give this some thought. The criteria is we need to work if there's a mix of assemblies where some have this enabled and some don't. What about not having a big global switch like this, but instead check the assembly for interface method. Could we have OperationDescription be populated with a bool saying whether to use a pre-generated invoker or not? You need to handle interface hierarchies, ie IService1 in Assembly1.dll was compiled with it, and IService2 derived from IService1 in Assembly2 was compiled without it. So methods IService1 have pregenerated invokers, but methods in IService2 don't. Or vice versa. Do you have any other ideas on how to handle this scenario?

g7ed6e: The code generation feature has to be enabled at service contract assembly level to generate code in the correct assembly.
However the usage of the generated code should be enabled at application host level. Would you be open to add an AppContext.Switch to force enabling usage of the generated code in unit tests ?

mconnew: That makes sense and keeps things simple. Go with the AppContext override.

g7ed6e: Great! PR is updated.

src/CoreWCF.Primitives/src/CoreWCF/EnableCoreWCFOperationInvokerGeneratorAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace CoreWCF
+{
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class EnableCoreWCFOperationInvokerGeneratorAttribute : Attribute
+    {
+        public string Value { get; }
+        public EnableCoreWCFOperationInvokerGeneratorAttribute(string value)
```
mconnew, line 14: Is the type of Value a string to enable more nuanced options than true and false in the future?

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs
```diff
@@ -12,8 +14,44 @@ public class OperationInvokerBehavior : IOperationBehavior
...
+            if (entryAssembly != null)
+            {
+                foreach (var attribute in entryAssembly.GetCustomAttributesData())
+                {
+                    if (attribute.AttributeType == typeof(EnableCoreWCFOperationInvokerGeneratorAttribute))
+                    {
+                        foreach (var argument in attribute.NamedArguments)
+                        {
+                            if (
+                                argument.TypedValue.ArgumentType == typeof(string)
```
mconnew, line 35: An easier way to see if this is the correct parameter is to do:
```c#
if (argument.MemberName == "Value")
{
    var stringValue = argument.TypedValue.Value as string;
    return string.Equals(stringValue, "true", StringComparison.OrdindalIgnoreCase);
}
```
This code only works if you use `[EnableCoreWCFOperationInvokerGenerator(Value = "true")]`, but you've also added a constructor parameter so you can also use `[EnableCoreWCFOperationInvokerGenerator("true")]` which this code won't work for. After completing the foreach around the NamedArguments, you also need to iterate over ConstructorArguments. Alternatively you could limit setting it to just one way, so remove the constructor argument from the class and require it be set via the named property as you've already got this code. You could also switch to only using ConstructorArguments and remove the property. We don't need to actually store the value as when using CustomAttributesData, you're reading out the code declaration and not actually instantiating an instance of the attribute.

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs
```diff
@@ -38,27 +76,60 @@ void IOperationBehavior.ApplyDispatchBehavior(OperationDescription description,
...
-                    //}
+                    if (DispatchOperationRuntimeHelpers.OperationInvokers.TryGetValue(DispatchOperationRuntimeHelpers.GetKey(description.SyncMethod), out IOperationInvoker invoker))
+                    {
+                        dispatch.Invoker = invoker;
+                    }
+                    else
+                    {
+                        // reaching this point means that the operation invoker generator is enabled but we did not succeed to fetch the invoker
+                        // for the operation. This is a bug in the DispatchOperationRuntimeHelpers.GetKey logic.
+                        throw new PlatformNotSupportedException();
```
mconnew, line 110: It could also be that the contract lives in a separate assembly from the main application assembly and generation of the invoker wasn't enabled in that other assembly. I think we should throw an InvalidOperationException and have a string resource saying if the contract interface is in a different assembly than the application, generation of the operation invoker must be enabled when compiling the contract interface assembly. This is a lot to describe in a string resource. I wonder if we should have a KB section at [https://corewcf.github.io/kb/[num](https://corewcf.github.io/kb/[num)] where we could document how to handle a particular error case. We shouldn't block on that for this PR, but it's something I think we should do.

src/CoreWCF.Primitives/src/CoreWCF/EnableCoreWCFOperationInvokerGeneratorAttribute.cs
g7ed6e: updated to use a boolean value

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs
g7ed6e: updated to use a ctor boolean arg. I reproduced what is done with common atributes like FileVersionAttribute.
```c#
namespace System.Reflection
{
    [AttributeUsage(AttributeTargets.Assembly, Inherited = false)]
    public sealed class AssemblyFileVersionAttribute : Attribute
    {
        public AssemblyFileVersionAttribute(string version)
        {
            Version = version ?? throw new ArgumentNullException(nameof(version));
        }

        public string Version { get; }
    }
}
```

src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/OperationInvokerBehavior.cs
g7ed6e: created #1417 to track creation of the KB section at corewcf.github.io

g7ed6e:
@mconnew This PR has been rebased with latest changes.

mconnew (approved):

Issue #1213, opened 2023-09-08
No-op issue to re-validate


birojnayak:
>            var unixDomainSocketFramingOptionsSetup = _serviceProvider.GetServices<IConfigureOptions<KestrelServerOptions>>().SingleOrDefault(options => options is  [](http://example.com/codeflow?start=0&length=164)

Looks like this is a no-op and can be removed.

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/1104#discussion_r1296571261_

PR #1262, opened 2023-10-24
Changes for CoreWCF UDS

### LINKED ISSUE #1214: Formatting fix for file UnixDomainSocketTransportSecurity
This didn't get fixed.

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/1104#discussion_r1306228397_

birojnayak:
Also fix #1214  , #1213

### DIFF
--- a/src/CoreWCF.UnixDomainSocket/src/CoreWCF/Channels/UnixDomainSocketHostedService.cs
+++ b/src/CoreWCF.UnixDomainSocket/src/CoreWCF/Channels/UnixDomainSocketHostedService.cs
@@ -35,13 +35,6 @@ namespace CoreWCF.Channels
             _logger = logger;
             _serviceProvider = serviceProvider;
             IApplicationLifetime appLifetime = _serviceProvider.GetRequiredService<IApplicationLifetime>();
-            Init();
-        }
-
-        private void Init()
-        {
-            // Check if Kestrel is already being used and if it is, it was already configured so there's nothing to do.
-            var unixDomainSocketFramingOptionsSetup = _serviceProvider.GetServices<IConfigureOptions<KestrelServerOptions>>().SingleOrDefault(options => options is UnixDomainSocketFramingOptionsSetup) as UnixDomainSocketFramingOptionsSetup;
         }
 
         public async Task StartAsync(CancellationToken cancellationToken)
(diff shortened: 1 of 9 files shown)


src/Common/src/CoreWCF/Resources/Strings.resx
```diff
@@ -174,4 +174,7 @@
   <data name="CommunicationObjectCannotBeUsed" xml:space="preserve">
     <value>The communication object, {0}, is in the {1} state.  Communication objects cannot be used for communication unless they are in the Opened state.</value>
   </data>
+    <data name = "UnsupportedSecuritySetting" xml:space="preserve">
+    <value>The value '{1}' is not supported in this context for the binding security property '{0}'.</value>
```
mconnew, line 178: You added this to the wrong string resource file. It should have been added to the UDS specific one

src/CoreWCF.UnixDomainSocket/src/CoreWCF/Channels/UnixDomainSocketHostedService.cs
```diff
@@ -41,7 +41,7 @@ public UnixDomainSocketHostedService(IServiceBuilder serviceBuilder, ILogger<Uni
         private void Init()
         {
             // Check if Kestrel is already being used and if it is, it was already configured so there's nothing to do.
-            var unixDomainSocketFramingOptionsSetup = _serviceProvider.GetServices<IConfigureOptions<KestrelServerOptions>>().SingleOrDefault(options => options is UnixDomainSocketFramingOptionsSetup) as UnixDomainSocketFramingOptionsSetup;
+          //  var unixDomainSocketFramingOptionsSetup = _serviceProvider.GetServices<IConfigureOptions<KestrelServerOptions>>().SingleOrDefault(options => options is UnixDomainSocketFramingOptionsSetup) as UnixDomainSocketFramingOptionsSetup;
```
mconnew, line 44: >          //  var unixDomainSocketFramingOptionsSetup = _serviceProvider.GetServices<IConfigureOptions<KestrelServerOptions>>().SingleOrDefault(options => options is UnixDomainSocketFramingOptionsSetup) as  [](http://example.com/codeflow?start=0&length=206)

Remove instead of leaving commented out. You should be able to remove this entire method and remove the call to it.

mconnew:
Can you add a test to make sure that the security mode and client credential type incompatibilities throw? It can just be a unit test which sets the relevant properties on the binding and calls `CreateBindingElements()`, you don't need an entire service startup to validate it.

PR #1277, opened 2023-11-06
Add KafkaMessageProperty

### LINKED ISSUE #1275: [Feature Request] [Kafka]: Expose Kafka Message properties to CoreWCF
**Is your feature request related to a problem? Please describe.**
A Kafka message on the wire is composed of 
- a key (exposed as a generic type in the .NET api). The key is commonly used in conjonction with the default partioner to ensure message are processed in FIFO order when a topic have multiples partitions as messages with the same key are delivered to the same consumer instance of a given consumer group. The current implementation of both CoreWCF.Kafka / CoreWCF.Kafka.Client ignore the message key
- a value (exposed as a generic type in the .NET api). The value is the raw payload of the message, the current implementation uses the value to transport the SOAP message on the wire.
- headers  (exposed a Dictionary<string, byte[]> in the .NET api). The headers are metadata / additional data joined to a message and may contain information about the value being serialized. The current implementation ignores headers.

**Describe the solution you'd like**
These values should be accessibles from the OperationContext (something similar to the way HTTP headers can be added/retrieved with BasicHttpBinding).
Also at consumer side, when receiving a message, the ReceiveContext.Properties should contain additional information about the received message so a custom IDispatchMessageFormatter can use the headers to deserialize the payload and route it to a specific operation.

**Describe alternatives you've considered**
At consumer side, the ReceiveContext instance is actually an KafkaReceiveContext which is a CoreWFC internal type. This type does contain the ConsumeResult<TKey, TValue> which is a Confluent.Kafka type. Using reflection it could possible to get these information but the code would 
- be using reflection
- rely on a Confluent.Kafka type
At producer side there is no hook to send the key or custom Kafka headers on the wire.

g7ed6e:
Introduce a KafkaMessageProperty both client side and server side to be able to read/write kafka message properties.

Resolves #1275 

Note: this implementation is agnostic of SOAP.


src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageHeader.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace CoreWCF.Channels;
+
+public sealed class KafkaMessageHeader
+{
+    public KafkaMessageHeader(string key, byte[] value)
+    {
```
mconnew, line 9: Should you represent the Kafka body as byte[] or Memory\<byte\>? Maybe even ReadOnlyMemory\<byte\>?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
```diff
@@ -0,0 +1,32 @@
...
+using Confluent.Kafka;
+
+namespace CoreWCF.Channels;
+
+public sealed class KafkaMessageProperty : IMessageProperty
+{
+    private readonly IList<KafkaMessageHeader> _headers = new List<KafkaMessageHeader>();
+
+    public static readonly string Name = "kafkaMessage";
+
```
mconnew, line 14: `RemoteEndpointMessageProperty.Name` has value `"CoreWCF.Channels.RemoteEndpointMessageProperty"`, but `HttpRequestMessageProperty.Name` has value `"httpRequest"`. It's a bit erratic and inconsistent, but I like to try to avoid name clashes. When adding `HttpContext` to the message properties I used the name `"Microsoft.AspNetCore.Http.HttpContext"`. I think it's a good coding guideline that we should adopt that message property names should be the class full name. So in this case, `"CoreWCF.Channels.KafkaMessageProperty"`

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            _headers.Add(new KafkaMessageHeader(messageHeader.Key, messageHeader.GetValueBytes()));
+        }
+
+        PartitionKey = consumeResult.Message.Key;
+        Topic = consumeResult.Topic;
+    }
+
+    public IMessageProperty CreateCopy() => this;
+
```
mconnew, line 27: If you are going to just return `this`, then drop implementing the interface. This is used when adding a property to clone it if needed so if something in the original is disposed, the copy is still intact. When a message property doesn't implement `IMessageProperty`, it just stores the property as is, like you're doing.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
```diff
@@ -0,0 +1,32 @@
...
+    public IReadOnlyCollection<KafkaMessageHeader> Headers => _headers as IReadOnlyCollection<KafkaMessageHeader>;
+    public byte[] PartitionKey { get; }
+
+    public string Topic { get; }
```
mconnew, line 31: Nit: inconsistent spacing. I like to leave no spaces between single line properties so probably best to remove this line rather than add one in above.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaQueueMessageContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using CoreWCF.Queue.Common;
+
+namespace CoreWCF.Channels;
+
+public class KafkaQueueMessageContext : QueueMessageContext
+{
```
mconnew, line 10: Do you need this class when QueueMessageContext has a Properties property which you've fixed the lazy constructor for?

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueMessageContext.cs
```diff
@@ -18,7 +18,7 @@ public class QueueMessageContext : RequestContext
         private Message _requestMessage;
         private Exception _requestMessageException;
         private ReceiveContext _receiveContext;
-        private readonly Lazy<IDictionary<string, object>> _properties = new();
+        private readonly Lazy<IDictionary<string, object>> _properties = new(() => new Dictionary<string, object>());
```
mconnew, line 22: You could have changed the Lazy field to be a `Lazy<Dictionary<string,object>>` instead so that the default constructor will be used without providing a delegate. The Lazy doesn't need to be the exact same type as the Properties property.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageHeader.cs
g7ed6e: Done. I changed the KafkaMessageProperty.PartitionKey property type to ReadOnlyMemory<byte> too. 
Client side I kept byte[] as this the type we need to provide to Confluent.Kafka.

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
g7ed6e: Done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaQueueMessageContext.cs
g7ed6e: I no longer need it indeed. good catch.

src/CoreWCF.Queue/src/CoreWCF/Queue/Common/QueueMessageContext.cs
g7ed6e: done

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageProperty.cs
g7ed6e: done

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaMessageProperty.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace CoreWCF.ServiceModel.Channels;
+
+public sealed class KafkaMessageProperty
+{
+    public static readonly string Name = "CoreWCF.ServiceModel.Channels.KafkaMessageProperty";
+
```
mconnew, line 12: With the idea I discussed via chat about making properties injectable, if that's something we're going to pursue, this would be better as a const string rather than a static field.

mconnew: Nevermind for this file, just realized this is the client. But the comment still stands for the service equivalent.

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
```diff
@@ -136,6 +138,22 @@ private async Task SendAsync(Message message, TimeSpan timeout)
...
+                    ? property
+                    : new KafkaMessageProperty();
+            kafkaMessage.Headers = new();
+            foreach (KafkaMessageHeader kafkaMessageHeader in kafkaMessageProperty.Headers)
+            {
+                kafkaMessage.Headers.Add(kafkaMessageHeader.Key, kafkaMessageHeader.Value);
+            }
+
+            kafkaMessage.Key = kafkaMessageProperty.PartitionKey;
+        }
```
mconnew, line 155: Is setting a null value here okay?

mconnew (approved):

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
g7ed6e: Yes the existing unit tests cover that case.
Please see [here](https://github.com/confluentinc/confluent-kafka-dotnet/blob/a67bd6c06b7eef4293e6476d9ff6f3e93f0e4cd9/src/Confluent.Kafka/Serializers.cs#L170) for serialization side and [here](https://github.com/confluentinc/confluent-kafka-dotnet/blob/a67bd6c06b7eef4293e6476d9ff6f3e93f0e4cd9/src/Confluent.Kafka/Deserializers.cs#L243) for deserialization side.

PR #1282, opened 2023-11-08
Fix CoreWCFWebApplicationExtensions when Microsoft.AspNetCore.App FrameworkReference isn't included

mconnew:
Tested with 2 projects, a library project with contract and implementation referencing CoreWCF packages, but no references to asp.net core, and a service app which launches an asp.net core app which doesn't reference the CoreWCF packages and inherits them transitively.
I tested with the library targeting netstandard2.0, and multi-targeting net6.0 and net472. Also tested with the service app targeting net6.0 and net472, as well as net6.0 using a FrameworkReference for  Microsoft.AspNetCore.App, or a PackageReference for Microsoft.AspNetCore, as well as using the Sdk Microsoft.NET.Sdk.Web. It appears to work correctly in all scenarios now.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
+++ b/src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
@@ -1,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <ItemGroup Condition="'$(_CoreWCFSuppressWebApplicationExtensions)' != 'true' AND
+                        '$(TargetFramework.TrimEnd(`0123456789`))' != 'net' AND
+                        !$(TargetFramework.StartsWith('netstandard')) AND
+                        (@(FrameworkReference->WithMetadataValue('Identity', 'Microsoft.AspNetCore.App')->Count()) != 0)">
+    <Compile Include="$(MSBuildThisFileDirectory)/../contentFiles/CoreWCFWebApplicationExtensions.cs">
+      <Visible>false</Visible>
+    </Compile>
+  </ItemGroup>
+</Project>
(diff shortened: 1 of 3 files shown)


src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <ItemGroup Condition="'$(_CoreWCFSuppressWebApplicationExtensions)' != 'true' AND
```
g7ed6e, line 3: Can `$(_CoreWCFSuppressWebApplicationExtensions)` be `$(CoreWCFSuppressWebApplicationExtensions)` (without `_`) ?

g7ed6e (approved):

src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
mconnew: Generally properties starting with an underscore are for internal purposes and aren't intended as a build api. My thinking here is that it's not intended as something people need to be using and is what I call a "get out of jail free" card that we only inform people of if there's any further problems we didn't expect, with the intention it would be a workaround until we fixed whatever problem they are having.  
I don't feel strongly about this, so if you still feel it shouldn't have the underscore, then I'll remove it.

g7ed6e: Ok i was wondering if the _ prefix is a magic feature of msbuild but I see this is just a naming convention. That's fine to me to keep it private as it should not be used in theory.

PR #1284, opened 2023-11-08
Refactor/drop useless code from source generator ut infrastructure

g7ed6e:


mconnew:
As you were able to simplify the 3.11 and the 4.0 tests, I presume this was because you learnt more about how to write these tests rather than any new changes to the source generators?

g7ed6e:
> As you were able to simplify the 3.11 and the 4.0 tests, I presume this was because you learnt more about how to write these tests rather than any new changes to the source generators?

Exactly these changes reduce our test related infrastructure code

PR #1286, opened 2023-11-08
Add IncomingMessageProperty injection support in source generator

g7ed6e:
~~Requires #1277~~


mconnew:
You could use RemoteEndpointMessageProperty for the tests which would work for HTTP and NeTcp so wouldn't need to add Kafka as a dependency. I'm okay leaving it as is too if you prefer.

g7ed6e:
> You could use RemoteEndpointMessageProperty for the tests which would work for HTTP and NeTcp so wouldn't need to add Kafka as a dependency. I'm okay leaving it as is too if you prefer.

Done. I added integration tests too by referencing the source gen from NetTcp, Http and Kafka UT projects

src/CoreWCF.Http/tests/InjectedIncomingPropertiesTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            services.AddServiceModelServices();
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseServiceModel(builder =>
+            {
+                builder.AddService<Services.EchoService>();
+                builder.AddServiceEndpoint<Services.EchoService, ServiceContract.IEchoService>(new CoreWCF.BasicHttpBinding(), "/BasicWcfService/basichttp.svc");
+            });
```
mconnew, line 57: Should this test be using InjectedIncomingPropertiesService/IInjectedIncomingPropertiesService instead of EchoService/IEchoService?

src/CoreWCF.BuildTools/src/SourceGenerationSpec.cs
```diff
@@ -14,15 +14,23 @@ namespace CoreWCF.BuildTools
         INamedTypeSymbol? TaskSymbol,
         INamedTypeSymbol? GenericTaskSymbol,
         INamedTypeSymbol? CoreWCFInjectedSymbol,
-        INamedTypeSymbol? MicrosoftAspNetCoreMvcFromServicesSymbol)
+        INamedTypeSymbol? MicrosoftAspNetCoreMvcFromServicesSymbol,
+        INamedTypeSymbol? RemoteEndpointMessagePropertySymbol,
+        INamedTypeSymbol? KafkaMessagePropertySymbol,
```
mconnew, line 19: We shouldn't be special casing the individual property types. There are many more property types than listed here and we shouldn't be limited to these types. This should be able to be implemented in a general way.

mconnew:
If the message property is missing, or the wrong type, we should pass null to the method. This matches the semantics of calling `IncomingMessageProperties.TryGetValue`. There are scenarios where a message property might not exist and that's okay. For example, someone could have an implementation which has multiple endpoints with different bindings so different message properties, and a specific message property might only exist on some of them.  

We should also not have the code generator care about specific message property types. The code should be general purpose. Everything you need is in the method signature and the Injected attribute. You know the name of the property from the Injected.PropertyName property, and the type of the property from the method parameter type.

src/CoreWCF.Http/tests/InjectedIncomingPropertiesTests.cs
g7ed6e: Good catch thanks

src/CoreWCF.BuildTools/src/Extensions.cs
```diff
@@ -46,6 +46,22 @@ internal static class ParameterSymbolExtensions
...
             => SymbolEqualityComparer.Default.Equals(symbol.Type, parameterSymbol.Type);
+
+        public static string? GetInjectedPropertyNameValue(this IParameterSymbol symbol, INamedTypeSymbol? coreWcfInjectedAttributeSymbol)
+        {
+            foreach (AttributeData attributeData in symbol.GetAttributes())
+            {
+                if (SymbolEqualityComparer.Default.Equals(attributeData.AttributeClass, coreWcfInjectedAttributeSymbol))
+                {
+                    if (attributeData.NamedArguments.Length > 0 && attributeData.NamedArguments[0].Key == "PropertyName")
+                    {
```
mconnew, line 57: This will be fragile if we add any more properties to the attribute. Maybe use something like:
```c#
var propNameArg = attributeData.NamedArguments.Where(na => na.Key == "PropertyName").FirstOrDefault();
if (propNameArg != default)
{
    return propNameArg.Value!.ToString();
}
```
What happens if someone sets the PropertyName to null? Does it still show up in the NamedArguments? Will `Value!.ToString()` throw a NRE? What type is the Value? Is it the type of the actual property, ie a string? If so, would a cast be better? That way you would return a null. Do we want to differentiate the difference between not specifying the PropertyName vs setting it to null? If they set it to null, my thinking is that should yield a compilation error.

src/CoreWCF.BuildTools/src/Extensions.cs
```diff
@@ -123,4 +139,23 @@ public static string ToSafeCSharpString(this TypedConstant typedConstant)
...
+        public static string? ToMessagePropertyVariableName(this string? propertyName)
+        {
+            if (propertyName is null)
+            {
+                return null;
+            }
+
+            var parts = propertyName.Split('.');
+            return parts.Last().ToCamelCase();
+        }
```
mconnew, line 154: Why are we forcing this to be camel case? Someone could put in their own property with any casing they like.

src/CoreWCF.BuildTools/src/Extensions.cs
mconnew: I see you did something along these lines in OperationParameterInjectionGenerator.EmitOperationContract

src/CoreWCF.BuildTools/src/Extensions.cs
mconnew: Nevermind, I see this is for generating nice looking parameter names in the generated code.

mconnew:
Please fix the extensions method which presumes the property name will be the first named argument, and add a compilation error if the property name passed is null. We probably want to protect against the empty string too, as they could specify PropertyName = "".

g7ed6e:
> Please fix the extensions method which presumes the property name will be the first named argument, and add a compilation error if the property name passed is null. We probably want to protect against the empty string too, as they could specify PropertyName = "".

@mconnew Passing null or empty string to PropertyName now results to a COREWCF_0103 build error.

mconnew (approved):

PR #1287, opened 2023-11-09
Detect kafka client side timeout

### LINKED ISSUE #1192: [Investigation] CoreWCF.Kafka.Client - Detect a timeout scenario
> Is there any way to detect a timeout scenario. This would happen if the CancellationToken is cancelled. In that scenario, we would normally throw TimeoutException. It's the only exception we throw regularly which isn't derived from CommunicationException

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/943#discussion_r1306262983_

g7ed6e:
Resolves #1192

### DIFF
--- a/src/CoreWCF.Kafka/tests/Helpers/DockerEx.cs
+++ b/src/CoreWCF.Kafka/tests/Helpers/DockerEx.cs
@@ -1,0 +1,68 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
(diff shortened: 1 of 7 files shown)


src/CoreWCF.Kafka/tests/Helpers/DockerEx.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace CoreWCF.Kafka.Tests.Helpers;
+
+public static class DockerEx
+{
+    public static async Task PauseAsync(ITestOutputHelper output, string containerName)
+    {
```
mconnew, line 15: Would it be worth using [Docker.DotNet](https://www.nuget.org/packages/Docker.DotNet/)?

mconnew (approved):

PR #1290, opened 2023-11-12
Add multiple Kafka topic subscription using regex support

g7ed6e:


src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageContext.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using CoreWCF.Queue.Common;
+
+namespace CoreWCF.Channels;
+
+internal class KafkaMessageContext : QueueMessageContext
+{
+    protected override void OnRequestMessageSet(Message message)
```
g7ed6e, line 10: @mconnew Here we override the `To` header with the regex supplied in the binding address. Another option that would not override the `To` header at transport level would be to implement something similar to CoreWCF.WebHttp by adding a new extension method around `IServiceCollection` to register dedicated `IServiceBehavior`, `IEndpointBehavior` and `IOperationBehavior` that pull a custom attribute from the service implementation and ease routing the message to the correct operation while dispatching.
Please let me know what you think is best

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
```diff
@@ -127,7 +127,7 @@ private async Task SendAsync(Message message, TimeSpan timeout)
                 kafkaMessage.Value = new Span<byte>(messageBuffer.Array, messageBuffer.Offset, messageBuffer.Count).ToArray();
                 await _producer.ProduceAsync(_topic, kafkaMessage, cts.Token);
             }
-            catch (ProduceException<Null, byte[]> produceException)
+            catch (ProduceException<byte[], byte[]> produceException)
             {
```
mconnew, line 131: Is it no longer possible for `ProduceException<Null, byte[]>` to get thrown? Is there a non-generic base type to ProduceException that might better fit here?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageContext.cs
mconnew: Why is this needed? `QueueFetchMessage.InvokeAsync` has this code:
```c#
                var message = await encoder.ReadMessageAsync(memStream, (int)maxReceivedMessageSize);
                if (message.Headers.To == null)
                {
                    message.Headers.To = queueMessageContext.LocalAddress.Uri;
                }
```
It looks like the incoming message is missing the To header anyway, which is normal when the transport inherently has a To address in order to get the message. When using Regex topics, does the message get a To address header added?

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
mconnew: I see the base type for `ProduceException<TKey, TValue>` is `KafkaException`. It looks like the only property we use from `ProduceException<TKey, TValue>` is `Error`, which lives on the base exception `KafkaException`. Would it make sense to switch the catch and the helper method to using KafkaException so we don't care what the generic shape is?

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageContext.cs
g7ed6e: @mconnew The `To` header is added by the CoreWCF Kafka client side package no matter if regex topics is used or not please see here https://github.com/CoreWCF/CoreWCF/blob/f770965d85af08925e93769ec4731591fd0204db/src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs#L110.
When using regex topics, only the service side is using regex, client side only "knows" one topic, the one to which the message is produced to. In that case the To header is set to a specific topic name but the service side needs a regex to correctly route the message.
The OnRequestMessageSet is there to ovewrite the To header from the message with the localaddress including the regex so the message gets routed.

src/CoreWCF.Kafka.Client/src/CoreWCF/ServiceModel/Channels/KafkaOutputChannel.cs
g7ed6e: KafkaException is now caught

src/CoreWCF.Kafka/src/CoreWCF/Channels/KafkaMessageContext.cs
mconnew: There are a few extensibility points that could enable a really nice experience here. You can create your own attribute which replaces OperationContractAttribute altogether. You have your attribute implement IOperationContractAttributeProvider so you don't need to double up your attributes and can keep the interface design clean (and removes the need to sepcify `IsOneWay=true` as the `OperationContractAttribute` you provide would have that set). You then implement `IOperationBehavior` which makes sure a custom `IDispatchOperationSelector` is set on the parent `DispatchRuntime` which can understand your attribute and use a regex to route the request to the correct operation. Is this the kind of thing you were thinking of? Where the developer experience would be something like this:
```c#
public interface IMyService
{
    [KafkaOperation(Regex = "...")]
    public void DoSomething(string customerId, string orderId);
}
```

mconnew (approved):
:shipit:

PR #1328, opened 2024-01-28
Add OpenApi document defaults so more methods are documented and prefix address

yowl:
This PR adds default for the `WebInvoke` - "post", and the `Method` properties for `WebInvoke` and `WebGet` so that more methods can be documented.  Also prefixes the registered address to the path.

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
@@ -140,6 +140,7 @@ namespace CoreWCF.OpenApi
                         tagsToHide,
                         basePathAttribute?.BasePath,
                         contractInfo.ResponseFormat,
+                        contractInfo.Address,
                         GetMethodUriWebGet);
 
                     PopulateOpenApiPath(
@@ -148,6 +149,7 @@ namespace CoreWCF.OpenApi
                         tagsToHide,
                         basePathAttribute?.BasePath,
                         contractInfo.ResponseFormat,
+                        contractInfo.Address,
                         GetMethodUriWebInvoke);
                 }
             }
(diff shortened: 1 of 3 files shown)


yowl:
thread pool starvation ?

yowl:
Test's all failed, but I can't see why, "thread pool starvation." ?

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
```diff
@@ -193,6 +196,7 @@ private static void PopulateOpenApiPathsOperations(OpenApiDocument document, IEn
 
             string uri = Regex.Replace(operationInfo.UriTemplate, @"\?.*", "");
 
+            uri = address.OriginalString + "/" + uri;
             if (!string.IsNullOrEmpty(additionalBasePath))
```
mconnew, line 200: The tests are failing because there are code paths to get here which don't require calling ServiceModelWebServiceBuilderExtensions.AddServiceWebEndpoint, which means OpenApiContractInfo.Address might not be populated. That results in address being null, so a NRE thrown when trying to get address.OriginalString. You need to guard against this possibility by null checking address and only executing this line of code if it's not null. I just did a local validation of this and all the tests were now passing.

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
```diff
@@ -193,6 +196,7 @@ private static void PopulateOpenApiPathsOperations(OpenApiDocument document, IEn
 
             string uri = Regex.Replace(operationInfo.UriTemplate, @"\?.*", "");
 
+            uri = address.OriginalString + "/" + uri;
```
mconnew, line 199:
```suggestion
            if (address != null)
            {
                uri = address.OriginalString + "/" + uri;
            }
```

src/CoreWCF.WebHttp/src/CoreWCF/OpenApi/OpenApiSchemaBuilder.cs
yowl: Thanks, sorry for the late reply, and thanks for the commit.  Not in front of the computer as on a break.

mconnew:
I'd like to see a test which validates that this behaves the way it's expected.

g7ed6e:
@yowl do you think you can add a unit test ?

yowl:
Sorry for the late reply.  I will look at adding the unit test.

Issue #1331, opened 2024-02-01
[Bug]: NullReferenceException thrown when using invalid URL


bjornbouetsmith:
### Duplicate ?

- [X] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.5.1

### Describe expected behavior

I have a net.tcp WCF Service bound to the following URL:
net.tcp://localhost:1234/service/bootstrap2/0

If I call this service with a wcf client, but uses endpoint address:

net.tcp://localhost:1234/service/bootstrap2/1

I get a NullReferenceException logged in my log - instead of a more "correct" 404 not found.

The reason for client calling the "wrong" url - is because our client can also connect to a service running on netframework, where portsharing is enabled, so services on the same server share the port, but only differentiate on the last part of the url, i.e. one service binds to /0 the other binds to /1.

In net framework WCF, this never causes errors to be logged server side, and client just tries next number - with corewcf when a client tries a "wrong" url, the null reference exceptions is logged

### Describe actual behavior

Null reference exception thrown instead of a more correct exception that should be captured by the framework and turned into a communication exception

### Which binding

NetTcp

### security

None

### Which .NET version

.NET 6

### Which os platform

Windows

### Code snippet used to reproduce the issue

```c#
No real code example - but create a net.tcp service, bind to specific url - call service with a WCF client on correct port, but slightly different url.
```

### Stacktrace if any

```shell
00:18:36.358|3788037464|ERROR|119|Kestrel|Unhandled exception while processing 0HN132C763JHH.|System.NullReferenceException
Object reference not set to an instance of an object.
   at CoreWCF.Channels.Framing.FramingModeHandshakeMiddleware.OnConnectedCoreAsync(FramingConnection connection)
   at CoreWCF.Channels.Framing.FramingModeHandshakeMiddleware.OnConnectedAsync(FramingConnection connection)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.KestrelConnection`1.ExecuteAsync()
```

bjornbouetsmith:
I have pulled the code and it seems like adding:

```csharp
 if (connection.ServiceDispatcher == null)
{
    await connection.SendFaultAsync(FramingEncodingString.EndpointNotFoundFault, Timeout.InfiniteTimeSpan/*GetRemainingTimeout()*/,
[...]

PR #1335, opened 2024-02-07
Prevent null reference exception when a client makes request to wrong endpoint address (#1331)

bjornbouetsmith:
Simple code fix with test that asserts that a NullReferenceException is no longer thrown when a client makes a request to a wrong endpoint


bjornbouetsmith:
Hmm, seems like tests are quite unstable - and the tests that fails should really not be affected by my change.

bjornbouetsmith:
@mconnew Do you have time to take a look at this PR?
Or will it happen automatically at some predestined time?

mconnew:
Looks good, thank you for this

bjornbouetsmith:
Awesome, thanks - will this be part of the 1.6 version or do you plan on releasing a 1.5.2 version?

PR #1357, opened 2024-03-27
Fix SymmetricKey class access modifier

tarikhasimoglu:
Summary of the changes:
- <s> Renamed the CoreWCF.Primitives/CoreWCF.IdentityModel.Tokens/**SymmetricKey.cs** file to match the class name **InMemorySymmetricSecurityKey**. </s> Reverted as I found out that the file name is also SymmetricKey.cs in MS source code.
- Changed the access modifier to Public as in the original class in .Net Framework. 
 
Ref:
https://referencesource.microsoft.com/#System.IdentityModel/System/IdentityModel/Tokens/SymmetricKey.cs,508a851b9eb69307,references


tarikhasimoglu:
@dotnet-policy-service agree

g7ed6e:
Thanks @tarikhasimoglu for the PR. Do you think you can add a unit test for this?

tarikhasimoglu:
Hi @g7ed6e , I added a unit test to ensure that the class is public. Please let me know if it needs any modifications.

g7ed6e:
Thanks @tarikhasimoglu . The general guidance is that WCF features ported to CoreWCF are ported as internal types. Then, to make types public, unit tests should cover the feature. In that case an integration test where a client call a server with a binding configured to use this type made public would be sufficient.

g7ed6e:
@tarikhasimoglu Having a second look to this i checked the dotnet/wcf repo and found this type declared internal https://github.com/search?q=repo%3Adotnet%2Fwcf%20inmemorysymmetricsecuritykey&type=code . Can you share why you need this type public ?

mconnew:
@tarikhasimoglu, we're nearly ready to release the next version. I'd like to resolve this PR one way or the other before the release. Can you explain what your usage scenario is where you want this class made public?

tarikhasimoglu:
@mconnew @g7ed6e I'm sorry for the late reply. The intended scenario is to create a custom class representing a security token that contains multiple security keys generated using the symmetric algorithms. I have updated the unit test to ensure the keys are created using the InMemorySymmetricSecurityKey class.

g7ed6e:
> @mconnew @g7ed6e I'm sorry for the late reply. The intended scenario is to create a custom class representing a security token that contains multiple security keys generated using the symmetric algorithms. I have updated the unit test to ensure the keys are created using the InMemorySymmetricSecurityKey class.

Do you need a key rotation feature? Can you share how you are using this server side ? (Please only share the CoreWCF related code)

mconnew:
I'm not happy with exposing the inbox implementation as it currently is because it's intended usage was to solve an internal problem and it's not suitable for public consumption. Internal usage we know we're using it correctly so don't need to worry about as much validation. But a public class should do things like validate the crypto algorithm's supported key length with the provided key when encrypting/decrypting data.  
As @g7ed6e said, I'd like to see how you're using this? Normally security key's are instantiated by a KeyIdentifierClause which comes from a key material carrying SecurityToken that's been received such as a Saml token. We don't have an STS implementation like WCF does, so I'm struggling to think of what the scenario would be where you need to instantiate this yourself for CoreWCF server (client is a different situation, but still a stretch to instantite yourself).  
My current inclination is to say the code is open source, copy it into your own app. I'd want to put more time and thought into what a public implementation type would look like. E.g. do we need to be thinking about key vault scenarios? How does that play with key rotation? Do we need to support having multiple keys and either have a way to know which is needed to be used when decrypting, or should we be trying all of them starting from the newest (should generally only be 2 unless a server has been up a LONG time).

tarikhasimoglu:
I've closed this PR after reviewing the original WCF implementation, as the InMemorySymmetricSecurityKey class is internal in the original WCF source. Making it public in CoreWCF could introduce unintended issues. For my specific use case, I'll implement a custom solution within my project. Thanks for the feedback and discussion.

PR #1362, opened 2024-04-04
Fix/endpoints substring bug

### LINKED ISSUE #1254: [Bug]: Incorrect mapping of dispatchers to handlers
### Duplicate ?

- [X] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.0.0+

I found an [bug](https://github.com/CoreWCF/CoreWCF/blob/main/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs#L166), that causes us to execute incorrect handlers for dispatchers.
If we have endpoints, for example:
`/myservice`
and
`/mysevice/mtom`
We will get the first initialised dispatcher.

For example, we use:
```
builder.AddService<Serv>();
builder.AddServiceEndpoint<Serv>(.., "/myService")
builder.AddServiceEndpoint<Serv>(.., "/myService/mtom")
```
and We will request the endpoint `/myService/mtom` we will have  handler for `/myService`

### Describe actual behavior

For fix please replace:
```
context.Request.Path.StartsWithSegments(dispatcher.BaseAddress.AbsolutePath, out _, out _))
```
to (or similar)
```
context.Request.Path.HasValue &&
context.Request.Path.Value.Equals(dispatcher.BaseAddress.AbsolutePath, StringComparison.OrdinalIgnoreCase)
```

### Which binding

NetTcp

### security

None

### Which .NET version

.NET 6

### Which os platform

Linux

_No response_

### Stacktrace if any

_No response_

MeikTranel:
Attempts to resolve #1254

I still have some questions regarding the underlying problem - will formulate them tomorrow. Draft to get some CI validation rolling.

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
@@ -163,8 +163,7 @@ namespace CoreWCF.Channels
                     bool ExecHandler(HttpContext context)
                     {
                         return (dispatcher.BaseAddress.AbsolutePath == "/"
-                               || context.Request.Path.StartsWithSegments(dispatcher.BaseAddress.AbsolutePath, out _, out _)) &&
-                               dispatcher.Binding.Scheme == context.Request.Scheme;
+                               || context.Request.Path is PathString reqPath && reqPath.Equals(dispatcher.BaseAddress.AbsolutePath) && dispatcher.Binding.Scheme == context.Request.Scheme);
                     }
 
                     branchApp.MapWhen(ExecHandler, wcfApp =>
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/ServiceModelHttpMiddleware.cs
@@ -162,8 +162,8 @@ namespace CoreWCF.Channels
 
                     bool ExecHandler(HttpContext context)
                     {
-                        return (dispatcher.BaseAddress.AbsolutePath == "/"
-                               || context.Request.Path is PathString reqPath && reqPath.Equals(dispatcher.BaseAddress.AbsolutePath) && dispatcher.Binding.Scheme == context.Request.Scheme);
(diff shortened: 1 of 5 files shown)


mconnew:
I would expect this to break WebHttp. The fact it hasn't is likely because we don't have tests covering those scenarios. With WebHttp, you add the service endpoint "/MyApp/MyService.svc" and an operation can have an attribute that looks like `[WebGet(UriTemplate = "customers/{id}")]`. In this case, a valid request could have the path "/MyApp/MyService.svc/customers/123". Your change will result in this failing. The solution is to use `UriPrefixTable`. Basically you register your base paths for some type of object at setup. Then when it comes to dispatch, you pass it the entire Uri and it finds the closest match being as greedy as possible. Basically it goes segment by segment down a tree like graph until it finds an end node. It then checks to see if there's any further child nodes (another path with the same start but more segments) and keeps checking. If there's anything which matches the next segment, it keeps searching deeper, otherwise it returns the found node. It's what's used by NetTcp. `UriPrefixTable` is generic, and in this case the generic type you would use would probably be RequestDelegate, which is what you get as a result of calling `IApplicationBuilder.Build()`.

MeikTranel:
I thought about this - will try to get some more tests in to cover those cases. I still feely wholly iffy about the entire predicate being a reimplementation of the  AddressFilter thingy - it feels like i'm implementing the address matching code as a second level of truth right now. I also am super irritated by the whole empty slash short circuit part of the second commit.

mconnew:
> The point where this all goes south is when we act like all dispatchers are made equal specifically regarding BaseAddress - we lose one measly but critical information during the abstraction - whether the BaseAddress is meant to be an actual base address or an exact match spec. This is information that can be inferred from the specific types but i'm struggling to decide where to store this information without touching to much stuff.

This can't be known due to the extensibility of CoreWCF as demonstrated by WebHttp. The key concept built into UriPrefixTable is that it returns the greediest match. So if you register /foo/bar and /foo, and a request comes in for /foo/bar/fuzz, it will match to /foo/bar no matter the order they are registered with CoreWCF. You can work around this issue today to some extent by calling AddServiceEndpoint in the order of most specific to least specific. If you call AddServiceEndpoint for /foo first, then /foo/bar, then /foo will hide /foo/bar and will get dispatched calls to /foo/bar. If you call passing in /foo/bar first, then call passing /foo, then things work as you would expect.

PR #1372, opened 2024-04-19
Use Docker.Dotnet in CoreWCF.Kafka.Tests

### LINKED ISSUE #1347: Use Docker.DotNet in Kafka UT
Would it be worth using [Docker.DotNet](https://www.nuget.org/packages/Docker.DotNet/)?

_Originally posted by @mconnew in https://github.com/CoreWCF/CoreWCF/pull/1287#discussion_r1513542345_

g7ed6e:
Resolves #1347

### DIFF
--- a/src/CoreWCF.Kafka/tests/CoreWCF.Kafka.Tests.csproj
+++ b/src/CoreWCF.Kafka/tests/CoreWCF.Kafka.Tests.csproj
@@ -32,6 +32,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Confluent.Kafka" Version="2.3.0" />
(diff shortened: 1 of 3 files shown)


mconnew (approved):

mconnew:
As a further step it might be worth investigating integrating [TestContainers](https://dotnet.testcontainers.org/). Looking at their list of [modules](https://dotnet.testcontainers.org/modules/),they have nuget packages for [Kafka](https://www.nuget.org/packages/Testcontainers.Kafka) and [RabbitMq](https://www.nuget.org/packages/Testcontainers.RabbitMq). That's probably a project in it's own right, so I think we should merge this as it's an improvement over manually executing shell commands.

Issue #1376, opened 2024-04-30
[Bug]: CoreWCF.NetTcp.Tests.TransportFrameworkTests.RemoteEndpointMessageProperty test can fail due to concurrent tests running


mconnew:
### Duplicate ?

- [X] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.5.2

### Describe expected behavior

CoreWCF.NetTcp.Tests.TransportFrameworkTests.RemoteEndpointMessageProperty should pass

### Describe actual behavior

Sometimes it fails stating the port is incorrect. This can be caused if there's another test running concurrently and the call to netstat to get the list of opened ports picks up the port for a different test that's running concurrently. The failure has only been observed on NetFx so far, but there's no reason why it couldn't be observed on any OS or runtime.  
It's possible this could be solved by using an EventListener to get the WCF ETW events. I think there's an event that's emitted telling you the port number of the local end of the socket. If there is, then we could instead check the returned port number is within the set of local ports which were opened during the channel.Open call. If I'm wrong about there being an ETW event that can be used for this (there might be a networking ETW event we can use instead, at least on .NET), then we might have to skip this test.

### Which binding

NetTcp

### security

None

### Which .NET version

.NET Framework

### Which os platform

Windows

PR #1379, opened 2024-05-02
Fix Templates pipeline

g7ed6e:


mconnew (approved):

g7ed6e:
@mconnew this PR removes the azdo pipeline dedicated to templates too. The stage that run templates tests is ran conncurrently to the product unit tests.

mconnew:
The test that failed was due to #1376. I suspect we're going to see more of this in future.

mconnew (approved):

Issue #858, opened 2022-09-29
WSDL and "landing page" are only exposed on first endpoint; returns HTTP 400 for additional ones


cyanite:
If you add a service with two endpoints, say like this

    config.AddService<Foo>();
    config.AddServiceEndpoint<Foo, IFoo>(new BasicHttpBinding(), "http://localhost:8765/foo");
    config.AddServiceEndpoint<Foo, IFoo>(new BasicHttpBinding(), "http://localhost:8765/foo2");
    ...
    var serviceMetadataBehavior = app.ApplicationServices.GetRequiredService<CoreWCF.Description.ServiceMetadataBehavior>();
    serviceMetadataBehavior.HttpGetEnabled = true;

Then WSDL metadata and landing page output only works for the first endpoint. So these work as expected:

http://localhost:8765/foo (landing page)
http://localhost:8765/foo?singleWsdl (WSDL)

And these return 400 Bad Request:

http://localhost:8765/foo2
http://localhost:8765/foo2?singleWsdl

Now, the metadata from the first endpoint does contain both endpoints in the "port" list, but shouldn't this work for both of them?

ondrejtucny:
Experiencing the same problem. IMO each endpoint should expose its own metadata. In my case, one endpoint is SOAP 1.2 and the other one SOAP 1.1. Hence, this shloud not be mixed in the same WSDL generated for both endpoints, but one WSDL per each endpoint.

Chaosg5:
I've got a fairly similar problem, but with a bit of difference.
I have an application with 7 separate contracts setup as 7 different services each with their own enpoint.
In localhost each of them can show their own WSDL after enabling ServiceMetadataBehavior.HttpsGetEnabled.
But when publishing (to an IIS host) only the WSDL for the first service can will be displayed and at the root page, none of the sub-addresses can show their metadata and returns 400 Bad Request, but do work correctly when calling them through a client.

PR #1409, opened 2024-07-12
Fix WSDL endpoint URLs for multiple services

a-tze:
Fix WSDL endpoint URLs for multiple services
 - when multiple WCF services are added to a single host, only the WSDL of the first service could be retrieved from the base url
 - GETting the WSDL from any other URL than the base url returned http status 400 bad request
 - successfully tested locally with v.1.3.2 and v1.5.2
 - current behaviour still applies, e.g. GETting WSDL from the base url still returns metadata of first WCF service

#858

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
@@ -136,8 +136,8 @@ namespace CoreWCF.Description
             mex.HttpGetEnabled = HttpGetEnabled;
             mex.HttpsGetEnabled = HttpsGetEnabled;
 
-            mex.HttpGetUrl = host.GetVia(Uri.UriSchemeHttp, _httpGetUrl ?? new Uri(string.Empty, UriKind.Relative));
-            mex.HttpsGetUrl = host.GetVia(Uri.UriSchemeHttps, _httpsGetUrl ?? new Uri(string.Empty, UriKind.Relative));
+            mex.HttpGetUrl = host.GetVia(Uri.UriSchemeHttp, GetFirstEndpointUriForScheme(Uri.UriSchemeHttp, _httpGetUrl, description));
+            mex.HttpsGetUrl = host.GetVia(Uri.UriSchemeHttps, GetFirstEndpointUriForScheme(Uri.UriSchemeHttp, _httpsGetUrl, description));
 
             foreach (ChannelDispatcherBase dispatcherBase in host.ChannelDispatchers)
             {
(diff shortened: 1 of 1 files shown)


a-tze:
@dotnet-policy-service agree

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
```diff
@@ -162,6 +162,14 @@ private void SetExtensionProperties(ServiceDescription description, ServiceHostB
             }
         }
 
+        private static Uri GetFirstEndpointUriForScheme(string uriScheme, Uri configuredGetUrl, ServiceDescription description)
+            => configuredGetUrl
+                ?? description?.Endpoints?.FirstOrDefault(endpoint =>
+                       endpoint?.Address?.Uri != null
+                       && endpoint.Address.Uri.Scheme == uriScheme)
```
mconnew, line 169: I think you can simplify this:
```suggestion
                       endpoint?.Address?.Uri?.Scheme == uriScheme)
```

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
```diff
@@ -162,6 +162,14 @@ private void SetExtensionProperties(ServiceDescription description, ServiceHostB
...
+                   ?.Address.Uri
```
mconnew, line 170: You are accepting that there might not be a matching endpoint and FirstOfDefault could return the default (which is null) by using `?.Address`, but if it's null, that will also evaluate to null, so `.Uri` will throw NullReferenceException.
```suggestion
                   ?.Address?.Uri
```

src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataBehavior.cs
a-tze: In the past I had different experiences with something that ended up as

`(string)null == string `

which is then actually executed as 

`(string)null.Equals(string)`

wich then yielded a NullReferenceException. But I'm not sure wich language level is targeted in this project currenty, this behaviour seemed to have changed in the last years between the different framework versions.

My preference is

`string.Equals(endpoint?.Address?.Uri?.Scheme, uriScheme) `

but that is only 100% correct when uriScheme is never given as null, therefore I wrote the separate explicit null-check.

mconnew: I usually like string.Equals too as you can specify the string comparison mode and skip some null checks. FYI `(string)null == string` will use the string == operator method, which calls into the static string.Equals method. One advantage of using string.Equals over == is you can't accidentally call the object == operation. If you have an object you expect to be a string and forget to cast to string first, == could do something different than you intend. Feel free to switch to string.Equals if you prefer, but use the overload which takes a StringComparison and pass in StringComparison.Ordinal.

mconnew:
I will merge this after a couple of other fixes are merged so if you want to make a change to use string.Equals, you have a couple of days.

PR #1418, opened 2024-07-23
Drop net7.0 from templates and UT

g7ed6e:


g7ed6e:
@mconnew as net7.0 is EOL it's time to no longer propose it as TFM.

mconnew (approved):

PR #1424, opened 2024-08-01
Remove NegotiateInputLengthDecider from CoreWCF.NetFramingBase

### LINKED ISSUE #1013: NetTcp Negotiate framing code is wrong and leads to crash
The frame size code in DuplexPipeStream's NegotiateInputLengthDecider uses big endian to parse the frame size, which is incorrect according to the standard (https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-NNS/%5bMS-NNS%5d.pdf) and reality, and leads to a crash when large amounts of data are transferred.

DuplexPipeStream sits below the system's own NegotiateStream (when using negotiate auth), and tries to follow along with the frames (for reasons that don't make sense to me; see below). The first data frame that is read typically has a size of 17 bytes, which is interpreted as 285212672 due to the endian bug. This means that for the next 285212672 bytes (if using the same channel), everything is fine, and the framing code in DuplexPipeStream is effectively completely disabled. After this, four more, essentially random, bytes are read, often parsing to a negative number, which leads to a full crash (debug mode) or connection termination (release mode).

This is, of course, easy to fix, by doing the conversion little-endian instead. But the question is why bother? Why does DuplexPipeStream even peek at the frame size? It's used for just one thing: To truncate reads from the containing stream early, in case they read across a frame boundary. But why do this? Because,

1. NegotiateStream (the containing stream) never reads across a frame boundary, because it obviously has to parse the frame counts and strip them away from the data.
2. Even if it did, why bother trying to adapt to an upstream protocol? It can handle its own data just fine. It evidently works just fine without this feature, as it is, as noted, effectively disabled in practice.

The mechanism is clearly not important since, due to the bug, it's effectively disabled anyway, and communication works just great... for the first 280MB on a given channel.

Framing code is also used for TLS authentication streams, which I didn't analyze; it may have a necessary function there (although I am skeptical of that; it seems to do exactly the same: truncate reads to the frame size, which, presumably, is already done by the containing stream; it also seems brittle, as the containing protocol stream is complex).

I propose that the framing code for negotiate be removed entirely, as it serves no purpose. I'll help with pull requests when we've made a decision on this. Note the framing code has been wrong since it was introduced.

This is the same issue that causes the bug reported in #846

SaxoAnders:
This should fix #1013 

It looks like it already been fixed in the other DuplexPipeStream files, but for some reason the version in CoreWCF.NetFramingBase was not changed, which leads to random bytes being interpreted as _frameSize in our project, which ultimately leads to an exception when the random bytes are interpreted as a negative number.

### DIFF
--- a/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/Framing/DuplexPipeStream.cs
+++ b/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/Framing/DuplexPipeStream.cs
@@ -167,8 +167,6 @@ namespace CoreWCF.Channels.Framing
         {
             if (contentType == FramingUpgradeString.SslOrTls)
                 _inputLengthDecider = new TlsInputLengthDecider();
-            else if (contentType == FramingUpgradeString.Negotiate)
-                _inputLengthDecider = new NegotiateInputLengthDecider();
             else _inputLengthDecider = NoopInputLengthDecider.Instance;
         }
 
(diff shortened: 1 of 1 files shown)


SaxoAnders:
@dotnet-policy-service agree

mconnew:
I'm curious why it's failing. NegotiateStream uses a basic length prefix framing so this should never fail. The reason this was added is because of a combination of SslStream being greedy when reading from the wrapped stream and PipeReader coalescing multiple reads from the wrapped socket. We need it to not be greedy so we can close a channel and reuse the socket. Being greedy would sometimes read the bytes from the next client channel that's connecting. I implemented the same protection mechanism for NegotiateStream in case the runtime team decided to make that "more efficient" and be greedy too. I'm okay with removing this, but if NegotateStream ever becomes greedy, it will need to come back and then work out what's going wrong.

mconnew:
Just read the issue you linked to, looks like it's a BigEndian vs LittleEndian bug, so we know the fix if it needs to come back.

PR #1433, opened 2024-08-13
Add missing UDS ProtectionLevel property

mconnew:
There was also a problem where the unit tests had an extra catch block which would hide test failures.


g7ed6e (approved):
Updated NetTcp test does not relate to the PR title but I guess it is the flaky test we chase since a few commits.

mconnew:
Fixes ##1414

PR #1437, opened 2024-08-19
Update package versions in preparation for release, update docs on how to do so

mconnew:
We were missing updating the Microsoft.IdentityModel packages because we minor version lock Microsoft.* when updating. Updated release process documentation to include this.
Added more logging as well as hopefully fixed the RemoteEndpointMessageProperty test. I believe it was GC cleaning up the channel before netstat was able to run that caused the failures.

### DIFF
--- a/src/CoreWCF.NetTcp/tests/TransportFrameworkTests.cs
+++ b/src/CoreWCF.NetTcp/tests/TransportFrameworkTests.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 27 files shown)


src/CoreWCF.NetTcp/tests/TransportFrameworkTests.cs
```diff
@@ -132,52 +142,64 @@ private void CheckIP(string ip)
...
         {
             string line;
             string originPort;
             string destinationPort = null;
             string rawLine = null;
             string matchingPortRawLine = null;
             bool verifiedClientPort = false;
-
+            bool extraLogging = false;
+#if NETFRAMEWORK
```
g7ed6e, line 154: something like this would be more concise 
bool extraLogging = Environment.Version.Major == 4 && RuntimeInformation.IsOSPlatform(OSPlatform.Windows)

mconnew: I know, this is quick and dirty. I plan to remove it at some point once I have confidence the issue is fixed.

g7ed6e (approved):

PR #1449, opened 2024-09-03
feat(logo): CoreWCF logo artifacts and psd

### LINKED ISSUE #1092: CoreWCF needs an official logo
We'd like to invite submissions for a logo for CoreWCF. It will be used as the logo for the nuget packages, as well as being featured on web pages such as the blog website, or as a twitter account profile picture. Each unique submission should be in its own comment so that people can vote on their favorite.  

The CoreWCF maintainers will pick one which has a large number of votes, but reserve the right to not choose the highest voted (we don't want a [Mister Splashy Pants](https://en.wikipedia.org/wiki/Mister_Splashy_Pants) situation),  

The creator of the winning image will be expected to submit a pull request with the image to ensure the copyright is transferred to the .NET Foundation the same way any submitted code is. We will credit the author in a way that's mutually agreeable to them and the CoreWCF maintainers.  

The ideal format to commit to the repo would be a rasterized format such as SVG, although a sufficiently high resolution bitmap such as PNG is likely to be accepted as long as it can scale well.  

Avoid any inspiration from existing company logos. For example, including a window because WCF had it's origin in Microsoft Windows isn't acceptable.  

### Inspiration for the design
These are just suggestions, nothing is required to be a part of the logo.  
The original project codename for WCF in .NET Framework was Indigo. A lighter shade of Indigo is Capri. I used the [Bing AI image creator](https://bing.com/create) to generate a logo using these colors to see what kinds of things it could come up with and am submitting them as possibilities to consider. I used combinations of the following words and phrases as input to Bing.
* The colors Indigo and Capri
* Geometric shapes
* Idea of flow
* Connectedness
* Logo

I've submitted some of my favorite generated logos. Feel free to go for a completely different theme, the logo doesn't need to be an abstract shape. For example, an original animal, or anthropomorphic shape (e.g. the .dotnet bot) would be reasonable submissions.

ChrisMcKee:
Resolves #1092

Let me know if you need any more sizes; formats 👍


ChrisMcKee:
@dotnet-policy-service agree

ChrisMcKee:
Added in svgs and the data uris in a readme file.
The badge/logo, text; combined svgs have been converted to glyphs so the text isn't reliant on a font being present.
The svgs are based on the same base files but with the gradients flattened as svg gradients are a mixed bag.

mconnew:
I was going to say we need a 128x128 png as the [nuget icon guidelines](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file) recommend this, but then I went looking at the .NET nuget package logo and that is the weird size of 456x456 so we're good with the 256 file :smile:.

ChrisMcKee:
Ha! I went with the slack avatar sizes. Most are happy as long as it's under 1024x and square

PR #1454, opened 2024-09-12
Use Assert.Fail instead of true/false to fail the test

Havunen:
Small clean up to test case asserts


Havunen:
@dotnet-policy-service agree

g7ed6e (approved):
LGTM

PR #1462, opened 2024-10-07
Fix buffer not being returned to pool

SaxoAnders:
We discovered that byte arrays were not being returned to the pool that they were taking from when sending a message through ServerFramingDuplexSessionChannel, which caused a lot of allocations.

I checked how it worked in old WCF, there the buffer was returned by the Connection:
image

But since the code shape has changed a bit, I decided to do it in OnSendCoreAsync instead. This is also where it is done in WebSocketTransportDuplexSessionChannel.

### DIFF
--- a/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
+++ b/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
@@ -332,8 +332,15 @@ namespace CoreWCF.Channels
             allowOutputBatching = message.Properties.AllowOutputBatching;
             Connection.Logger.SendMessage(message);
             messageData = EncodeMessage(message);
-            await Connection.Output.WriteAsync(messageData, token);
-            await Connection.Output.FlushAsync();
+            try
+            {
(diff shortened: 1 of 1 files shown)


src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
```diff
@@ -332,8 +332,15 @@ protected override async Task OnSendCoreAsync(Message message, CancellationToken
...
-            await Connection.Output.WriteAsync(messageData, token);
-            await Connection.Output.FlushAsync();
+            try
+            {
+                await Connection.Output.WriteAsync(messageData, token);
+                await Connection.Output.FlushAsync();
+            }
+            finally
+            {
+                BufferManager.ReturnBuffer(messageData.Array);
```
mconnew, line 342: Looking at the WCF code, the async code semantically does it in the same place (code is all over the place with APM, but it's this class that does the return and not the connection.Write call). This is exactly how I would have fixed this.

mconnew (approved):

mconnew:
Thank you for root causing and submitting a fix. We will be making a new release for some changes needed for .NET 9 so there won't be much delay before this is released.

PR #1464, opened 2024-10-09
Default value for AutoDisposeParameters should be true

### LINKED ISSUE #932: IDisposable Reponse is not disposed
Hi,

in WCF responses which implement the IDisposable interfaces are automatically disposed. In CoreWCF that does not work. I have to manually implement in each operation following workaround:

```
OperationContext.Current.OperationCompleted += delegate
{
    response?.Dispose();
};
```

Is this a known bug? Could i move this workaround to a general behavior?

Thanks!

SaxoAnders:
In old WCF default value for OperationBehaviorAttribute.AutoDisposeParameters was true, but in CoreWCF it is false.

https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/OperationBehaviorAttribute.cs,b18c31b796a39833

This is a breaking change for anyone coming from .NET Framework, who relies on WCF to dispose their parameters/return value.

This PR fixes #932

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/OperationBehaviorAttribute.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/OperationBehaviorAttribute.cs
@@ -15,7 +15,7 @@ namespace CoreWCF
         private ImpersonationOption _impersonation = ImpersonationOption.NotAllowed;
         private ReleaseInstanceMode _releaseInstance = ReleaseInstanceMode.None;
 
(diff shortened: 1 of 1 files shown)


mconnew:
Ideally we should have tests for this. Looking at the implementation, auto dispose parameters is true without applying the attribute, so the only ones affected by this are people using this attribute. This means auto-disposing is already a well exercised code path so enabling it when this attribute is applied isn't risking exposing latent bugs in CoreWCF. I'll open an issue to create tests around this.

PR #1465, opened 2024-10-09
Issues/1193

### LINKED ISSUE #1193: [Bug]: UseNetTcp() only uses IPv4
### Duplicate ?

- [X] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.3.2 (also tested on 1.4.0-preview1)

```cs
builder.WebHost.UseNetTcp(9999);
```
Kestrel should listen on [::]:9999 (IPv6 address for Any, which includes any IPv4).

### Describe actual behavior

Kestrel only listens on 0.0.0.0:9999 (IPv4 only).

This means that when I try to connect with localhost, it takes a long time since it tries v6 first, and then falls back to v4. I thought that CoreWCF was just slow, but it was a name resolution/connect error.

Current `ServiceModelWebHostBuilderExtensions` code:

```cs
public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder, int port)
{
    return webHostBuilder.UseNetTcp(IPAddress.Any, port);
}
```

Desired `ServiceModelWebHostBuilderExtensions` code:

```cs
public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder, int port)
{
    return webHostBuilder.UseNetTcp(IPAddress.IPv6Any, port);
}
```

Alternative workaround:

```cs
builder.WebHost.UseNetTcp(IPAddress.IPv6Any, 9999);
```

### Which binding

NetTcp

### security

Transport

### Which .NET version

.NET 7

### Which os platform

Windows

```c#
var builder = WebApplication.CreateBuilder(args);

builder.WebHost.UseNetTcp(9999);
// Workaround
// builder.WebHost.UseNetTcp(IPAddress.IPv6Any, 9999);
```

### Stacktrace if any

_No response_

arontsang:
Fixes #1193

### DIFF
--- a/src/CoreWCF.ConfigurationManager/tests/ServiceHelper.cs
+++ b/src/CoreWCF.ConfigurationManager/tests/ServiceHelper.cs
@@ -29,7 +29,7 @@ namespace CoreWCF.ConfigurationManager.Tests
             });
 #endif // DEBUG
             if (ipAddress == IPAddress.Any)
-                result = result.UseNetTcp(port);
+                result = result.UseNetTcp(IPAddress.Any, port);
             else
                 result = result.UseNetTcp(ipAddress, port);
             return result
--- a/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
+++ b/src/CoreWCF.NetTcp/src/CoreWCF/Configuration/ServiceModelWebHostBuilderExtensions.cs
@@ -31,6 +31,8 @@ namespace CoreWCF.Configuration
 
         public static IWebHostBuilder UseNetTcp(this IWebHostBuilder webHostBuilder, int port)
         {
+            if (Socket.OSSupportsIPv6)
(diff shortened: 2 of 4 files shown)


mconnew:
This is unnecessary. We have a newer UseNetTcp api which closer models the UseKestrel api which takes a Uri (or string Uri). If you do something like this:
```c#
host.UseNetTcp(netTcpOptions =>
{
    netTcpOptions.Listen("net.tcp://localhost/");
});
```
Then it will listen on IPv6Any is the OS supports IPv6, otherwise it listens on IPAny. Basically it checks Uri.HostNameType, and if it's of type UriHostNameType.Dns, it listens on IPv6Any. Localhost is seen like a dns name. Basically Dns means it needs to use the OS hostname resolution, which is how localhost is handled.
Closing as your change causes existing apps to start listening on interfaces they weren't previously, which could be a security problem. That and lots of tests broke with your change. Hopefully the newer api is a better choice. There's an overload to provide options for setting any of the previous NetTcpBinding settings which come into play if you have more than one endpoint configured on the port.

PR #1495, opened 2024-11-24
Use async IO to write MTOM XOP includes in responses

MarcWils:
Possible fix for #1496

### DIFF
--- a/src/CoreWCF.Http/tests/Helpers/AsyncOnlyStream.cs
+++ b/src/CoreWCF.Http/tests/Helpers/AsyncOnlyStream.cs
@@ -3,41 +3,20 @@
 
 using System;
 using System.IO;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace CoreWCF.Http.Tests.Helpers
 {
-    internal class AsyncOnlyStream : Stream
+    internal class AsyncOnlyStream : DelegatingStream
     {
-        private readonly Stream _inner;
-
         public AsyncOnlyStream(Stream inner)
+            : base(inner)
         {
-            _inner = inner;
         }
 
-        public override bool CanRead => _inner.CanRead;
-
-        public override bool CanSeek => _inner.CanSeek;
-
-        public override bool CanWrite => _inner.CanWrite;
-
-        public override long Length => _inner.Length;
-
-        public override long Position { get => _inner.Position; set => _inner.Position = value; }
-
         public override void Flush() => throw new NotSupportedException("Synchronous IO is not allowed");
-        public override Task FlushAsync(CancellationToken cancellationToken) => _inner.FlushAsync(cancellationToken);
 
         public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException("Synchronous IO is not allowed");
 
-        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => _inner.ReadAsync(buffer, offset, count, cancellationToken);
-
-        public override long Seek(long offset, SeekOrigin origin) => _inner.Seek(offset, origin);
-        public override void SetLength(long value) => _inner.SetLength(value);
         public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException("Synchronous IO is not allowed");
-
-        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => _inner.WriteAsync(buffer, offset, count, cancellationToken);
     }
 }
--- a/src/CoreWCF.Http/tests/Helpers/AsyncOnlyStream.cs
+++ b/src/CoreWCF.Http/tests/Helpers/AsyncOnlyStream.cs
@@ -1,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CoreWCF.Http.Tests.Helpers
+{
+    internal class AsyncOnlyStream : Stream
+    {
+        private readonly Stream _inner;
+
+        public AsyncOnlyStream(Stream inner)
+        {
+            _inner = inner;
+        }
+
+        public override bool CanRead => _inner.CanRead;
+
+        public override bool CanSeek => _inner.CanSeek;
+
+        public override bool CanWrite => _inner.CanWrite;
+
+        public override long Length => _inner.Length;
+
+        public override long Position { get => _inner.Position; set => _inner.Position = value; }
+
+        public override void Flush() => throw new NotSupportedException("Synchronous IO is not allowed");
+        public override Task FlushAsync(CancellationToken cancellationToken) => _inner.FlushAsync(cancellationToken);
+
+        public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException("Synchronous IO is not allowed");
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => _inner.ReadAsync(buffer, offset, count, cancellationToken);
+
+        public override long Seek(long offset, SeekOrigin origin) => _inner.Seek(offset, origin);
+        public override void SetLength(long value) => _inner.SetLength(value);
+        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException("Synchronous IO is not allowed");
+
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => _inner.WriteAsync(buffer, offset, count, cancellationToken);
+    }
+}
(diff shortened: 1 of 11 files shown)


src/CoreWCF.Primitives/src/CoreWCF/Channels/Message.cs
```diff
@@ -830,6 +830,15 @@ internal void WriteMessagePostamble(XmlDictionaryWriter writer)
             }
         }
 
+        internal async Task WriteMessagePostambleAsync(XmlDictionaryWriter writer)
+        {
+            if (Version.Envelope != EnvelopeVersion.None)
+            {
+                await writer.WriteEndElementAsync();
```
MarcWils, line 837: Not all writers support async IO. Not sure yet how to check if async is supported.

MarcWils:
@dotnet-policy-service agree

src/CoreWCF.Primitives/src/CoreWCF/Channels/MtomMessageEncoder.cs
```diff
@@ -561,7 +561,14 @@ private async Task WriteMessageInternalAsync(Message message, Stream stream, str
                 await xmlWriter.WriteEndDocumentAsync();
             }
 
-            await xmlWriter.FlushAsync();
+            if (xmlWriter.SupportsAsync())
```
MarcWils, line 564: Not really sure why this check is needed. But my failed without it.

src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlDictionaryWriterExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    internal static class XmlDictionaryWriterExtensions
+    {
+        private static readonly bool s_frameworkSupportsAsyncWriteOperations = Environment.Version.Major >= 6;
+
+        internal static bool SupportsAsync(this XmlDictionaryWriter writer)
+        {
+            // Unfortunately not every XmlWriter supports asynchronous IO.
+            // And there's no API to check for this.
+            return s_frameworkSupportsAsyncWriteOperations && writer is IAsyncXmlWriter;
```
mconnew, line 17: There's a second check you can also add to this. You can check `writer.Settings?.IsAsync ?? false`. This should light up this code path for more scenarios and make more things async.

mconnew: You still need your IAsyncXmlWriter method because XmlMtomWriter.Settings returns the Settings property from the inner XmlWriter, which in the scenarios I was investigating was always null.

src/CoreWCF.Primitives/src/CoreWCF/Channels/Message.cs
```diff
@@ -748,7 +751,14 @@ public virtual async Task OnWriteMessageAsync(XmlDictionaryWriter writer)
             // otherwise EnsureWriteMessageState would get called twice. Also see OnWriteMessage()
             // for the example.
             await OnWriteBodyContentsAsync(writer);
-            WriteMessagePostamble(writer);
+            if (writer.SupportsAsync())
```
mconnew, line 754: You probably want to update BodyWriterMessage.OnWriteMessageAsync with the same logic too. That class/method is further down in this same file.

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
```diff
@@ -126,6 +128,7 @@ public static IWebHostBuilder CreateWebHostBuilder<TStartup>(ITestOutputHelper o
             CreateBaseWebHostBuilder<TStartup>(outputHelper, callerMethodName)
             .UseKestrel(options =>
             {
+                options.AllowSynchronousIO = true;
```
mconnew, line 131: This enables synchronous IO for all tests. We don't want this always on. If a change is made which regresses the other scenarios and starts writing out synchronously, we might miss it and ship a pretty nasty bug. You should be able to configure kestrel in the test which calls this method. In the calling test you can do something like this:
```c#
var hostBuilder = ServiceHelper.CreateWebHostBuilder(_output, startupType);
hostBuilder.ConfigureServices(services =>
{
   services.Configure<KestrelServerOptions>(options =>
   {
       options.AllowSynchronousIO = true;
   });
});
IWebHost host = hostBuilder.Build();
```

mconnew: You could make a helper extension method in ServiceHelper for this to make it a one lines. Like this:
ServiceHelper.cs
```c#
public static IWebHostBuilder AllowSynchronousIO(this IWebHostBuilder webHostBuilder)
{
    webHostBuilder.ConfigureServices(services =>
    {
       services.Configure<KestrelServerOptions>(options =>
       {
           options.AllowSynchronousIO = true;
       });
    });
    return webHostBuilder;
}
```
In unit test method:
```c#
IWebHost host = ServiceHelper.CreateWebHostBuilder<Startup>(_output)
                .AllowSynchronousIO()
                .Build();
```

src/CoreWCF.Http/tests/Helpers/ServiceHelper.cs
```diff
@@ -302,6 +305,18 @@ public static Stream GetStreamWithStringBytes(string s)
             return stream;
         }
 
+        public static Stream GetAsyncStreamWithStringBytes(string s)
+        {
+            Stream inner = new MemoryStream();
+            PopulateStreamWithStringBytes(inner, s);
+#if NETFRAMEWORK
```
mconnew, line 312: I prefer to avoid conditional compilation whenever possible. It has a habit of breaking refactoring tooling and makes it harder to see coding mistakes in the IDE before doing a full multi-targeted build. Can you switch to a runtime check similar to what you did in XmlDictionaryWriterExtensions? It doesn't need to be cached in a static, you can just look it up each time to keep the code simpler.

src/CoreWCF.Http/tests/Helpers/AsyncOnlyStream.cs
```diff
@@ -0,0 +1,43 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CoreWCF.Http.Tests.Helpers
+{
+    internal class AsyncOnlyStream : Stream
```
mconnew, line 11: This is an optional suggestion, so feel free to not do it. This could be simplified by copying DelegatingStream from the implementation into the tests. You can then make AsyncOnlyStream derive from it, and only override the methods you want different behavior for. You would only need to override the 3 methods.

mconnew:
@MarcWils, this is a great contribution. I had a poke at this too, but I think yours is a little cleaner approach than the direction I was going (and yours was completed too whereas mine isn't). Just a few small changes and I'll merge this. The only real must have change is around not enabling synchronous IO for all tests as missing a regression in that area would be really bad.

MarcWils:
@mconnew , thanks for the review. Hopefully I've covered most of the comments now. Still have one or two open questions.

mconnew:
What are the questions? I'm not seeing any of my comments that you haven't addressed. Are these questions you have that you want to ask before this if merged?

MarcWils:
> What are the questions? I'm not seeing any of my comments that you haven't addressed. Are these questions you have that you want to ask before this if merged?

There's one more open question regarding BodyWriterMessage.OnWriteMessageAsync. I think it's already updated. Or am I missing something?

src/CoreWCF.Primitives/src/CoreWCF/Channels/Message.cs
mconnew: I missed that. Yes it is.

src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlDictionaryWriterExtensions.cs
mconnew: The inner writer supporting async is really only going to happen with a custom message encoder. This code path is also hit with all the other bindings etc. It's possible in the future that the runtime modifies one of the XmlWriter's or XmlDictionaryWriter's to support async in the future. Or someone could provide a custom message encoder where they supply a built in XmlWriter with the settings set specifying Async.

src/CoreWCF.Primitives/src/CoreWCF/Channels/MtomMessageEncoder.cs
```diff
@@ -561,7 +561,14 @@ private async Task WriteMessageInternalAsync(Message message, Stream stream, str
...
+            {
+                await xmlWriter.FlushAsync();
```
snakefoot, line 566: Any reason for not using `ConfigureAwait(false)` whenever using `await` ?

MarcWils: I think it's partly explained in https://github.com/CoreWCF/CoreWCF/pull/711

Issue #1516, opened 2024-12-18
Need test for Disco endpoint


mconnew:
### Duplicate ?

- [X] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.6.0

### Describe expected behavior

We need unit tests for the Disco endpoint. I accepted a PR to fix the Disco endpoint (#1514) as the change was trivial and obvious based on the error that was being seen.

### Describe actual behavior

 We should have a unit test to make sure we don't regress this in the future if the implementation is reverted to a Message based solution instead of the current implementation (WCF implements a full dispatched endpoint for this). We might do that if we implement the SOAP mex endpoint.

### Which binding

BasicHttp

### security

None

### Which .NET version

.NET 7

### Which os platform

Windows

mconnew:
Some tooling still uses the Disco endpoint (WCF Test Client for example) to discover the WSDL endpoint so despite everyone moving on from Disco a very long time ago, it's still needed.

PR #1518, opened 2024-12-21
Add Test for DiscoResult

stefannikolei:
Added Test for #1516

I had to make those classes public, because I had problems accessing the DiscoResult from the TestAssembly. I did not manage to get InternalsVisibleTo working. Probably need help with that, to not make those classes public

Is this test how you wanted it to be, or should it be done in a different way?

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Description/ServiceMetadataExtension.cs
@@ -439,7 +439,7 @@ namespace CoreWCF.Description
             }
         }
 
-        internal class HttpGetImpl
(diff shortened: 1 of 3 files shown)


mconnew:
The tests for metadata are done indirectly, basically set things up and make the http request and make sure the response is what's expected. Although the implementation is in CoreWCF.Primitives, we have a separate test project just for metadata. You can find the project at src\CoreWCF.Metadata\tests\CoreWCF.Metadata.Tests.csproj. Making internals public is something to be avoided otherwise you can't change the implementation later.

mconnew:
I just did a lot of work to write tests for the Html help page so while I was at it, I created tests for the disco endpoint. You can see that in PR #1524

PR #1528, opened 2025-01-14
Update dependencies in preparation for next release

mconnew:


src/CoreWCF.Http/tests/CoreWCF.Http.GeneratedOperationInvokers.Tests.csproj
```diff
@@ -1,5 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
-  <Import Project="CoreWCF.Http.Tests.csproj" />
+  <PropertyGroup>
+    <ImportDirectoryPackagesProps>false</ImportDirectoryPackagesProps>
+  </PropertyGroup>
+  <Import Project="CoreWCF.Http.Tests.proj" />
```
g7ed6e, line 5: I did not know that trick. Nice

src/CoreWCF.NetNamedPipe/src/CoreWCF/Channels/NamedPipeListener.cs
```diff
@@ -40,6 +40,7 @@ public NamedPipeListener(NamedPipeListenOptions options)
             _logger = options.ApplicationServices.GetRequiredService<ILogger<NamedPipeListener>>();
             _framingConnectionLogger = options.ApplicationServices.GetService<ILogger<FramingConnection>>();
             _transportConnectionManager = new TransportConnectionManager();
+            AddNamedPipeListenOptionsToConnectionContext(options);
```
g7ed6e, line 43: Does this relate to the dependencies update?

Directory.Build.props
```diff
@@ -7,14 +7,15 @@
     <ConfigPathSegment>Debug</ConfigPathSegment>
     <ConfigPathSegment Condition="'$(Configuration)'!=''">$(Configuration)</ConfigPathSegment>
     <LangVersion>11.0</LangVersion>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
```
g7ed6e, line 10: This property is declared in the Directory.Packages.props too. Is it necessary to declare it here ?

g7ed6e (approved):

src/CoreWCF.NetNamedPipe/src/CoreWCF/Channels/NamedPipeListener.cs
mconnew: It doesn't. One of the PR runs triggered this bug to manifest. We've had a couple of new bugs come up with .NET 9 having different timing and this was one of them.

Directory.Build.props
mconnew: Good catch. I'll remove it from here.

src/CoreWCF.Http/tests/CoreWCF.Http.GeneratedOperationInvokers.Tests.csproj
mconnew: It's not needed now though as I moved the Http.Tests.csproj bulk of work into a proj file. The problem I was having was there's now a build warning about pulling in the SDK twice, which resulted in Directory.Packages.Props getting pulled in twice which was causing a large amount of noise on restore. Moving it to a proj file solved that so I'm going to remove this now.

PR #1540, opened 2025-01-31
Final package version updates in preparation for 1.7 release

mconnew:


mconnew:
I couldn't update the Microsoft.IdentityModel packages to the latest version as they've introduced a bug which breaks package restore. I've opened issue https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3112 so hopefully it will get fixed. The 8.3.0 package doesn't have any security issues so we can stay at that version for now.

mconnew:
Merged without CR as this only updated test references and had no code changes.

PR #1562, opened 2025-03-13
Update dependencies

mconnew:
AspNetCore 2.1 has been republished as 2.3 to deal with the problem of tooling trying to update to 2.2.  
Includes other random updates.
WCF Client packages now fully compatible to be used on .NET Framework so removed the special handling for some tests needing to use the 4.3.x packages.  
Some dependency dropped a transitive dependency of Newtonsoft.Json so needed to add to WebHttp test project.
Bumped the ReferenceAssemblies for build tools tests from `.Default` to `.Net.Net80` as the default was incompatible with moving to AspNetCore 2.3.0 packages (default was netcoreapp3.0).


mconnew:
@g7ed6e can you review this?

g7ed6e (approved):

g7ed6e:
@mconnew Is there anything preventing to merge this PR ?

PR #1574, opened 2025-04-11
TimeoutException on claim lock operation

brunojuliao:
Added a 1s timeout.

That is needed for the EnsureInputClosedAsync. Otherwise, the SynchronizedMessageSource will acquire one lock, sit waiting for messages forever. Then a second call won't be able to acquire one lock (because the previous never reached the finally block). That will make it throw:
"System.TimeoutException: Cannot claim lock within the allotted timeout of 00:00:00. The time allotted to this operation may have been a portion of a longer timeout."

Also changed the WaitForMessageAsync catch block to just return null, instead of this TimeoutException.

### DIFF
--- a/src/Common/src/DuplexChannels/CoreWCF/Channels/SynchronizedMessageSource.cs
+++ b/src/Common/src/DuplexChannels/CoreWCF/Channels/SynchronizedMessageSource.cs
@@ -31,10 +31,7 @@ namespace CoreWCF.Channels
             }
             catch (OperationCanceledException)
             {
-                // TODO: Fix the timeout value reported
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new TimeoutException(
-                                                SR.Format(SR.WaitForMessageTimedOut, TimeSpan.Zero),
-                                                TimeoutHelper.CreateEnterTimedOutException(TimeSpan.Zero)));
+                return null;
             }
             finally
             {
--- a/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
+++ b/src/CoreWCF.NetFramingBase/src/CoreWCF/Channels/ServerFramingDuplexSessionChannel.cs
@@ -101,11 +101,16 @@ namespace CoreWCF.Channels
 
             public async Task<Message> ReceiveAsync(CancellationToken token)
             {
-                // TODO: Apply timeouts
+                using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(1));
+                using var combinedCts = CancellationTokenSource.CreateLinkedTokenSource(token, timeoutCts.Token);
+                var combinedToken = combinedCts.Token;
+
                 Message message;
                 ReadOnlySequence<byte> buffer = ReadOnlySequence<byte>.Empty;
                 for (; ; )
                 {
+                    combinedToken.ThrowIfCancellationRequested();
+                    
                     System.IO.Pipelines.ReadResult readResult = await _connection.Input.ReadAsync(token);
                     if (readResult.IsCompleted || readResult.Buffer.Length == 0)
                     {


brunojuliao:
> @brunojuliao please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

mconnew:
Sorry I missed this, I've been really busy with unrelated tasks. This is going to be a major performance problem, if not also causing other problems. The default receive timeout for NetTcp is 10 minutes. That means a client can have a connection open without sending any messages for up to 10 minutes. This change will cause an exception to be thrown if the client hasn't sent a message in the last second (which will be a common case). The solution is probably have a CancellationTokenSource which lives for the lifetime of the connection which is cancelled when EnsureInputClosedAsync is called. Then in one of the classes (I'd need to re-examine the class design to work out exactly which one), create a linked CancellationTokenSource with the passed in CancellationToken and this connection lifetime CancellationToken. We might be able to do some optimizations of not linking it if the passed in CancellationToken isn't cancellable (e.g. default was passed in).

brunojuliao:
Oh, I see... Just point me in the direction of the class that suits better this linked CancellationTokenSource, and I can try proceeding with your solution. Also, thank you for commenting and evaluating 😇 (I'm asking this to ease the implementation and later review)

If you don't mind, how can I run those tests on my local, if possible? So I can check why it's failing and try to address it 😅

mconnew:
Only some platforms failing suggests it was a build infrastructure intermittent issue, which is why I commented the azp command, to trigger re-running all the tests. That and I could no longer see the failure as the logs had been purged from the run due to time.  
Can you explain the symptoms you're attempting to fix?

brunojuliao:
Sure thing! When trying to dispose a class we use that inherits from "ServiceHost", it throws an exception. Digging why this happens, I found what I reported below to my team:

_Ok. I've downloaded the source code to understand the issue..._
image
 
_The comment below is a good lead to the issue:_
image
 
_So, what happens is the ReceiveAsync is a loop to process messages. The token is not being checked inside the loop. So, it never breaks the loop if no message is received. But still, the ReceiveAsync is called to "Ensure it is closed":_
image
image
 
_So, it receives a token (we were not passing one, fine). But to make it worse, the token was not checked, so it remains waiting indefinitely for a message that will never come and prevent the lock, in the first print, from being retrieved 😅_

mconnew:
I'm not seeing how you are getting that problem, and your commentary doesn't quite line up with the code. I just set a breakpoint on the first line of `TransportDuplexSessionChannel.EnsureInputClosedAsync` and took a look at the token passed in. It's cancellable, so default hadn't been passed in. I checked the call stack and the relevant part is the following:
```
CoreWCF.NetFramingBase.dll!CoreWCF.Channels.TransportDuplexSessionChannel.EnsureInputClosedAsync(System.Threading.CancellationToken token) Line 355	C#
CoreWCF.NetFramingBase.dll!CoreWCF.Channels.TransportDuplexSessionChannel.OnCloseAsync(System.Threading.CancellationToken token) Line 222	C#
CoreWCF.NetFramingBase.dll!CoreWCF.Channels.ServerFramingDuplexSessionChannel.OnCloseAsync(System.Threading.CancellationToken token) Line 79	C#
CoreWCF.Primitives.dll!CoreWCF.Channels.CommunicationObject.CloseAsync(System.Threading.CancellationToken token) Line 165	C#
CoreWCF.Primitives.dll!CoreWCF.Channels.CommunicationObject.CloseAsync() Line 118	C#
CoreWCF.NetFramingBase.dll!CoreWCF.Channels.ServerFramingDuplexSessionChannel.OnOpened.AnonymousMethod__6_0() Line 67	C#
```
The bottom of that call stack is a callback registered like this:
```c#
    IApplicationLifetime appLifetime = _serviceProvider.GetRequiredService<IApplicationLifetime>();
    _applicationStoppingRegistration = appLifetime.ApplicationStopping.Register(() =>
    {
        _ = CloseAsync();
    });
```
So it's not passing a timeout value, but if you check CommunicationObject.CloseAsync, it creates a CancellationToken from DefaultCloseTimeout which comes from the binding, and defaults to 1 minute. `TransportDuplexSessionChannel.EnsureInputClosedAsync` calls `SynchronizedMessageSource.ReceiveAsync` passes that token to `ServerSessionConnectionMessageSource.ReceiveAsync` which passes the token to `_connection.Input.ReadAsync`. So when the timeout is reached (default of 1 minute), that `_connection.Input.ReadAsync` call will get cancelled.  

As far as I can see, everything is plumbed up correctly. In CoreWCF, we don't have the `ServiceHost` class. We instantiate an instance of `ServiceHostObjectModel<TService>` which is an internal type that derives from `ServiceHostBase`. The name comes from the fact that it doesn't host anything (unlike in .NET Framework), but is the object model used to configure everything (to provide as much compatibility with .NET Framework as possible) and then the relevant pieces and extracted and used in the actual hosting.  

Can you provide the full exception details? The ideal thing would be a minimal repro app.

brunojuliao:
I am trying to assemble something reproducible to share here. It took a while to even recollect the details. Sorry... For now, I will paste the exceptions here:

As it's an AggregateException, there are 3:
**Exception 1**
System.ExecutionEngineException: Exception of type 'System.ExecutionEngineException' was thrown.
 ---> System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Reading is not allowed after reader was completed.)
 ---> System.InvalidOperationException: Reading is not allowed after reader was completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_NoReadingAllowed()
   at System.IO.Pipelines.Pipe.ReadAsync(CancellationToken token)
   at CoreWCF.Channels.Framing.NamedPipeExceptionConvertingDuplexPipe.NamedPipeExceptionConvertingPipeReader.ReadAsync(CancellationToken cancellationToken)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.ServerSessionConnectionMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.SynchronizedMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.EnsureInputClosedAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.CommunicationObject.CloseAsync(CancellationToken token)
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace --- 

**Exception 2**
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Reading is not allowed after reader was completed.)
 ---> System.InvalidOperationException: Reading is not allowed after reader was completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_NoReadingAllowed()
   at System.IO.Pipelines.Pipe.ReadAsync(CancellationToken token)
   at CoreWCF.Channels.Framing.NamedPipeExceptionConvertingDuplexPipe.NamedPipeExceptionConvertingPipeReader.ReadAsync(CancellationToken cancellationToken)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.ServerSessionConnectionMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.SynchronizedMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.EnsureInputClosedAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.CommunicationObject.CloseAsync(CancellationToken token)
   --- End of inner exception stack trace ---

**Exception 3**
System.InvalidOperationException: Reading is not allowed after reader was completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_NoReadingAllowed()
   at System.IO.Pipelines.Pipe.ReadAsync(CancellationToken token)
   at CoreWCF.Channels.Framing.NamedPipeExceptionConvertingDuplexPipe.NamedPipeExceptionConvertingPipeReader.ReadAsync(CancellationToken cancellationToken)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.ServerSessionConnectionMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.SynchronizedMessageSource.ReceiveAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.EnsureInputClosedAsync(CancellationToken token)
   at CoreWCF.Channels.TransportDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.ServerFramingDuplexSessionChannel.OnCloseAsync(CancellationToken token)
   at CoreWCF.Channels.CommunicationObject.CloseAsync(CancellationToken token)

And one correction, you are absolutely right. The ServiceHost is a class we created internally. I'm sorry for missing this.

The snippet where this happens is:
        public async Task CloseAsync()
        {
            if (Application != null) {
                await Application.StopAsync().ConfigureAwait(false);
                await Application.DisposeAsync().ConfigureAwait(false);
                Application = null;
            }
        }

The funny thing is, these exceptions happen as unobserved task exceptions, which closes our application. There's one client connected in when this happens and we are trying to change the address (we ask the client to gracefully close. It's a .Net Framework client using regular WCF. The server is using CoreWCF).

I will keep working on sample. Once again, thank you very much for even taking time to read, try things, and share the results 😇

mconnew:
@brunojuliao, are you able to build and test my changes on this branch: https://github.com/mconnew/CoreWCF/tree/ImproveNamedPipeClosePath

It should both fix the root cause of the problem (reading from a closed PipeReader and letting an unexpected exception bubble out), but also if another exception comes from somewhere else, it will be gracefully handled and ignored and not left unhandled.

brunojuliao:
I will try it now. Let you know soon.

PR #1575, opened 2025-04-13
Update the Negotiate implementation for .NET 8+ compatibility

### LINKED ISSUE #978: NegotiateInternalState should use the public NegotiateAuthentication API with .NET 7
Following the discussion in PR #917, `CoreWCF.Security.NegotiateInternal.NegotiateInternalState` should be updated to use the new public API in .NET 7 : [`NegotiateAuthentication`](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.negotiateauthentication?view=net-7.0). This should improve the maintainability of the code since the public signature is stable. However it must still be invoked through reflection as CoreWCF doesn't multi-target for different .NET versions.
The [asp.net core PR](https://github.com/dotnet/aspnetcore/pull/42364) can be used as a reference for the implementation. But unlike asp.net core, CoreWCF also uses `IsValidContext` and `Encrypt`. Encrypt and Decrypt should be equivalent to Wrap and Unwrap. However IsValidContext has no clear equivalent. This point has to be investigated.

nmangue:
Fix an issue on the use of Windows client credentials with a WsHttpbinding using TransportWithMessageCredential and also the use of Transport Security with a NetTcpBinding.
The issue was caused by the reflection on internal dotnet code that was refactored in .NET 8. This PR fixes #978.

* As stated in the issue, `IsValidContext` has no match in the new public API. After reading the original implementation, I decided to track if the last `GetOutgoingBlob` call has not returned an error code. The implementation works on the happy path but I don't know if it matches the original intent
* This heavily uses reflection code to work. Having CoreWCF target .NET 8 would greatly simplify the implementation


nmangue:
@dotnet-policy-service agree

mconnew:
@nmangue, I rebased your work, and made a few changes. I removed the older implementations as they aren't supported any more. I also removed the IBufferedWriter implementation as PipeWrite.Create can create a buffered writer from a MemoryStream. I'm always of the mind to own less code unless it's absolutely needed.  
I also went and updated the WCF Client to have this feature too. It isn't shipped yet, but the idea is to be able to have tests that run as part of the CI run, at least for Windows. Once that's released, I'll enable the tests to run (which I also wrote and added to your PR, and work against the unreleased WCF Client).

Issue #1582, opened 2025-05-13
[Bug]: Streamed BasicHttpBinding with TransportWithMessageCredential fails to add SOAP security header


VitezslavImrysek:
### Duplicate ?

- [x] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.7.0

### Describe expected behavior

WCF client (.NET and .NET Framework) should be able to communicate with CoreWCF endpoint configured with BasicHttpBinding (TransferMode = Streamed) and TransportWithMessageCredential security.

### Describe actual behavior

CoreWCF service method gets invoked but WCF call fails on client-side (on both .NET and .NET Framework) with MessageSecurityException. This is apparently due to a missing Security header in the reply SOAP message. 

Exception:
System.ServiceModel.Security.MessageSecurityException
  HResult=0x80131500
  Message=Security processor was unable to find a security header in the message. This might be because the message is an unsecured fault or because there is a binding mismatch between the communicating parties.   This can occur if the service is configured for security and the client is not using security.
  Source=System.ServiceModel.Primitives

### Which binding

BasicHttp

### security

TransportWithMessageCredential

### Which .NET version

.NET 8

### Which os platform

Windows

### Code snippet used to reproduce the issue

```c#
namespace Server
{
    public class Program
    {
        public const int HTTPS_PORT = 8443;

        public static void Main(string[] args)
        {
            IWebHost host = CreateWebHostBuilder(args).Build();
            host.Run();
        }

        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
            .UseKestrel(options =>
            {
                options.AllowSynchronousIO = true;
                options.ListenLocalhost(HTTPS_PORT, listenOptions =>
                {
                    listenOptions.UseHttps();
                    if (Debugger.IsAttached)
                    {
                        listenOptions.UseConnectionLogging();
                    }
                });
            })
            .UseStartup<WSHttpUserPassword>();
    }

    public class WSHttpUserPassword
    {
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddAuthentication();
            services.AddServiceModelServices();
        }

        public void Configure(IApplicationBuilder app)
        {
            var streamBinding = new BasicHttpBinding(BasicHttpSecurityMode.TransportWithMessageCredential);
            streamBinding.Security.Message.ClientCredentialType = BasicHttpMessageCredentialType.UserName;
[...]

PR #1583, opened 2025-05-13
Fix streamed BasicHttpBinding with TransportWithMessageCredential

VitezslavImrysek:
Fixes issue: #1582 

The issue appears to be specific to streamed and secured messages, as can be seen in `HttpChannelHelper.SendAsync`. When `TransferMode` is set to `Streamed`, the method ends up calling `MessageEncoder.WriteMessageAsync` instead of the synchronous `WriteMessage`. 
This leads us to the `SecurityAppliedMessage`, where only the synchronous `OnWriteMessage` method is implemented — this is where the security header is added to the SOAP message. However, the asynchronous counterpart, `OnWriteMessageAsync` - which is the one that ends up being invokned, still uses the base `Message.OnWriteMessageAsync` implementation which doesn't handle anything security-related.

This appears to be an oversight dating back to when asynchronous methods were introduced. The fix here is to just call the the synchronous implementation.

### DIFF
--- a/src/CoreWCF.Http/tests/BasicHttpTransportMessage.cs
+++ b/src/CoreWCF.Http/tests/BasicHttpTransportMessage.cs
@@ -39,6 +39,12 @@ namespace BasicHttp
             yield return new[] { typeof(BasicHttpTransportWithMessageCredentialWithUserName<CustomAsynchronousTestValidator>) };
         }
 
+        public static IEnumerable<object[]> GetStreamedTestsVariations()
+        {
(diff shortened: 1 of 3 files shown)


VitezslavImrysek:
@dotnet-policy-service agree

g7ed6e (approved):
LGTM

mconnew:
I'm going to reopen the issue after this is merged as I think we can do a better solution making more things async. But this fixes the problem so I'm not going to ask for any improvements beyond this.

mconnew:
Scratch that, I'll open a new issue to keep it simple. I'll close the bug for this.

PR #1585, opened 2025-05-16
Add more asynchronism to SecurityAppliedMessage.OnWriteMessageAsync

### LINKED ISSUE #1584: [Feature Request]: Implement OnWriteMessageAsync for SecurityAppliedMessage
As per #1582, we are missing an async implementation of SecurityAppliedMessage.OnWriteMessageAsync. We have a temporary workaround of calling the synchronous implementation from the Async method, but we should try to do as much asynchronously as possible.

g7ed6e:
resolves #1584

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Security/SecurityAppliedMessage.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Security/SecurityAppliedMessage.cs
@@ -150,6 +150,14 @@ namespace CoreWCF.Security
 
         public override async Task OnWriteMessageAsync(XmlDictionaryWriter writer)
         {
+            // With NetFx the writer passed to SecurityAppliedMessage.OnWriteMessageAsync is of type XmlUtf8TextWriter which doesn't implement WriteEndElementAsync.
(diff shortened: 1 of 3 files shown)


mconnew:
This is failing on NetFx because the writer passed to `SecurityAppliedMessage.OnWriteMessageAsync` is of type `XmlUtf8TextWriter` which doesn't implement `WriteEndElementAsync`. On .NET, the type is `XmlDictionaryAsyncCheckWriter` which wraps `XmlUtf8TextWriter`, both of which do implement `WriteEndElementAsync`. The solution is to do a platform check at the start of `SecurityAppliedMessage.OnWriteMessageAsync`, and if it's .NET Framework, call `OnWriteMessage`.

g7ed6e:
> This is failing on NetFx because the writer passed to `SecurityAppliedMessage.OnWriteMessageAsync` is of type `XmlUtf8TextWriter` which doesn't implement `WriteEndElementAsync`. On .NET, the type is `XmlDictionaryAsyncCheckWriter` which wraps `XmlUtf8TextWriter`, both of which do implement `WriteEndElementAsync`. The solution is to do a platform check at the start of `SecurityAppliedMessage.OnWriteMessageAsync`, and if it's .NET Framework, call `OnWriteMessage`.

Thanks for the hint @mconnew. I updated the PR

PR #1592, opened 2025-05-22
Move MessagePropertyAttribute to correct namespace

g7ed6e:


mconnew:
:ShipIt: Once that extra blank line is removed, this is good to merge

PR #1597, opened 2025-05-30
Prevent unhandled exceptions during service shutdown which can kill a process

mconnew:
See the later comments on #1574 for context.

Basically netnamedpipe client from NetFx still connected when closing the host causes the process to quit unexpectedly due to unhandled exception.

This fixes it 3 ways:
1. Tracks when we complete a PipeReader and tries to avoid calling ReadAsync if it's completed
2. If we do call ReadAsync on a completed PipeReader, handle the exception which is thrown and convert to a CoreWCF exception as if the connection was closed
3. At the top level application stopping event registered for the channel, don't allow any exceptions from calling CloseAsync to be unhandled


mconnew:
@brunojuliao, this is the PR to fix your issue.

g7ed6e (approved):

Issue #701, opened 2022-06-02
NullReferenceException in RequestDelegateHandler when wrong HTTP method is requested.


gvkries:
I have an duplex contract with a NetHttp(s)Binding which is using WebSockets. If this endpoint gets called with a GET request, an `NullReferenceException` is getting thrown by `RequestDelegateHandler.HandleRequest()`:

> System.NullReferenceException: Object reference not set to an instance of an object.
   at CoreWCF.Channels.RequestDelegateHandler.HandleRequest(HttpContext context)
   at CoreWCF.Channels.ServiceModelHttpMiddleware.InvokeAsync(HttpContext context)
   at CoreWCF.Description.ServiceMetadataExtension.HttpGetImpl.<>c__DisplayClass39_0.<<MetadataMiddleware>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at CoreWCF.Channels.MetadataMiddleware.InvokeAsync(HttpContext context)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Serilog.AspNetCore.RequestLoggingMiddleware.Invoke(HttpContext httpContext)
   at Prometheus.HttpMetrics.HttpRequestDurationMiddleware.Invoke(HttpContext context)
   at Prometheus.HttpMetrics.HttpRequestCountMiddleware.Invoke(HttpContext context)
   at Prometheus.HttpMetrics.HttpInProgressMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

The correct line is 
>   at CoreWCF.Channels.RequestDelegateHandler.<HandleRequest>d__15.MoveNext() in /_/src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs:line 104

.NET Framework is returning status code 405 (method not allowed) in this case. Maybe it is IIS that is returning this status code, I haven't checked that in detail. But I think WCF should not throw an NRE.

mconnew:
Leaving open for community contribution as this is a case of a failing scenario failing the wrong way rather than a successful scenario failing so is lower priority.

Issue #756, opened 2022-07-17
NetHttpBinding not working as expected


spur19:
Hi, I am using the NetHttpBinding with a sample application, and it gives me this error:

**Server Side Error** 
```
fail: Microsoft.AspNetCore.Server.HttpSys.MessagePump[0]
      ProcessRequestAsync
System.NullReferenceException: Object reference not set to an instance of an object.
   at CoreWCF.Channels.RequestDelegateHandler.HandleRequest(HttpContext context)
   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
   at CoreWCF.Channels.ServiceModelHttpMiddleware.InvokeAsync(HttpContext context)
   at CoreWCF.Channels.MetadataMiddleware.InvokeAsync(HttpContext context)
   at Microsoft.AspNetCore.Server.HttpSys.MessagePump.ProcessRequestAsync(Object requestContextObj)
   at Microsoft.AspNetCore.Server.HttpSys.MessagePump.ProcessRequestAsync(Object requestContextObj)

```
**Client Side Error**
```
Unhandled exception. System.ServiceModel.CommunicationException: The server did not provide a meaningful reply; this might be caused by a contract mismatch, a premature session shutdown or an internal server error.
   at System.ServiceModel.Channels.ServiceChannel.Call(String action, Boolean oneway, ProxyOperationRuntime operation, Object[] ins, Object[] outs, TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannel.Call(String action, Boolean oneway, ProxyOperationRuntime operation, Object[] ins, Object[] outs)
   at System.ServiceModel.Channels.ServiceChannelProxy.InvokeService(MethodCall methodCall, ProxyOperationRuntime operation)
   at System.ServiceModel.Channels.ServiceChannelProxy.Invoke(MethodInfo targetMethod, Object[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Reflection.DispatchProxyGenerator.Invoke(Object[] args)
   at generatedProxy_1.ReverseString(String )
   at NetCoreClient.Program.CallBasicHttpBinding(String hostAddr) in C:\Users\t-spuri\OneDrive - Microsoft\Desktop\CoreWCFEcho\EchoClient\Program.cs:line 46
[...]

PR #1599, opened 2025-06-03
Return bad request instead of throwing null reference exception (#701)

fremo917:
Return bad request instead of throwing null reference exception.

Fixes: #701, #756

### DIFF
--- a/src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
+++ b/src/CoreWCF.Http/src/CoreWCF/Channels/RequestDelegateHandler.cs
@@ -125,6 +125,13 @@ namespace CoreWCF.Channels
 
             if (!context.WebSockets.IsWebSocketRequest)
             {
+                if (WebSocketOptions != null && _replyChannelDispatcher == null && _replyChannelDispatcherTask == null)
+                {
+                    context.Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                    context.Features.Get<IHttpResponseFeature>().ReasonPhrase = SR.WebSocketEndpointOnlySupportWebSocketError;
+                    return;
+                }
+
                 if (_replyChannelDispatcher == null)
                 {
                     _replyChannelDispatcher = await _replyChannelDispatcherTask;


g7ed6e:
Thanks @fremo917 for the contribution. Would you be able to add a unit test covering that case ? One issue you linked to that PR mentions that .Net Framework http status code is 405, would you be able to check CoreWCF behaves the same as WCF ?

mconnew:
This test is validating a failure case, so the service doesn't need any implementation. Can you modify the contract and service implementation to be the bare minimum. Just give it a simple `string Echo(string message)` operation. The important thing is that a callback contract is defined, which can have a simple dummy method too as it won't get called. I'm going to take a look at the test failure and see whether you need to modify your test case assert or if something else needs to happen with the implementation.

mconnew:
@fremo917, the tests are failing because the exception message is:
> The remote server returned an unexpected response: (400) This service only supports WebSocket connections..

But you are calling `Assert.StartsWith("This service only supports WebSocket connections.", exception.Message);`. You can't do an `Assert.Equals` with the longer error message as the WCF client is localized and the first part of the error message could be in another language. You also can't do `EndsWith` as different languages might not place the returned error text as the last thing. The only reliable option is to use contains like this:
```c#
Assert.Contains("This service only supports WebSocket connections.", exception.Message);
```

g7ed6e:
> > Thanks @fremo917 for the contribution. Would you be able to add a unit test covering that case ? One issue you linked to that PR mentions that .Net Framework http status code is 405, would you be able to check CoreWCF behaves the same as WCF ?
> 
> 
> 
> I have trouble reproducing the same behaviour in WCF. I can neither figure out how to make NetHttpBinding listen to both web socket and http or make it listen to only web socket.

According to [this](https://learn.microsoft.com/en-us/dotnet/framework/wcf/feature-details/using-the-nethttpbinding) the `NetHttpBinding` detects and adapts its behavior depending on the contract and can be overriden with an attribute named `WebSocketTransportUsage`.

mconnew:
The behavior is [documented in the WebSocketTransportUsage](https://learn.microsoft.com/en-us/dotnet/api/system.servicemodel.channels.websockettransportusage?view=net-9.0-pp) enum values:
Name | Value | Description
-- | -- | --
WhenDuplex | 0 | Indicates WebSocket transport will be used for duplex service contracts only.
Always | 1 | Indicates WebSocket transport will always be used.
Never | 2 | Indicates WebSocket transport will never be used.

The default is `WhenDuplex`, so if there's a callback contract, it uses WebSockets, no callback contract, it uses HTTP. You can specify `Always` so it uses WebSockets even without a callback contract. `Never` is really just a check to throw an exception if you add a callback contract at a later time when you have clients expecting HTTP. There's no other practical reason.

mconnew:
Thank you @fremo917. I've got a couple of other things to sort out (signing is broken right now), but I hope to make a new release soon.

fremo917:
@g7ed6e @mconnew Yes I had tried `binding.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always`. I tried again and I get 400 Bad Request from _ws_ and 405 Method Not Allowed from _http_.
With `WebSocketTransportUsage.Never` and `WhenDuplex` I get 400 from both _ws_/_http_. I still can't get 101/200 from WebSocket. Maybe that isn't needed?

Does the result from `Always` mean that this PR should have included `context.Response.StatusCode = (int)HttpStatusCode.MethodNotAllowed;` instead?

### HTTP 


### WS

PR #1601, opened 2025-06-06
Fix non-structured log records

MaceWindu:
Fix found logger calls where structured logging was not used


g7ed6e (approved):
LGTM

MaceWindu:
Initially I was "triggered" by a lot of different `base address xxx` messages in our logs instead of single message template with different parameter values. I just inspected logging calls manually (CoreWCF doesn't have a lot of them, so it wasn't hard).

> Is there a code analyzer which flagged them?

Unfortunatelly I'm not aware of such analyzer. Only found https://github.com/fedarovich/structured-logging-analyzers which looks abandoned. Not even sure it works (not tried to use it).

Also such analyzers will be useful only when used with Microsoft logging APIs. E.g. in my code I use wrapper over `M.E.Logging` with custom interpolation handlers.

PR #1603, opened 2025-06-14
Add ServiceName to the exception thrown when service does not have a default constructor

g7ed6e:


mconnew:
I didn't see this before I got copilot to work on this same issue. I think that other PR covers all the cases you have here, if so, we can close this.

PR #1649, opened 2025-11-10
fix vulnerable package Microsoft.AspNetCore.Server.Kestrel.Core

StojanovicMilos:
Microsoft.AspNetCore.Server.Kestrel.Core is vulnerable on versions <= 2.3.0: https://github.com/advisories/GHSA-5rrx-jjjq-q2r5

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -26,7 +26,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.Hosting.Server.Abstractions" Version="2.3.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Server.HttpSys" Version="2.3.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.3.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="2.3.6" />
     <PackageVersion Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" Version="2.3.0" />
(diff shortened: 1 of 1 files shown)


mconnew:
Copilot just gave me this snippet for getting the major version number:
```c#
public static int GetDotNetMajorVersion()
{
    // Example: ".NET 8.0.2" or ".NET Framework 4.8.1"
    string description = RuntimeInformation.FrameworkDescription;

    // Find the first digit sequence after ".NET"
    var parts = description.Split(' ');
    foreach (var part in parts)
    {
        if (char.IsDigit(part[0]))
        {
            // Split by '.' and take the first number
            if (int.TryParse(part.Split('.')[0], out int major))
            {
                return major;
            }
        }
    }

    throw new InvalidOperationException("Unable to determine .NET version.");
}
```
I also checked, and TheoryAttribute derives from FactAttribute, so you can do the exact same thing. We don't need to skip specific iterations, just the whole test, so a custom theory can just set the Skip property to skip the entire test.

g7ed6e:
> Hi @g7ed6e , I've fixed the pipelines in PR #1653 per @mconnew's suggestion, so we'll have to merge that one first, unless it's been fixed already...

@StojanovicMilos .NET10 unit tests pass in CI pipeline I think #1653 won't be needed.

/cc @mconnew

StojanovicMilos:
> > Hi @g7ed6e , I've fixed the pipelines in PR #1653 per @mconnew's suggestion, so we'll have to merge that one first, unless it's been fixed already...
> 
> @StojanovicMilos .NET10 unit tests pass in CI pipeline I think #1653 won't be needed.
> 
> /cc @mconnew

Oh, I see, well great then, we can discard #1653 and merge this one if you're happy with it?

mconnew:
I've created and merged a PR updating all the dependencies so closing this PR as the work has been done.

PR #1653, opened 2025-11-28
add new attribute to skip failing tests on .net 10

StojanovicMilos:
Per conversation from #1649, I added new attributes that will skip (currently) failing tests on .net 10 until it's resolved. I made attributes configurable so that it's easy to change min and max supported .net (core) version.


g7ed6e:
Closing that PR as #1649 CI is green

PR #1655, opened 2025-12-09
Move to v4 of SonarCloud pipeline tasks as previous versions deprecated

mconnew:
We're getting build errors on the CI build due to the SonarCloud tasks being deprecated.


mconnew:
Merging without review as this is a pipeline yaml file change that's only verifiable after it's been merged due to the steps only working in a CI pipeline run.

PR #1710, opened 2026-04-24
Correctly propagate the reader quotas to the TextMessageEncoder

### LINKED ISSUE #1704: [Bug]: MaxArrayLength quota on WebHttpBinding is ignored when using TransferMode.Streamed
### Duplicate ?

- [x] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.8.0

Hi,

We are in the process of migrating another of our server applications from .net framework to .net8.0, and for this we switched from System.ServiceModel to CoreWcf.

The application uses a WebHttpBinding, which is defined as follows:

```
var binding = new WebHttpBinding();
binding.MaxReceivedMessageSize = 500 * 1024 * 1024;
binding.SendTimeout = new TimeSpan(2, 0, 0);
binding.TransferMode = TransferMode.Streamed;
binding.ReaderQuotas.MaxArrayLength = 2^31 - 1;
```

Even though the binding uses **TransferMode.Streamed**, there are some byte[] values that can become quite large. Hence the large **ReaderQuotas.MaxArrayLength**.

We expect this value to be honored.

### Describe actual behavior

After switching to .net8 and CoreWCF however, the server failed to deserialize many messages with the following error.  

> The server encountered an error processing the request. The exception message is 'The formatter threw an exception while trying to deserialize the message: There was an error while trying to deserialize parameter http://tempuri.org/:file. The InnerException message was 'There was an error deserializing the object of type XXX. The maximum array length quota (16384) has been exceeded while reading XML data. This quota may be increased by changing the MaxArrayLength property on the XmlDictionaryReaderQuotas object used when creating the XML reader. Line 1, position 2248.

I realized that when switching to **TransferMode.Buffered**, then it works. Of course this is no real solution because we have some methods in the contract that require the streamed transfer mode.

I tried to debug the corewcf assemblies and found out that at some point a reader is created without passing the bindings readerquotas.

The WebMessageEncoderFactory class has a private property ‘private MessageEncoder TextMessageEncoder’.
In this property an object of type TextMessageEncodingBindingElement is created without passing the ReaderQuotas.
From that point on the deserialization of the input object uses the default ReaderQuotas which are too low in our case.

### Which binding

WebHttp

### security

Transport

### Which .NET version

.NET 8

### Which os platform

Windows

```c#

```

### Stacktrace if any

```shell

```

mconnew:
Fixes #1704

### DIFF
--- a/src/CoreWCF.WebHttp/src/CoreWCF/Channels/WebMessageEncoderFactory.cs
+++ b/src/CoreWCF.WebHttp/src/CoreWCF/Channels/WebMessageEncoderFactory.cs
@@ -135,7 +135,13 @@ namespace CoreWCF.Channels
                         {
                             if (_textMessageEncoder == null)
                             {
-                                _textMessageEncoder = new TextMessageEncodingBindingElement(MessageVersion.None, _writeEncoding).CreateMessageEncoderFactory().Encoder;
+                                TextMessageEncodingBindingElement textBindingElement = new TextMessageEncodingBindingElement(MessageVersion.None, _writeEncoding)
+                                {
+                                    MaxReadPoolSize = _maxReadPoolSize,
+                                    MaxWritePoolSize = _maxWritePoolSize,
+                                };
+                                _readerQuotas.CopyTo(textBindingElement.ReaderQuotas);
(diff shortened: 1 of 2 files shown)


mconnew:
Change and fix created by copilot, I'm the reviewer of the work that copilot did.

PR #1711, opened 2026-04-27
Fix template tests on Windows and isolate from inherited CPM

mconnew:
The template tests recently regressed on Windows for net4xx variations with CS8370 errors (C# 'file types' / 'module initializers' not available in C# 7.3). Two underlying issues caused this:

1. CoreWCF.Primitives 1.9.0 began defaulting EnableCoreWCFOperationInvokerGenerator to true for all consumers. The generator emits C# 11 'file' types and '[ModuleInitializer]', but the .NET SDK defaults LangVersion to 7.3 for .NETFramework targets (and to lower-than-11 for .NETStandard, net5.0, net6.0). Default-on with no LangVersion bump produces uncompilable code for those targets.

2. Prepare-Run.ps1's `dotnet add package -v <version>` calls have been silently failing since Central Package Management was enabled at the repo root (CPM rejects versions on PackageReference items). The script ignored exit codes, so the template was packed referencing the wildcard version and the tests have been pulling whatever's on nuget.org instead of the locally-built bits.

Fixes:

- CoreWCF.Primitives.targets: Default EnableCoreWCFOperationInvokerGenerator to true only on .NETCoreApp 7.0 or later (where the SDK default LangVersion is at least 11). All other TFMs default to false and use the existing reflection-based fallback. Consumers can still opt in explicitly on older targets by raising LangVersion themselves.

- CoreWCFService.csproj template: emit <LangVersion>Latest</LangVersion>
  whenever isNetFramework is true (was previously gated on enableOperationInvokerGenerator), so net4xx projects created from the template can compile generator output when the user opts in.

- Prepare-Run.ps1:
  * Add fail-fast error handling ($ErrorActionPreference='Stop' plus an Invoke-Checked wrapper around every dotnet call).
  * Drop empty sentinel Directory.Build.props/targets/Packages.props files at src/CoreWCF.Templates/src/templates/ before mutating the template csproj, so MSBuild stops walking upward and doesn't inherit the repo-root CPM. This unblocks `dotnet add package -v`.
  * Make `dotnet nuget add source` idempotent.
  * Remove the duplicate `dotnet nuget add source` call that was erroring.
  * Fix pre-existing typo: Directory.Build.target -> .targets.

- Clean-Run.ps1: remove the new template-source sentinel files.

- Run-Templates-UnitTests.ps1: move Prepare-Run inside the try block so Clean-Run still runs if Prepare-Run throws partway through.

- CoreWCF.Templates.csproj: exclude the new Directory.Packages.props sentinel from template package content.

- Project.cs (test helper): include the local Artifacts feed in the test's `dotnet restore -s` invocation so the locally-built pre-release packages referenced by the generated csproj can be resolved.

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
+++ b/src/CoreWCF.Primitives/src/CoreWCF.Primitives.targets
@@ -1,7 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <EnableCoreWCFOperationInvokerGenerator Condition="'$(EnableCoreWCFOperationInvokerGenerator)'==''">true</EnableCoreWCFOperationInvokerGenerator>
+    <!-- Default the OperationInvokerGenerator to enabled only on TFMs where the SDK's default C#
+         language version is >= 11 (the version required by the generator's emitted 'file' types and
+         '[ModuleInitializer]'). That is .NETCoreApp 7.0 and later. .NETFramework, .NETStandard, and
+         older .NETCoreApp targets remain off-by-default and use the reflection-based fallback at
+         runtime. Consumers on those targets can still opt in by explicitly setting
+         EnableCoreWCFOperationInvokerGenerator=true and raising LangVersion to 11 or higher. -->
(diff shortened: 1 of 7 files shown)


g7ed6e:
@mconnew good catch. Thanks!

g7ed6e (approved):

PR #1722, opened 2026-05-11
Removing sparse checkout as it breaks gh-pages work

mconnew:


mconnew:
Skipping review as no product changes

PR #1724, opened 2026-05-12
Fix restore step for CI builds

mconnew:
Trying to get CI build working now.


mconnew:
Merging without review as change is in CI pipeline run so won't change anything in PR build.

PR #1726, opened 2026-05-12
Making signing vars into secrets and configuring enviornment

mconnew:


mconnew:
Skipping review as no product changes, infrastructure changes to CI workflow to sign packages

PR #1727, opened 2026-05-13
Add missing checkout step to sign and publish

mconnew:


mconnew:
Skipping review as no changes that would affect PR build, or product code.

PR #1730, opened 2026-05-14
Have sync workflow have knowledge about limits synching .github/workflows/ changes

mconnew:


mconnew:
Doesn't affect PR run so skipping review

Issue #1677, opened 2026-01-27
[Bug]: Memory leak in OpenTelemetry implementation


jklawrence:
### Duplicate ?

- [x] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.8.0

### Describe expected behavior

`System.Diagnostics.Activity` instances should always be stopped after requests are handled or timed out.

### Describe actual behavior

`System.Diagnostics.Activity` instances do not get stopped consistently under heavy load, which leads to ever-increasing Gen2 heap size until the application runs out of memory.

This leak occurs on both .NET 8 and .NET 10.

I think this may be because the [Activity is stored in the ImmutableDispatchRuntime](https://github.com/CoreWCF/CoreWCF/blob/d671ed576441f2df8ad9b59fff16bcc9624ef8f9/src/CoreWCF.Primitives/src/CoreWCF/Dispatcher/ImmutableDispatchRuntime.cs#L30), which I believe has only a single instance per channel endpoint? So if a request comes in before the previous request has finished being handled, the first request's Activity will be lost and never closed. Potentially there are other scenarios where the Activity is not closed too.

If there is no `ActivityListener` for CoreWCF Activities, there is no leak.

### Which binding

CustomBinding

### security

None

### Which .NET version

.NET 10

### Which os platform

Windows

### Code snippet used to reproduce the issue

```c#

```

### Stacktrace if any

```shell

```

PR #1759, opened 2026-08-06
Fix Activity memory leak in dispatcher under concurrency (#1677)

g7ed6e:
The OpenTelemetry Activity for a request was stored in a single instance field (ImmutableDispatchRuntime._activity) that is shared by every request flowing through an endpoint's dispatch runtime. Under concurrent load a later request overwrote the field before an earlier request reached the code that stops the Activity, so the earlier Activity was never stopped. The never-stopped Activity instances accumulated on the Gen2 heap and eventually exhausted memory. This only manifested when an ActivityListener was registered for the CoreWCF ActivitySource.

Move the Activity to per-request state on MessageRpc so each request owns and stops its own Activity, eliminating the cross-request race. This mirrors the original WCF design, which carried the activity on the RPC.

Add a regression test that dispatches several requests concurrently, holds them all inside the operation via a gate so their Activities are alive simultaneously, and asserts every started Activity is also stopped. The test fails before the fix (only the last Activity is stopped) and passes after it.


randy-wilkinson-allscripts:
What is the status of this PR?

PR #1765, opened 2026-08-08
perf(mtom): accumulate MIME headers into a StringBuilder

### LINKED ISSUE #1764: [Bug]: MimeHeaderReader.AppendValue / AppendName uses string concatenation - O(N^2) on long/folded MIME headers
### Duplicate ?

- [x] I have searched issues/discussions and did not find other issues/discussions reporting this bug.

### Product version

1.9.1

**Description**
MimeHeaderReader.AppendValue/AppendName perform string += per 1024-byte buffer fill, giving O(N²/1024) copies on unbounded folded MIME-part header values. Bounded by the Kestrel body size limit (28.6 MB), which permits hundreds of GB of copy work per request.

**Suggested Fix**
Move from string concatenation to StringBuilder when processing MTOM headers. StringBuilder is Mutable while String itself is immutable which is why the copy take place. StringBuilder can allow this to run linearly since it is modifying an internal buffer without creating new objects.

### Which binding

WebHttp

### security

Transport

### Which .NET version

.NET 8

### Which os platform

Windows

```c#

```

### Stacktrace if any

```shell

```

g7ed6e:
Fixes #1764

## The bug

`MimeHeaderReader.AppendName` / `AppendValue` concatenated each chunk onto an immutable string. Both are called once per 1024-byte read buffer, and once more per folded continuation line, so an N-char MIME-part header cost O(N²) copying.

Two things make this cheaper to trigger than the issue describes:

**Folding, not buffer chunking, is the cheap path.** `ReadWS` jumps back to `ReadValue` on a leading space or tab, so every folded continuation line is its own `AppendValue`. A minimal fold is 4 bytes (`\r\n a`), so N bytes of input produce N/4 appends rather than N/1024.

**The buffer quota does not contain it.** `DecrementBufferQuota` caps a single header at `MaxBufferSize / 2` chars (~32K at the default 65536), but `MimeHeaders.Add` refunds the quota for every *unrecognised* header name. The budget therefore resets per header while the header count stays unbounded.

## Measured impact

MTOM decode path, default `MaxBufferSize` of 65536, net8.0:

| Payload | Before | After |
| --- | --- | --- |
| 65 KB message, one ~30 000-char folded header | 430 MB allocated | 0.7 MB |
| 1.3 MB message, 20 such headers | 10.2 GB allocated, 1.1 s | 11.4 MB, 74 ms |

Scaling linearly in header count, a request at Kestrel's default 28.6 MB body limit reaches a few hundred GB of copying — matching the reporter's estimate. This only affects endpoints using MTOM message encoding; `MimeHeaderReader` has no other caller.

## The change

`MimeHeaderReader` accumulates into two reusable `StringBuilder`s and materialises the string lazily in `Name` / `Value`. That is the whole fix — quadratic to linear, and no quota-accounting change is needed once accumulation is linear.

Deliberately unchanged:

- The `DecrementBufferQuota` call stays first and identical in both helpers, so quota accounting and the exact point at which it throws are unaffected.
- The per-chunk `new string((sbyte*)start, 0, len)` temporaries stay. They are bounded by the 1024-byte read buffer and so linear in total, and removing them would mean hand-widening bytes to chars — that would change how header-value bytes ≥ 0x80 decode (`new string(sbyte*, …)` already resolves differently on net472 vs net8, and this fix should not inherit that).
- Explicit `hasName` / `hasValue` flags rather than builder length. `x:<CRLF>` legitimately produces an empty value and must still count as a header that was read; length-based logic would return `false` from `Read` and silently drop every remaining header in the part.

## Tests

New `src/CoreWCF.Primitives/tests/MtomMimeHeaderTests.cs`. `MimeHeaderReader` is internal and the repo has no `InternalsVisibleTo`, so the tests drive it through `MtomMessageEncoder` — the payload is generated by the encoder's own writer, then the root part's header block is rewritten, so there is no hand-maintained boundary or content-id.

- Four correctness tests: a `Content-Type` that is both folded and long enough to span several read buffers (a recognised header, so mis-assembly fails to parse rather than being discarded), an empty header value, a header name spanning a read buffer, and a header value spanning a read buffer.
- One guard test, `HeavilyFoldedHeaderValue_DoesNotAllocateQuadratically` (net8.0+): asserts allocated bytes stay under 20 MB for a ~30 000-char header folded 4 bytes at a time. On the previous code this reports 430 MB. Allocated bytes rather than wall-clock time keeps it deterministic under CI load, and the measured read runs on a dedicated thread so the thread-local counter is not disturbed by an `await` resuming elsewhere or by parallel tests.

Verified the correctness tests pass identically before and after the change (including the pre-existing rejection of a fold placed immediately after the colon, which is unchanged), and that the guard test fails only on the old code.

`dotnet test src/CoreWCF.Primitives/tests` — 148 passing on net8.0/net9.0/net10.0, 145 on net472 (the guard test is net8.0+).
`dotnet test src/CoreWCF.Http/tests --filter "MessageEncoding|Streaming"` — 31 passing, covering MTOM end to end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlMtomReader.cs
+++ b/src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlMtomReader.cs
@@ -2099,6 +2099,14 @@ namespace CoreWCF.Xml
             EOF
         }
 
+        // Names and values are appended one buffer-full (or one folded continuation
+        // line) at a time, so they accumulate into StringBuilders rather than by string
+        // concatenation, which was O(N^2) on long or folded headers. The materialized
+        // string is cached because Name/Value are read once per Read() by the callers.
+        private readonly StringBuilder valueBuilder = new StringBuilder();
+        private readonly StringBuilder nameBuilder = new StringBuilder();
+        private bool hasValue;
+        private bool hasName;
         private string value;
         private byte[] buffer = new byte[1024];
         private int maxOffset;
@@ -2115,9 +2123,9 @@ namespace CoreWCF.Xml
             this.stream = stream;
         }
 
-        public string Value => value;
+        public string Value => hasValue ? (value ??= valueBuilder.ToString()) : null;
 
-        public string Name => name;
+        public string Name => hasName ? (name ??= nameBuilder.ToString()) : null;
 
         public void Close()
         {
@@ -2129,6 +2137,10 @@ namespace CoreWCF.Xml
         {
             name = null;
             value = null;
+            hasName = false;
+            hasValue = false;
+            nameBuilder.Clear();
+            valueBuilder.Clear();
 
             while (readState != ReadState.EOF)
             {
@@ -2143,7 +2155,7 @@ namespace CoreWCF.Xml
                     break;
             }
 
-            return value != null;
+            return hasValue;
         }
 
         private bool ProcessBuffer(int maxBuffer, ref int remaining)
@@ -2282,19 +2294,17 @@ namespace CoreWCF.Xml
         private void AppendValue(string value, int maxBuffer, ref int remaining)
         {
             XmlMtomReader.DecrementBufferQuota(maxBuffer, ref remaining, value.Length * sizeof(char));
-            if (this.value == null)
-                this.value = value;
-            else
-                this.value += value;
+            valueBuilder.Append(value);
+            hasValue = true;
+            this.value = null;
         }
 
         private void AppendName(string value, int maxBuffer, ref int remaining)
         {
             XmlMtomReader.DecrementBufferQuota(maxBuffer, ref remaining, value.Length * sizeof(char));
-            if (name == null)
-                name = value;
-            else
-                name += value;
+            nameBuilder.Append(value);
+            hasName = true;
+            name = null;
         }
 
     }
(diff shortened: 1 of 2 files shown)


src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlMtomReader.cs
```diff
@@ -2115,9 +2123,9 @@ public MimeHeaderReader(Stream stream)
             this.stream = stream;
         }
 
-        public string Value => value;
+        public string Value => hasValue ? (value ??= valueBuilder.ToString()) : null;
```
jakin-wang, line 2126: It seems like`value` and `name` is no longer set anymore and this can be simplified to `public string Value => hasValue ? valueBuilder.ToString() : null;`

src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlMtomReader.cs
jakin-wang, line 2191: Because `name` is no longer set, this check no longer makes sense here. We should use `!hasName` here since that inherits the logic.

src/CoreWCF.Primitives/src/CoreWCF/Xml/XmlMtomReader.cs
g7ed6e: value is assigned here with ??=.

jakin-wang: I understand that. My point is that it can only ever run once. Read(), AppendValue, and AppendName all clear it. So, the field can be populated at most once per header, and both callers read Name and Value exactly once per Read(), so a second access can never happen. hasValue/hasName now carry what value != null/name != null used to mean.

